PR #22, opened 2016-08-18
logout directive

robawi:
Please review my (probably to be amended) attempt for a logout directive. In particular I am unsure if it was ok to implement the directive in the home.js file. 
Thanks and regards Robert

### DIFF
--- a/client/app/components/home/home.html
+++ b/client/app/components/home/home.html
@@ -5,5 +5,6 @@
 <main>
   <div>
     <h1>Found in {{ $ctrl.name }}.html</h1>
+    <div><yunity-logout/></div>
   </div>
 </main>
--- a/client/app/components/home/home.js
+++ b/client/app/components/home/home.js
@@ -2,12 +2,12 @@ import angular from 'angular';
 import uiRouter from 'angular-ui-router';
 import homeComponent from './home.component';
 import homeHook from './home.hook';
-
+ 
 let homeModule = angular.module('home', [
   uiRouter
-])
+]);
 
-.config(($stateProvider) => {
+homeModule.config(($stateProvider) => { 
   "ngInject";
 
   $stateProvider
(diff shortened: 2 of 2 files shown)


nicksellen:
This needs quite a bit more work, just a quick list of things I noticed, maybe @Sneezoo has some other points:

* please use es6 style arrow functions
* try and not make whitespace only changes, it makes the diff hard to read
* make use of builtin angular directives (e.g. `ng-show`, `ng-click`) rather than manipulate elements directly (rarely needed)
* tests fail (`npm test`)
* linting fails (`npm run lint`) - having it setup in your editor is the best way
* does not belong in `home.js`, actually not sure it needs a directive just for a button, I would do something like `<button ng-click="ctrl.logout()">Log Out</button>` and the controller for that view would call the authentication service
* setting location.href then calling reload is redundant, and they both do a full page reload, so actually use the ui router methods, search the code for `$state.go` to find an example from elsewhere

It might be worth checking out https://github.com/AngularClass/NG6-starter (which we are basing the app on) to get familiar with how they do things, and https://github.com/rwwagner90/angular-styleguide-es6 might contain some useful things to (although we are not exactly following that).

robawi:
First of all thank you for taking the time to check and give feedback! Most of your hints are quite useful to me. I am obviously still early on my learning curve with yunity's tech stack and I will take some time looking at the resources you hinted at and at other code. 
As for using a directive I also believe that a super simple thing such as logout should have a much leaner implementation. In particular since the logout button should be required only once, i.e. no reuse potential. However, I understood that it was expected to have a directive and given that I am a newcomer I felt I had to meet the expectation. Talking about simplicity however I wonder if the yunity architecture is generally a few orders of magnitude more complicated than adequate. Personally I do not begrudge that, I am very happy to learn new technology. But probably the current approach means that yunity will need a very long time to reach it's goals. A strong sign for the unnecessary complexity are the many explicitly coded layers which even that simple  logout request has to ripple through to have it fulfilled by the backend + workarounds such as setting that csrftoken (which didn't even work when I tried it). In other projects of mine, logout was just 5 lines of code, including backend. It is basically just a side product when implementing the login which in itself is usually just around one hour of work, if it can't be even directly reused from another project.

nicksellen:
I'll reply to discussion points in slack, to keep this focused on the actual code changes.

Issue #32, opened 2016-09-20
Test for user service


tiltec:
This issue is just pro forma. I'm working on an test for the user data service (`app/common/user`)

PR #33, opened 2016-09-20
User service tests and improvements

tiltec:
#32


nicksellen:
@Sneezoo some of the reasons for the changes to Promise stuff are in the [commit messages](https://github.com/yunity/foodsaving-frontend/pull/33/commits/8be32977e13a825e17c740b392178ed276e2d089), but to pull it out here for clarity:

> They were not testable. Probably $httpBackend.flush() does not resolve the
Promises, unless they are derived from Angulars' $q object. Removing
them for now because they are not needed.

nicksellen:
What is the rationale for having both the user wrapper service and the user communication service? The wrapper service just passes everything through.

tiltec:
It looks to me like removing the wrapper is ok for now. The functions have the same signature as the functions they call, so actually adding the wrapper back in later seems to be easily possible.

PR #35, opened 2016-10-02
Group service#19

tiltec:


tiltec:
Looks really good! I would remove the wrapper layer, but this is a more general thing.

PR #38, opened 2016-10-03
Store service#37

D0nPiano:
Store service with get(params) method, that you can pass:
get({}) -> all stores
get({id: 5}) -> api/stores/:id
get({otherParam: xxx}) -> api/stores/?otherParam=xxx


D0nPiano:
Should be all done. Only question from my side: Does the get method make sense, or would you do it differently?

Sneezoo:
Good work!
We need to add query parameters to groups GET, too. So we can get all groups of a single user.

D0nPiano:
@Sneezoo group service is fixed already :slightly_smiling_face:

PR #39, opened 2016-10-03
Add fstool.yunity.org as default backend

nicksellen:
Added yConfig, with yConfig.defaults and yConfig.local (optional), to configure options for running the app.

Out the box will use dev.yunity.org as the backend for /api only (use swagger via direct backend url).


nicksellen:
Ready for review :)

nicksellen:
In conversation with @Sneezoo there is a possible issue with csrf tokens not working when provided over the webpack http proxy. Need to investigate. Currently the logout gives a 403 error, might be related.

nicksellen:
Hmm, tricky. Django is always checking if the referer uses https too.
https://github.com/django/django/blob/master/django/middleware/csrf.py#L226

Aha, I can set the referer header in the webpack proxy config. It's a bit hacky, but only used in dev, and only when using an https backend, and lets us keep the app deployed at dev.yunity.org matching production config, no hacks there (no cors, csrf enabled, etc...).

nicksellen:
Ready for review again @Sneezoo - login/signup/logout, cookies, csrf, all seem to work fine.

nicksellen:
It was intended as a general config file, not specific to the proxy.

One reason to not have all the proxy config in the yConfig is to minimise
the possibility of invalid configuration. The change origin option should
always be true, for example. Would also seem wierd to include functions in
the yConfig file.

Sneezoo:
Ok, so than we review the code and look for additional configs to be set there. I would still rename as the yXYZ convention is not used anywhere else.

I would override those properties which are essential. In this case functions in config seem reasonable as this setting solves an environment related issue.

nicksellen:
> Ok, so than we review the code and look for additional configs to be set there. I would still rename as the yXYZ convention is not used anywhere else.

My reasoning for `yConfig` naming was because I used the variable yConfig inside the gulp/webpack/proxy config, as `config` is used for the webpack config, then named the file the same for consistency. What would you name it? I am happy with another name. I wouldn't bother hunting for stuff just so it can go in the config, I would just add to it as needed.

> I would override those properties which are essential. In this case functions in config seem reasonable as this setting solves an environment related issue.

I am hesitant to put logic in the config that might need changing, when people develop with local backend they will need to add a local config file, and if this later needs changing it requires announcing to people, so keeping the bare minimum "config api" helps achieve this.

Sneezoo:
Made more toughts about it and I can't find reuse potential of yConfig right now. What about make backend a environment variable? If it's set we can add `onProxyReq`.. or something like that. I think it's not reasonable to have a config with only one property.
Otherwise I would just leave it like that and approve.

nicksellen:
I think a lot of people don't know how to set environment variables, and it's easy to run a command to forget to do so, e.g. `BACKEND=https://oisjefoijesfoijesf.com npm start` is very verbose compared to having it in a file.

Perhaps the perfect middleway is to use a `.env` file with https://github.com/motdotla/dotenv ? I would still handle the loading from env in a yConfig/<or other name> file, to have a place to push it into js objects, and potentially validate it.

Sneezoo:
We can set it in package.json temporarily. So `npm run dev` runs `BACKEND="https://dev.yunity.org" gulp`.
Keeps it simple and no one needs to set a environment variable. But we need to remove it later.

nicksellen:
Then `npm run dev` will not be usable for people with a local backend (i.e. me). The requirements are to support either local or remote backends, and to default to remote (for now).

tiltec:
I'm pretty fine with how this pull request is right now... given that this only touches devops stuff and no production code, I'd set different standards for naming here than for Angular services/components.
+1 to merge

Sneezoo:
And what about a *env* as @nicksellen suggested. But to minimize impacts on the gulpfile, `source` it right before `npm run dev`: `. config.env && gulp`?
I don't want this discussion to be to big so if there are any flaws with my suggestion I'd just approve this PR, I think the yConfig is too much, though.

nicksellen:
Ok, using a .env approach now (`.env` is standard filename for loading env vars).

I used `env $(cat .env)` pattern so you don't need to prefix contents with `export`, so we can use `BACKEND=https://foo.com` rather than `export BACKEND=https://foo.com` as using `.` would require.

Removed yConfig and load value from `process.env` directly.

nicksellen:
(rebased to master)

tiltec:
Would merge tomorrow morning.

PR #40, opened 2016-10-03
Add fontawesome

Sneezoo:


D0nPiano:
Thanks! (:

tiltec:
Nice! Seems to be hard to test at this stage.. I think we can merge, though.

PR #41, opened 2016-10-03
Tighten up style (inc. eslint rules)

nicksellen:
I'll use this PR to discuss/implement rule changes.

Our eslint config is quite slack right now, and allows much inconsistency. To prevent it drifting futher we need to tighten it up a lot.

Here's a quick overview of what I would change, it is intended to form the basis for discussion:
note: some of these things might not be eslint enforceable, hard to know without research...

* `a = b` over `a=b`
* `(a, b) => {}` over `(a, b, /* c */) => {}` (i.e. remove, not comment unused vars)
* `.then(successFn).catch(failureFn)` over `then(successFn, failureFn)`
* `firstName` over `first_name` (and move camel <-> snake case conversion further down the chain (in the service, just before http call))
* `'` over `"` (e.g. `"ngInject"` to `'ngInject`, etc)

Actually I don't think it's so much stuff, mostly it's just tightening the rules to enforce the existing style (will start those now).

Comments welcome @Sneezoo / @D0nPiano.


nicksellen:
You can see all the eslint errors with this new config in the [circle ci build output](https://circleci.com/gh/yunity/foodsaving-frontend/118?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link)

D0nPiano:
Sounds good. I don't find whitespace restrictions as necessary, but I'm fine with all the rules.

nicksellen:
Would value your input @Sneezoo, I would aim to have this done by tomorrow evening.

Sneezoo:
First things I'd like to point out:
- do we really want single quotes? Why not using double quotes only. Right now we'd have to fix this  tons of times. Single quotes are not used that often.
- existing services are causing problems as they are not factories. As we remove the base class it would be possible to use factories, but I can't see a point why it'd be worth changing. Class-like coding might be easier to understand for programmers that are more fimiliar with OOP than Prototype.

Will read through the eslint errors again this afternoon.

nicksellen:
> do we really want single quotes? Why not using double quotes only. Right now we'd have to fix this tons of times. Single quotes are not used that often.

Totally fine with double quotes. Although fixing a ton of times is not a problem (see `eslint --fix`)

> existing services are causing problems as they are not factories. As we remove the base class it would be possible to use factories, but I can't see a point why it'd be worth changing

Yup, will change that, I just used default angular rules.

nicksellen:
I pushed the autofixes with `--fix`, can rerun if we decide to change rules though... just wanted to get something going. It fixed almost all the linting errors (a few hundred down to a few).

Sneezoo:
Great! I am going to fix the remaining errors, except the ones related to service/base.js, as it will be removed anyway.

Sneezoo:
Had problems yesterday fixing window-service linting errors. I did a quick fix, ignoring this rule in *.spec* files.
The problem is that I can't replace `window` with `$window` for calling `window.module`
Didn't have a clue, maybe someone else has a good idea.

nicksellen:
I rebased to master, and resolved the small number of remaining issues, either directly, or by local eslint disabling, and adding a TODO.

This is ready to merge from my POV.

nicksellen:
I prefer single too. Lets keep it double for now though. Can discuss and change very easily in the future.

PR #46, opened 2016-10-07
PickupDate Service

tiltec:
Ready to review!


client/app/common/pickupDate/pickupDate.service.js
```diff
@@ -0,0 +1,47 @@
...
+    'ngInject';
+    super();
+    this.$http=$http;
+  }
+
+  create(pickup) {
+    return this.$http.post('/api/pickup-dates/', pickup).then(res => res.data);
+  }
+
+  get(params) {
```
tiltec: @D0nPiano this method can also be used to list all pickups. No need for a extra list()/pickupDates() method.
Do you like it? Would introduce this style also to the other services.

tiltec: See issue #55

D0nPiano: Yes, actually like that even better than get({}).

PR #47, opened 2016-10-07
Pickup list#34

tiltec:


D0nPiano:
I have two questions:
1. Where is the best place to put a filter? Or should they not be used in general? - currently, I just created one in pickupList.js, but it seems a little messy.
2. Shall the pickupListItem be defined in components/components.js file, even though at the moment it will probably only be used by pickupList? Or is it supposed to be defined in pickupList.js?

Sneezoo:
@D0nPiano 
Do we really need `groupByDate` and why? Haven't tried it with my workstation, yet. Is it important to group all goods in a pickup by date?

If we really need it we may think about doing data transforming elsewhere because filters are kind of a mess in Angular. Especially using them for complex stuff like grouping makes it ugly to read.
I am ok with filters that do easy stuff, though.

General proposal for defining filters (if anybody has crazy strong objections about what I have said above):
create a file like *pickup.filter.groupByDate.js* or whatever, put your code in there, import it to *pickup.js* and call `.filter(groupBy)`

D0nPiano:
I'd definitely group them at some point for the headlines. Even though in one store you don't have multiple pickups on one day, in one group there probably are multiple ones on one day, and I'd like to display them like:
```
-----------------------------
26.10.2016
------------------
12:30 Store Name
Members
-------------------
13:40 Store Name
Members
--------------

27.10.2016
-------------
12:30 ....and so on
```
wherefore the option to group them would be necessary.

I definitely see your point though, and I could just rewrite is without using a filter.

Sneezoo:
For me your code looks like you made groups with goods while your example is about pickups. PickUps already have a date to sort them, giving them a headline. In these pickups there are goods in a list meaning they are kind of grouped.

I am with you it maybe makes sense to group pickups by date.

tiltec:
I'm also ok with grouping pickups by date, just thought that foodsharing.de runs fine without and nobody complained so far (but they complain about many other things). Still, very nice idea.

screenshot_2016-10-13_17-04-33

PR #51, opened 2016-10-08
Remove duplicate hook implementation

Sneezoo:
@nicksellen @Xbaal @tiltec please review this
To have everything clean and not warning red we need to check the tests


nicksellen:
@Sneezoo we got the hook stuff under test coverage now :) It worked out simply in the end [1].

We did not mock the http in the end so you will probably say it is testing too much, but I argue it is a fine approach right now, using the well tested `$httpBackend` means we have to write less test/mock code (and reduce the chance of introducing errors in that).

[1] - https://github.com/yunity/foodsaving-frontend/blob/3c95db656c08aed664d81392534b57b61677b85d/client/app/common/authentication/hook.spec.js#L39 and below

Issue #45, opened 2016-10-06
Remove use of ES6 Promise


nicksellen:
ES6 `Promise` is not usable in angularjs, as `$digest` and `$httpBackend` (in testing) make use of internally fiddling with promise resolution. You can use `$q` service instead.

Issue #49, opened 2016-10-08
Remove base service, create validation service


tiltec:
base superclass is collecting random code, better separate it into individual services.

Xbaal:
What exactly do we want to validate?
I mean, what should be validated client-side and what do we just leave up to the server?

I would just start with not validating on client-side, because it is easier to code.
On the other hand it would be nice, if any wrong use of a function would directly throw a meaningful error.

tiltec:
You're right, validation seems not necessary at this point. We should rely on the server to return the status, even if it takes a bit more time.
Validation could be an optimization to speed up the frontend in future.

Sneezoo:
Validation is important, you should do this on both sides client and server. But that's an issue we could work on later. But at least I want to make an issue for that so that we won't forget (but overlook) it.
This looks like a good validation library which would make our code cleaner: https://github.com/hapijs/joi

PR #59, opened 2016-10-10
User service#56

Xbaal:
I simplified the user service and made its tests using chai-as-promised methods, which are very nice to use with Promises.

Simplifying means throwing out all validation stuff. It didn't really seem to follow some guideline.


Xbaal:
it closes #45 as well, as all occurences of ES6 Promise are replaced.

Xbaal:
it is a step towards closing #49 as well, because no method of the base service/class is being used anymore

PR #61, opened 2016-10-11
Hook provider#60

### LINKED ISSUE #60: Register hook factory in angular
At the moment the hook factory is loaded as an ES6 module, it would be better to load it via angular system.

I would suggest looking into defining a [provider](https://docs.angularjs.org/guide/providers), such that the hooks can be registered in a `config` block.

Xbaal:
closes #60

### DIFF
--- a/client/app/common/authentication/authentication.js
+++ b/client/app/common/authentication/authentication.js
@@ -1,6 +1,6 @@
 import Authentication from "./authentication.service";
 import uiRouter from "angular-ui-router";
-import { hookProvider } from "./hook.js";
+import hookProvider from "./hook.js";
 
 let authenticationModule = angular.module("Authentication", [
   uiRouter
--- a/client/app/common/authentication/authentication.js
+++ b/client/app/common/authentication/authentication.js
@@ -1,9 +1,12 @@
 import Authentication from "./authentication.service";
 import uiRouter from "angular-ui-router";
+import { hookProvider } from "./hook.js";
 
 let authenticationModule = angular.module("Authentication", [
   uiRouter
 ])
+
+.provider("hook", hookProvider)
 
 .service("Authentication", Authentication)
 
--- a/client/app/common/authentication/hook.js
+++ b/client/app/common/authentication/hook.js
@@ -1,21 +1,15 @@
 import angular from "angular";
 
-const setReaction = ($state, reaction, reactionHook) => {
+const setReaction = ($state, reaction) => {
   return () => {
-    let r;
     if (angular.isString(reaction)) {
-      r = $state.target(reaction);
-    } else {
-      r = reaction;
+      return $state.target(reaction);
     }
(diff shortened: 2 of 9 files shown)


client/app/common/authentication/authentication.js
```diff
@@ -1,10 +1,13 @@
 import Authentication from "./authentication.service";
 import uiRouter from "angular-ui-router";
+import { hookProvider } from "./hook.js";
```
nicksellen: maybe use default export here and name it `hook.provider.js`? looks like we are expecting multiple things to be exported from the module.

client/app/common/authentication/hook.js
```diff
@@ -1,21 +1,15 @@
...
-    }
-    return r;
+    return reaction;
   };
 };
 
-const hookFactory = (target, detour = { authenticated: true, anonymous: "login" }, reactionHook) => {
+const createHook = (target, detour = { authenticated: true, anonymous: "login" }) => {
   let hook = ($transitions) => {
     "ngInject";
```
nicksellen: there is no injection actually being used here, so maybe remove `ngInject`

client/app/common/authentication/hook.spec.js
```diff
@@ -1,5 +1,5 @@
 import AuthenticationModule from "./authentication";
-import hookFactory from "./hook.js";
+import { createHook } from "./hook";
```
nicksellen: could add another test in here to check it was registered in angular correctly

client/app/components/home/home.spec.js
```diff
@@ -3,6 +3,7 @@ import HomeModule from "./home";
 describe("Home", () => {
   let $rootScope, $componentController;
 
+  beforeEach(window.module("Authentication")); //to load hookProvider
```
nicksellen: in other places it imports the angular module and passes the value into the `module()` function (it is just the string `"Authentication"`, but would keep it consistent)

nicksellen: ... and doesn't the home module depend on the authentication module anyway?

client/app/common/authentication/authentication.js
Xbaal: @nicksellen I did it like this, because my test needs the createHook function from this file. It gets a bit messy, if you have default and regular export: 
`import defaultExport,{ theThingYouWant } from "someModule"`
I think, i would like to change the test to not be dependent on createHook, but i didn't see an easy approach.

nicksellen: in this case I think default, and named export is perfect. for actually using it you would only use the default export, and for testing it handily exports the intermediate function.

client/app/common/authentication/hook.js
tiltec: "$transitions" is injected, isn't it?

PR #63, opened 2016-10-11
Replaced karma-chai-plugins

tiltec:
use karma-chai-as-promised instead, since it has less unneeded plugins


tiltec:
CI fail seems not obvious to me. Needs to investigate further (possibly via SSH access). It shouldn't be the cache!?

tiltec:
Hm, the chai-sinon library seems useful as well, so maybe we stick with the original package.

Sneezoo:
Do we need a new issue for picking test libraries and plugins?

Issue #71, opened 2016-10-15
Include login.form.js in login controller


tiltec:
Currently, the login functionality is provided by loginForm.
Rewrite it to include it in the controller and use ngClick.

PR #72, opened 2016-10-15
Rework login component

### LINKED ISSUE #4: Tests for component Login
Currently it's using the generated default.

tiltec:
Closes #4 and #71


client/app/components/login/login.component.js
```diff
@@ -3,10 +3,9 @@ import controller from "./login.controller";
 import "./login.styl";
 
 let loginComponent = {
-  restrict: "E",
-  bindings: {},
   template,
-  controller
+  controller,
+  controllerAs: "ctrl"
```
tiltec: @nicksellen `controllerAs` in a component defaults to "ctrl", so I would suggest to leave this away (as before). I think we should accepts the implicit behaviour.
https://toddmotto.com/exploring-the-angular-1-5-component-method/

PR #78, opened 2016-10-17
change naming convention

Xbaal:
New naming convention:
(partial=without a route) components should have a "_"-prefix (only the folder)

TODO: 
- [x] apply the new naming convention to this branch
- [x] create new generator for page component


Xbaal:
`gulp page --name <page>` -> page generator (additional routing-related stuff)
`gulp component --name <component>` -> component generator (_<component> folder)

You could only use component and search for a leading "_" to decide which generator to use. But as you need seperate template files anyway, using seperate commands might be the better option.

@Sneezoo @D0nPiano @tiltec have a look, if i did it the way, you had in mind

D0nPiano:
For me it's fine either way. Both are easy to remember, so all is good 🙂

PR #79, opened 2016-10-18
Group integration

D0nPiano:


D0nPiano:
test !#/group/:id to see the groupDetail page with pickups, members and stores

client/app/components/_pickupList/pickupList.spec.js
```diff
@@ -0,0 +1,305 @@
...
+      $httpBackend.when("/api/auth/status/").respond(authData);
+      $httpBackend.when("/api/pickup-dates/?store=9").respond(pickupData);
+      $httpBackend.when("/api/stores/9/").respond(storeData);
+      
+      controller.userId = 1;
+      controller.addPickuplistInfos(pickupData);
+      let updatedData = controller.allPickups;
+      expect(updatedData[0].store).to.eventually.deep.equal(storeData);
+    });
+  });
```
Xbaal: I just had a look at this test, it seems really messy.
addPickuplistInfos(pickupData) gets as wel called by updatePickups(), which gets called once the Authentication is finished.
But this means, that the entries of controller.allPickups will already have promises in their .store attribute.
I am on this, i will comment here later.

Xbaal: btw i might be telling bs, i have been working for to long ;)

Xbaal: I got this working now, i have to figure out, whats the best way to contribute it here. Should i simply push?

D0nPiano: Simply pushing sounds good! Thank you for all your input! Going to see what i get done tonight.

client/app/components/_pickupList/pickupListItem/pickupListItem.styl
```diff
@@ -0,0 +1,54 @@
+a
+    color: inherit
+    text-decoration: none
+    margin-left: .5em
+
+img
+    width: 2em
+    height: 2em
+    border-radius: 50%
```
nicksellen: For that, we would need something like https://github.com/css-modules/css-modules (I tried it out in a toy project and it worked nicely). We could look into it in the future, not sure how it integrates with angular.

D0nPiano: Ah, okay. Well, till then I'll just prefix the used classes with the component name then...

D0nPiano:
So, the TODO that are left:

- [ ] make pickuplist tests prettier
- [ ] remove unnecessary comments
- [ ] add more useful groupDetail tests

client/app/components/_pickupList/pickupList.spec.js
```diff
@@ -0,0 +1,294 @@
...
+  let { module } = angular.mock;
+
+  beforeEach(module(PickupListModule));
+  beforeEach(module(UserModule));
+  beforeEach(module(StoreModule));
+  beforeEach(module(AuthenticationModule));
+  beforeEach(module(PickupDateModule));
+
+  beforeEach(inject(($injector) => {
+    $httpBackend = $injector.get("$httpBackend");
```
tiltec: Each time you use $httpBackend, check if there are no outstanding & requests:
```
  afterEach(() => {
    $httpBackend.verifyNoOutstandingExpectation();
    $httpBackend.verifyNoOutstandingRequest();
  });
```
I'll add this here :)

tiltec:
Great work so far integrating this into Angular structure @D0nPiano!

I saw that opening `http://localhost:3000/#!/group/1` fires 5 requests to `/api/stores/2/` directly after each other.. I'll try to figure out why that happens and find a solution.

client/app/components/_pickupList/pickupList.controller.js
```diff
@@ -0,0 +1,107 @@
...
+     * - isFull
+     * - store (if showDetail == store)
+     */
+  addPickuplistInfos(pickups) {
+    angular.forEach(pickups, (currentPickup) => {
+      currentPickup.isUserMember = currentPickup.collector_ids.indexOf(this.userId) !== -1;
+      currentPickup.isFull = !(currentPickup.collector_ids.length < currentPickup.max_collectors);
+
+      if (this.showDetail === "store") {
+        currentPickup.store = this.Store.getById(currentPickup.store);
```
tiltec: I think overwrite a value with a promise is bad style. Would it be ok to write this?
```
this.Store.getById(currentPickup.store).then((store) => currentPickup.store = store)
```
does it need a closure?

tiltec:
Where is `showDetail: date` used? could only find the "store" option.

client/app/components/_pickupList/pickupList.controller.js
D0nPiano: Yes, that sounds better actually!

D0nPiano:
@tiltec showDetail: "date" is more like a implicit default behavior. So if you didn't type in "store", it's always going to be "date" currently, wherefor you  can't see it in the code.

alangecker:
integrated the group selection into the topbar/logo.
In my opinion it's much more chicer, especially 'cause most people are never in multiple communities.
groupselector
groupselector2

for clarification: is there a difference between 'groups' and 'communities'?

D0nPiano:
I think I kind of lost track of this branch - what else needs to be done to merge this into master? I think all the todos I could think of are done now.

tiltec:
> showDetail: "date" is more like a implicit default behavior. So if you didn't type in "store", it's always going to be "date" currently, wherefor you can't see it in the code.

PickupList is only used in GroupDetail so far, with the `showDetail: "store"` - did you inted the "date" default for future use, maybe inside a storeDetail page?

tiltec:
I would propose to make the commit of @alangecker into a new pull request, so that we can merge this one.
Regarding the TODO: I will spend some time right now to reduce the number of calls to `/api/stores/2/`.

PR #83, opened 2016-10-21
Design Improvements

alangecker:
i know, at the moment the design is not so important, but this orange? seriously? :D I personally would much more enjoy working one a more beautiful/less aggressive frontend :)

this is an proposal how I would colorize the theme:
### old

design_old
### new

design_new
## Issues I had
- could not find where `$lightBgColor`  is defined. 
- somehow the file-loader does not load .svg's correctly, so i used a .png, but we should use vector-graphics wherever it's possible


tiltec:
@alangecker - looks good, too! Me personally, I'm also fine if it would be just black & white :D 

Generally, both color schemes make sense to me. @D0nPiano picked up the color scheme that Manuele proposed, while @alangecker uses the colors of yunity.org, which were originally picked by Raphael Wintrich and have the _earth_ tones of foodsharing.de.

D0nPiano:
Haha, I can definitely see why you are saying the orange is aggresive :D And of course the "earthly" colors look more like foodsharing.de

Nevertheless, and even though it took me a while to actually like them, I really am in favor of the orange by now. Not only that it sticks out, it's also very unique to yunity - kind of like the purple-ish color to ubunt. So i'm actually more in favour of keeping the orange. But maybe we could just have a vote on it somewhere when we have more pages ready (quick syscon? some slack channel?

tiltec:
I would like to have a totally different logo - yunity should just be in the footer (powered by yunity). I'll try to get some drafts from Tais, who painted one for the hackathon.

tiltec:
@alangecker How about a Syscon to move this forward? Could be very quick: 4 days proposals, 2 days voting!

tiltec:
I put the "question" tag here to make clear that this is an open question.

D0nPiano:
Is there any progress on this question? Are we just using the lifebelt? And the foodsharing ("earthtone") colors?

tiltec:
We could add themes support ;)

djahnie:
I support @alangecker in his request to change the color scheme to the less aggressive brown, that also holds the benefit of reminding people a little of foodsharing.de, a platform, which many international foodsaving enthusiasts came to know and crave...

tiltec:
@djahnie and me used this pull request to improve our usage of angular-material themes and change the colors to a more brownish tone.
The color could need some more fine tuning :) Also the splash pages could get a nice background image.

tiltec:
group

splash

That's how it looks right now. Opinions?

The splash pages (login etc) could look weird on bigger screens, it's not so easy to test on my laptop.

D0nPiano:
Cool! Especially like the background picture - althought I think I'd prefer a colored logo - it looks pretty dark (almost sad) at the moment.

tiltec:
@alangecker The green parts of the colorful logo are almost not visible in the main toolbar (when logged in)
Do you have an idea how to improve it?

nicksellen:
Personally I don't like it with the background image. It's kind of cheesy image to me, and is a bit stock-photo-y. Hand drawn tiled backgrounds I think can look super nice, still personal and not too corporate, e.g.:

stock-vector-grocery-food-and-products-seamless-pattern-cute-cartoon-hand-drawn-tile-background-vector-image-437087353

tiltec:
I think the black logo looks better then the colored version, so I would leave it for now (and revert the last commit).

Also @nicksellen the background image is the same as on yunity.org. If someone does a nice hand-drawn picture, we can replace it then.

nicksellen:
@tiltec I don't think a designer will find their way in and just do that, but if there is a loose consensus that it might look nice we can ask around specifically

(I don't like it on the yunity website either :p )

tiltec:
What about this?

foodsharing-worldwide

It's ideal as background though, because of it's large dominant structures..

djahnie:
Well, I _really_ like the yunity background picture and the other two proposals are... erm... rejected by me...^^
The current brown colors and the colored logo without opacity fit very well IMHO, so I'm in favor if these changes! Oh, and @alangecker is, too...^^

PR #84, opened 2016-10-21
integrated group menu into topbar

tiltec:
Open points from @Sneezoo 

-     create a component (maybe with bindings like currentSelection and groupList)
-     joining groups in the box will be a mess when the list gets bigger, I'd like to list only groups where the user is a member
-     we need to add real logic to it so the group service is called


alangecker:
### Screenshots
groupselector
groupselector2

tiltec:
I'll work on this a bit!

D0nPiano:
Looks really good! I'm only wondering whether it makes sense to pass activeGroup through all components (Page -> topbar -> groupMenu), or if there is another way to store it (service etc...).

Other than that I'm all for merging it (:

tiltec:
Service also sounded simpler for me to code, but somehow it makes sense to pass down those information. After all, what is shown there depends on the state of the groupDetail page. So groupDetail sets the active group and passes it to its children.

It's the philosophy of "having one way of information flow", which hopefully makes the architecture easier to understand for outsiders (got inspired to this by reading on Flux and the angular component structure). Of course it's ok to break the philosophy ;)

PR #88, opened 2016-10-21
Store detail#68

D0nPiano:


D0nPiano:
The topbar commit still needs to be reversed. Other than that it's done I think.

PR #96, opened 2016-10-22
CreatePickup component

Xbaal:
html code from mvp-design
not much done yet.

### DIFF
--- a/client/app/components/_createPickup/createPickup.spec.js
+++ b/client/app/components/_createPickup/createPickup.spec.js
@@ -9,10 +9,10 @@ const { module } = angular.mock;
 describe("CreatePickup", () => {
   beforeEach(module(CreatePickupModule));
   beforeEach(module(PickupDate));
-  
+
   beforeEach(() => {
-    angular.mock.module($provide => {
-      $provide.value('$mdDialog', {});
+    angular.mock.module(($provide) => {
+      $provide.value("$mdDialog", {});
     });
   });
 
--- a/client/app/components/_createPickup/createPickup.spec.js
+++ b/client/app/components/_createPickup/createPickup.spec.js
@@ -8,7 +8,7 @@ const { module } = angular.mock;
 
 describe("CreatePickup", () => {
   let $rootScope, $componentController, $httpBackend;
-  
+
   beforeEach(module(CreatePickupModule));
   beforeEach(module(PickupDate));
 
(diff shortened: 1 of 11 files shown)


D0nPiano:
Mostly working now.

@tiltec and @nicksellen - lint is failing, because the backend needs `max_collectors` field for pickup-date creation, which is not camelCased. Is there a workaround for lint, or should the backend be changed from `max_collectors` to `maxCollectors`?

nicksellen:
I think the backend should stay as _snake_case_  as that is the convention in python. For now, you can use [eslint's disable](http://eslint.org/docs/user-guide/configuring#disabling-rules-with-inline-comments) comments.

For later, maybe we can implement a function/service to automatically convert between snake_case and camelCase as needed (as it gets quite verbose).

D0nPiano:
Ah, thanks! eslint disable is what i was looking for! :+1:

D0nPiano:
- [x]  Added as Dialog
- [ ]  Test for createPickup() // currently failing because sql of `api/pickup-date -> date` converts to local time in some cases
- [x] update Pickuplist afterwards

D0nPiano:
I can't really think of anything else at the moment. Is it done? (:

client/app/components/_createPickup/createPickup.spec.js
```diff
@@ -0,0 +1,91 @@
...
+            return 22;
+          }
+        },
+        maxCollectors: 5
+      };
+      controller.createPickup();
+
+      $httpBackend.expectPOST("/api/pickup-dates/", {
+        max_collectors: 5, // eslint-disable-line
+        date: controller.date.toISOString(),
```
tiltec: @D0nPiano - I changed it to use the `toISOString()` function. it works quite well for me, what do you think about it?

D0nPiano: Looks way better! Thanks for changing it. (and the slider looks awesome too :+1: )

Xbaal:
I would like to see some error message, if the time isn't correct. I will have a look into this

D0nPiano:
@Xbaal Maybe we could also add a tooltip (https://material.angularjs.org/latest/demo/tooltip) that explains the correct format.

tiltec:
Looks good so far, I'll file an issue for smaller improvements.
Would merge.

PR #112, opened 2016-10-26
Integrate translation support

tiltec:
Todo:
- [x] let the user choose language
- [x] autodetect main language from browser

For later:
- remember language choice of user
- replace all strings in the repo with translation IDs

Nice stuff to read: https://angular-translate.github.io/docs/#/guide/07_multi-language

### DIFF
--- a/client/app/app.translate.js
+++ b/client/app/app.translate.js
@@ -1,0 +1,13 @@
+import localeEN from "./locales/locale-en.json";
+import localeDE from "./locales/locale-de.json";
+
+let AppTranslate = ($translateProvider) => {
+  "ngInject";
+  $translateProvider
+  .translations("en", localeEN)
+  .translations("de", localeDE)
+  .useSanitizeValueStrategy("escape")
+  .preferredLanguage("en");
+};
+
+export default AppTranslate;
--- a/client/app/app.translate.js
+++ b/client/app/app.translate.js
@@ -4,10 +4,15 @@ import localeDE from "./locales/locale-de.json";
 let AppTranslate = ($translateProvider) => {
   "ngInject";
   $translateProvider
+  .useSanitizeValueStrategy("escape")
   .translations("en", localeEN)
   .translations("de", localeDE)
-  .useSanitizeValueStrategy("escape")
-  .preferredLanguage("en");
+  .fallbackLanguage("en")
+  .registerAvailableLanguageKeys(["en", "de"], {
+    "en_*": "en",
+    "de_*": "de"
+  })
+  .determinePreferredLanguage();
 };
 
 export default AppTranslate;
--- a/client/app/components/_languageChooser/languageChooser.controller.js
+++ b/client/app/components/_languageChooser/languageChooser.controller.js
@@ -1,0 +1,14 @@
+class LanguageChooserController {
+  constructor($translate) {
+    "ngInject";
+    Object.assign(this, {
+      $translate
+    });
+  }
+
+  changeLanguage(langKey) {
+    this.$translate.use(langKey);
+  }
+}
+
+export default LanguageChooserController;
--- a/client/app/components/_languageChooser/languageChooser.controller.js
+++ b/client/app/components/_languageChooser/languageChooser.controller.js
@@ -3,14 +3,12 @@ class LanguageChooserController {
     "ngInject";
     Object.assign(this, {
       $translate,
-      activeLang: $translate.use(),
       providedLangs: ["de", "en"]
     });
   }
 
   changeLanguage(langKey) {
     this.$translate.use(langKey);
-    this.activeLang = langKey;
   }
 }
 
--- a/client/app/components/_languageChooser/languageChooser.controller.js
+++ b/client/app/components/_languageChooser/languageChooser.controller.js
@@ -2,12 +2,15 @@ class LanguageChooserController {
   constructor($translate) {
     "ngInject";
     Object.assign(this, {
-      $translate
+      $translate,
+      activeLang: $translate.use(),
+      providedLangs: ["de", "en"]
     });
   }
 
   changeLanguage(langKey) {
     this.$translate.use(langKey);
+    this.activeLang = langKey;
   }
 }
 
--- a/client/app/components/_languageChooser/languageChooser.html
+++ b/client/app/components/_languageChooser/languageChooser.html
@@ -1,0 +1,2 @@
+<md-button ng-click="$ctrl.changeLanguage('de')">DE</md-button>
+<md-button ng-click="$ctrl.changeLanguage('en')">EN</md-button>
--- a/client/app/components/_languageChooser/languageChooser.html
+++ b/client/app/components/_languageChooser/languageChooser.html
@@ -1,7 +1,7 @@
 <md-menu>
   <md-button md-menu-origin ng-click="$mdOpenMenu()">
     <i class="fa fa-globe fa-2x"></i>
-    {{ $ctrl.activeLang }}
+    {{ $ctrl.$translate.use() }}
   </md-button>
   <md-menu-content>
     <md-menu-item ng-repeat="langKey in $ctrl.providedLangs">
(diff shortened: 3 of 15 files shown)


D0nPiano:
Did you think of a way to get store and group descriptions in multiple languages yet? Are those going to be extra fields in the backend? Or should store->description just store and return something json-like (e.g. {"de": storeDescDE, "en": storeDescEN}?

tiltec:
I think that we don't need to provide translation support for user fields in the next 2-3 years. Some groups might need it (e.g. big cities with language mix like Singapore), but the could also default to a primary language and sometimes put the other langs behind a slash in the same field.

client/app/app.translate.js
```diff
@@ -0,0 +1,13 @@
+import localeEN from "./locales/locale-en.json";
+import localeDE from "./locales/locale-de.json";
+
+let AppTranslate = ($translateProvider) => {
+  "ngInject";
+  $translateProvider
+  .translations("en", localeEN)
```
nicksellen: Can we use full locale ids with country code too? `en` -> `en-US|en-GB`.

Some things are country based (e.g. US date formats are different to UK ones).

Angular [1] and angular translate support it. And the browser locale might include the country code.

... even if `en-*` is just mapped to `en` initially...

[1] https://docs.angularjs.org/guide/i18n

nicksellen: ...or underscores (e.g. `en_US`) to make http://userguide.icu-project.org/locale happy

client/app/components/login/login.html
```diff
@@ -3,24 +3,25 @@
   <div flex="nogrow" layout="column">
     <div flex id="loginContent" layout-padding class="md-whiteframe-2dp">
       <form login-form ng-submit="$ctrl.login()">
-        <h2>Login</h2>
+        <h2 translate>LOGIN.TITLE</h2>
```
nicksellen: Personally I prefer the style where you keep the translate key as the attribute value, (I think `translate="LOGIN.TITLE"`) so there is no chance the user sees the raw key rendered as text.

and does it HAVE TO SHOUT?

client/app/locales/locale-de.json
```diff
@@ -0,0 +1,10 @@
+{
+  "LOGIN": {
+    "TITLE": "Login",
+    "EMAIL": "E-Mail",
+    "PASSWORD": "Passwort",
+    "WRONG": "Die Anmeldedaten sind fehlerhaft. Bitte überprüfe sie sorgfältig.",
+    "SUBMIT": "Login",
+    "SIGNUP": "Neuanmeldung"
+  }
+}
```
nicksellen: I used angular-translate at letsmake, and I didn't do very well at keeping the translations keys in order, and it got confusing after a while to work out where a particular piece of text would appear on the website (if at all).

I don't have a clever solution, but just to keep it in mind!

Maybe if there is something that can parse the js (even just a regexp) and extract the keys that are used, could be useful. Dunno. Might just be something we have to keep on top of semi-manually.

client/app/components/_languageChooser/languageChooser.spec.js
```diff
@@ -0,0 +1,41 @@
...
+    beforeEach(inject(($injector) => {
+      $componentController = $injector.get("$componentController");
+    }));
+
+    it("should exist", () => {
+      let ctrl = $componentController("languageChooser", {});
+      expect(ctrl).to.exist;
+    });
+
+    context("stubbed translate", () => {
```
nicksellen: context doesn't need to be called "stubbed translate", I think that is a given... it's just an implementation detail of the test

client/app/components/_languageChooser/languageChooser.spec.js
```diff
@@ -0,0 +1,41 @@
...
+  });
+
+  describe("Controller", () => {
+    let $componentController;
+    beforeEach(inject(($injector) => {
+      $componentController = $injector.get("$componentController");
+    }));
+
+    it("should exist", () => {
+      let ctrl = $componentController("languageChooser", {});
```
nicksellen: `ctrl` -> `$ctrl` so it's consistent with inside views?

client/app/components/_languageChooser/languageChooser.spec.js
```diff
@@ -0,0 +1,41 @@
...
+        $translate = $injector.get("$translate");
+        sinon.stub($translate, "use");
+      }));
+
+      it("calls use method", () => {
+        let ctrl = $componentController("languageChooser", {});
+        ctrl.changeLanguage("de");
+        expect($translate.use).to.have.been.calledWith("de");
+      });
+    });
```
nicksellen: Maybe a bigger test that checks translation text is in a template? Doesn't really matter for now...

client/app/app.translate.js
tiltec: I guess this is better: `registerAvailableLanguageKeys()`

```
    $translateProvider
      .translations('en', { /* ... */ })
      .translations('de', { /* ... */ })
      .registerAvailableLanguageKeys(['en', 'de'], {
        'en_*': 'en',
        'de_*': 'de'
      })
      .determinePreferredLanguage();
```
https://angular-translate.github.io/docs/#/guide/09_language-negotiation

client/app/components/login/login.html
tiltec: yes, that's the other option. I prefer this one.. it's not bad if the user sees this.. and it's a bit less to type ;)

tiltec: Oh.. with the shouting is easier to discern the IDs from attributes

client/app/components/_languageChooser/languageChooser.spec.js
tiltec: I had another test before that actually calls use(), but I couldn't make it work (even with reading the documentation). So I deleted it and was too lazy to remove this separated space for the stub. Will do..

client/app/app.translate.js
nicksellen: Seems useful :)

client/app/locales/locale-de.json
nicksellen: Not sure which you thought is the better solution. It's not obvious to me either way.

client/app/components/login/login.html
tiltec: Ah forgot. Yes, that could be a good reason. I only thought of the case that a translation is broken and therefore the key is shown (which would still be better than nothing).

client/app/components/_languageChooser/languageChooser.controller.js
```diff
@@ -2,12 +2,15 @@ class LanguageChooserController {
   constructor($translate) {
     "ngInject";
     Object.assign(this, {
-      $translate
+      $translate,
+      activeLang: $translate.use(),
```
nicksellen: This would not get the updated value if the language is changed from elsewhere.

Probably find for now, but could create a service that contains a bindable object value (like the `CurrentGroup` service) that gets updated on the `$translateChangeSuccess` event (https://angular-translate.github.io/docs/#/guide/18_events) and would always contain the correct value.

nicksellen:
Is it ready for a final review or are you making changes still?

client/app/locales/locale-de.json
tiltec: I will think about including `webpack-angular-translate` later, not in this issue. Would still use `translate-default`, it could make the templates much easier to understand. Include the english string there.

client/app/components/_languageChooser/languageChooser.spec.js
tiltec: Reasons for leaving it there: I want to have a separate context, not stubbing "use()" for all tests. Don't know how to name it better.

tiltec:
Ready for your review! (Especially @nicksellen  :D)

client/app/components/_languageChooser/languageChooser.html
```diff
@@ -1,7 +1,7 @@
 <md-menu>
   <md-button md-menu-origin ng-click="$mdOpenMenu()">
     <i class="fa fa-globe fa-2x"></i>
-    {{ $ctrl.activeLang }}
+    {{ $ctrl.$translate.use() }}
```
nicksellen: Oh does that work?

tiltec: It does, would love to know why. I feel a bit naughty doing that.

Xbaal: `$ctrl.$translate.use()` is being called several times every time something changes. I wasn't aware of that.

PR #115, opened 2016-10-27
Rejig template structure, add CurrentGroup service

nicksellen:
Added concept of layouts. In the route/state config you can now
set:

```
parent: "main"
```

And it will automatically get the topbar component added, and
put the main content in a 70% width container so it's not too wide.

This required decoupling the topbar from the page components by
adding CurrentGroup service that holds the data of the currently
selected group instead of passing in the group data via the template.

### DIFF
--- a/client/app/app.component.js
+++ b/client/app/app.component.js
@@ -1,9 +1,0 @@
-import template from "./app.html";
-import "./app.styl";
-
-let appComponent = {
-  template,
-  restrict: "E"
-};
-
-export default appComponent;
--- a/client/app/app.html
+++ b/client/app/app.html
@@ -1,4 +1,0 @@
-<!-- Place all UI elements intended to be present across all routes in this file -->
-<div class="app">
-  <div ui-view></div>
-</div>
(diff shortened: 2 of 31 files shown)


nicksellen:
The inspiration for seperating the currently active group from the group service came from https://johnpapa.net/sharing-data-in-an-angular-controller-or-an-angular-service/

He recommends always keeping stateless services that fetch data seperate from services that hold model data. Seems wise.

nicksellen:
Ready for review now :)

There are 2 new concepts in here:
1. services that exist just to hold data (https://johnpapa.net/sharing-data-in-an-angular-controller-or-an-angular-service/)
2. template layouts, implemented via nested ui-router states

nicksellen:
New view with central container so it doesn't get too wide (currently 70% width, could perhaps have some max/min widths set too):
foodshaving1

... and some concession to mobile view, goes to 100% width at 600px or less (`xs` angular material width):
foodshaving-mobile

nicksellen:
It has one bug where when you leave a group that you are currently selected, the CurrentGroup isn't updated. Will push a fix shortly...

PR #121, opened 2016-11-02
Make all pages translatable

### LINKED ISSUE #119: Add default English text to tags
Add `<translate translate-default="Default text">Translation-ID</div>` to all translation strings.

Main advantage: meaning is directly visible in the template.

Tools like (https://github.com/DatenMetzgerX/webpack-angular-translate) can be used to extract the strings and export to translatable JSON.

tiltec:
I switched back to the `<span translate="ID"></span>` style, it seems easier to parse for my brain, also in combination with other translate attributes.

Closes #119

### DIFF
--- a/client/app/app.js
+++ b/client/app/app.js
@@ -1,7 +1,9 @@
 import angular from "angular";
 import uiRouter from "angular-ui-router";
 import ngMaterial from "angular-material";
+import ngCookies from "angular-cookies";
 import translate from "angular-translate";
+import translateStorageCookie from "angular-translate-storage-cookie";
 import Common from "./common/common";
 import Components from "./components/components";
 import AppMaterial from "./app.material";
@@ -17,7 +19,9 @@ import splashLayout from "./layouts/splash.html";
 angular.module("app", [
   uiRouter,
   ngMaterial,
+  ngCookies,
   translate,
+  translateStorageCookie,
   Common,
   Components
 ]).config(($stateProvider, $locationProvider, $urlRouterProvider, $httpProvider) => {
--- a/client/app/app.translate.js
+++ b/client/app/app.translate.js
@@ -4,6 +4,7 @@ import localeDE from "./locales/locale-de.json";
 let AppTranslate = ($translateProvider) => {
   "ngInject";
   $translateProvider
+  .useCookieStorage()
   .useSanitizeValueStrategy("escape")
   .translations("en", localeEN)
   .translations("de", localeDE)
--- a/client/app/components/_createPickup/createPickup.html
+++ b/client/app/components/_createPickup/createPickup.html
@@ -3,7 +3,7 @@
     <div class="md-toolbar-tools">
       <h2>
         <i style="font-size: 1.5em; margin-right: 1em;" class="fa fa-shopping-basket"></i>
-        <span translate="CREATEPICKUP.TITLE" translate-default="Create Pickup"/>
+        <span translate="CREATEPICKUP.TITLE" translate-default="Create Pickup"></span>
       </h2>
     </div>
   </md-toolbar>
@@ -38,10 +38,10 @@
 
       <div layout="row" layout-align="end center">
         <md-button ng-click="$ctrl.closePanel()">
-          <span translate="CREATEPICKUP.CANCEL" translate-default="Cancel"/>
+          <span translate="CREATEPICKUP.CANCEL" translate-default="Cancel"></span>
         </md-button>
         <md-button class="md-raised md-primary md-hue-3" type="submit" ng-disabled="createPickupForm.$invalid">
-          <span translate="CREATEPICKUP.CREATE" translate-default="Create"/>
+          <span translate="CREATEPICKUP.CREATE" translate-default="Create"></span>
         </md-button>
       </div>
     </form>
--- a/client/app/components/_createPickup/createPickup.html
+++ b/client/app/components/_createPickup/createPickup.html
@@ -3,7 +3,7 @@
     <div class="md-toolbar-tools">
       <h2>
         <i style="font-size: 1.5em; margin-right: 1em;" class="fa fa-shopping-basket"></i>
-        <translate translate="CREATEPICKUP.TITLE" translate-default="Create Pickup"/>
+        <any translate="CREATEPICKUP.TITLE" translate-default="Create Pickup"/>
       </h2>
     </div>
   </md-toolbar>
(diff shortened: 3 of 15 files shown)


tiltec:
I will undo the namespaces, as they are not compatible with the extract tool (and also not with filters)

tiltec:
I wanted to use this extract tool (https://github.com/DatenMetzgerX/webpack-angular-translate) to have the default text in the template and extract them to JSON for translation purposes. But it has limitations.

What works: use the translate directive with a default text and extract the default text to a JSON file

What doesn't work:

- didn't find a way to supply a default text to an HTML attribute
- `translate` filter has no default text
- `translate-attr` directive has no default text, is not supported by the extract tool
- `translate` service seems really complicated
- exported JSON file is a flat representation without nested namespaces
- `translate-namespace` directive is not supported by the extract tool
- `translate` element is not supported. using an arbitrary `any` element instead

Options:

1. Continue without translating HTML attributes.
2. Don't use default text, only translation IDs (should have telling names). locale JSON needs to be updated as you add strings.
3. Only add the translations with a filter (e.g. attributes) to a locale JSON, extract the other values and merge the files somehow

I'd either choose (1) or (2). Not translating attributes or not having strings in templates?

nicksellen:
Not having default text in templates seems reasonable to me, this is partly the purpose of the extract too presumably, so we can easily get the things that need to have translations/originals written.

I also don't see a big need for translatable text in html attributes - title attributes (as an example of html attribute text) is not even visible on mobile. Are there other use cases for html attribute text?

tiltec:
@nicksellen Can you rephrase your first point?
These extract tools don't seem to be useful to me if there is no default text.

Attributes: I think it's mostly about the title, but also aria-label. 
In other cases, attributes can simplify the template, e.g. md-tab. See my rewrite here: https://github.com/yunity/foodsaving-frontend/pull/121/files#diff-167c1324858ee5923010dabf743c327eR21

nicksellen:
If you extracted the text using the extract tool, but only set the translate key, i.e.:
```
<span translate="SOME.KEY"></span>
```
What would it extract?

I presume it would create a translation file with empty values. Which would be perfectly usable. It would show you what needs writing.

I say this not knowing what the extract tool actually does.

Also, why `<any>` tag? `<span>` seems more sensible to me, or I start wondering what any is, looks like it might be an angular directive...

tiltec:
> If you extracted the text using the extract tool, but only set the translate key, i.e.:
> ```
> <span translate="SOME.KEY"></span>
> ```
> What would it extract?

```
{ SOME.KEY: "SOME.KEY" }
```

I don't consider this output useful for two reasons:

1. lost context when writing the text. It's much better to write it directly in the template
2. future strings must be added manually

If the tool only delivers this, I would scrap it and create the `locale_en.json` file manually from the beginning.
However, if it puts the default text in the extracted file, then the extraction can automatically run and be uploaded to transifex.com for translation.

> ... `<any>` tag

It's really arbitrary, shouldn't have any meaning or effect. `<span>` seems to be equally suited & is standard, so I will switch.

nicksellen:
Ah, so it can't merge new keys into an existing translation file? (but it wouldn't be hard to write this ourselves)

Would it not be useful to be able to modify the english translations in transifex too?

Another problem with using default text is if the translation key is used multiple times.

tiltec:
> Ah, so it can't merge new keys into an existing translation file? (but it wouldn't be hard to write this ourselves)

A `npm run build` removes the `./dist` folder with the exported translation file, so I didn't test it. So far, every export gave a new JSON file.

> Would it not be useful to be able to modify the english translations in transifex too?

This would totally separate the strings from the template. My consideration is to make it easy for template devs.
But it could also be ok to simply render the Translation ID in the first draft of the page (having no text at all).

> Another problem with using default text is if the translation key is used multiple times.

Good point, this would be another reason not to use default text.

All in all, it seems to me that it would be better to not use the extract tool, not use default text and use a manual workflow:

1. When writing templates, use the `translate` directive with translation IDs. Recommendation: use an extra `<span>` element, it seems clearer to read.
2. Add the translation IDs with the English text to `locales/locale_en.json`
3. Use nested IDs (PICKUPLIST.ITEM.TITLE) for specific content, use top-level IDs for common text (BUTTON_OK_TEXT)

Xbaal:
@tiltec I think generally the workflow you suggested seems good.
I would like to not edit the language file directly, i think editing a big json file manually might be not much of a pleasure. I thought about automating this step with a tool, which might not exist probably.

I have thought of two ways the tool could work.
1st: It could read the translate-default attributes and then delete it automatically. 
`<span translate="KEY" translate-default="default key"></span>` 
gets converted to `<span translate="KEY"></span>` + edit in `locales/locale_en.json`

2nd: Or provide the english translation simply as text in the element: 
`<span translate="KEY">default key</span>` 
gets converted to `<span translate="KEY"></span>` + edit in `locales/locale_en.json`

This way, we don't have to edit the json directly. It is only a shortcut. But i am not sure, if there can be a "clean" implementation.

tiltec:
@Xbaal The JSON file won't be big in the next months (hopefully!), so I wouldn't use time right now to code a tool for this.
But if somebody wants to do it, the mentioned extract tool could potentially be extended (maybe forked). Translate filters and translation for HTML attributes still don't have ways to provide a default text. Your proposal only tackles directives. Style (1) is the way to go, style (2) overlaps with alternative angular-translate syntax to provide the key.

N.B. shouldn't it be _default value/string_, not _default key_ in your proposal?

nicksellen:
Transifex seems to be pulling back quite a lot from their open source angle. Back in march they had this nice open source page (that Paul linked from #development channel) https://web.archive.org/web/20160315084539/https://www.transifex.com/customers/open-source/ - but that page is gone now.

https://www.transifex.com/pricing/ still has this section:

> What are the basic requirements to qualify as an Open Source account?

> Although Transifex’s current source code is no longer available to the public to collaboratively develop, test and extend its functionality, we continue to maintain an “Open Source” program with the intent of supporting organizations that have no funding, revenue, and/or commercialization model. Our OS program also supports our technology partners with distributed teams and limited localization resources. To understand whether you qualify for our Open Source program, please contact us.

But sounds a bit more closed than it seemed before, and not so explicit.

tiltec:
@nicksellen I don't have immediate concerns regarding Transifex, our account is already set up and running: https://www.transifex.com/yunity-1/foodsaving-tool/dashboard/
In the long run, we could switch to _some_ other translation helper service, if they would decide to cancel their open source offer. The plain key-value JSON format should be flexible enough. I don't see that coming in the next year, however.

I'm planning to merge this pull request today, after the other pending one.
Following planned changes:

1. Remove all default text entries, add then to `locale_en.json`
2. Remove the extract tool

nicksellen:
Ah cool, I didn't realise we already had an account. Thanks.

Issue #127, opened 2016-11-02
Remove duplicate "Pickups" heading from groupDetail


tiltec:
I think the name in the Tab is enough. I would separate both components in future, see #86

Issue #120, opened 2016-11-01
Show display_name in pickup lIst


tiltec:
Problem: pickup list tries to load an user image, but we don't have user images.

Solution: show display_name instead

tiltec:
Hey @D0nPiano :) I think this and #114 overlap slighty. I didn't work on the other issue yet, so you are free to go here :smile:

Issue #114, opened 2016-10-27
Simplify controllers and tests of pickupList and pickupListItem


tiltec:
I tried to add the names of collectors and noticed that it's really hard for me to introduce this change. I want to have a clearer information flow and .. easier to understand code.
I will also simplify the tests, they have huge amounts of data in them which are not related to the unit.

Issue #86, opened 2016-10-21
Different purposes of GroupDetail and StoreDetail


tiltec:
It appears to me that @D0nPiano designed PickupList to be usable in GroupDetail, as well as a future StoreDetail page.
The PickupList has the additional parameter `showDetail`, which defaults to `date`, but is [currently only used](https://github.com/yunity/foodsaving-frontend/blob/master/client/app/components/groupDetail/groupDetail.html#L9) with `store`.

The bigger question to me is how different the GroupDetail and StoreDetail pages will be?

My view on this: 
## GroupDetail

Central page for a group, with the upcoming empty pickup slots and pickups slots where the user entered himself (=tasks to do, tasks that I promised to do). A list of all users, possibility to send messages to them (to discuss), a map of all stores in the group (to discuss)

In future, it could show recent updates to store descriptions and to posts in team chats (not implemented yet)
## StoreDetail

Show and edit the name & description, show all pickups (also completely filled ones), add new pickups, show pickup history (to discuss)

In future, it could show a team chat and let you post in it (not implemented yet).
# Summary

The different purpose of those could lead to different designs, e.g. the list of pickups in the GroupDetail may be simplified, without the filtering and sorting options, whereas the StoreDetail page is much more complex, as should mostly used be users (think store managers) who need these functions anyway.

D0nPiano:
I pretty much think the same - I'd leave it as the same component for now and just change the options with bindings.

I'd still like to remove th groupByDate sticky-headers when showDetail is "date". Other than that I find "showDetail" and "showCreateButton" to be sufficient for a difference between store & group page at the moment.

D0nPiano:
Is this still up-to-date, or can we close it? The changes in pickuplist solved the points mentioned above I think.

tiltec:
@D0nPiano I'm in favor of splitting, but other things are more important. The split gives us more flexibility, e.g. when redesigning the groupDetail page.

tiltec:
Hm... how do you feel about moving it to two different components? I added another ng-if to the template in #219 and I found it a bit hard to understand what content is actually shown.

tiltec:
Ok, I'm going to close this for now. Seems to be flexible enough ;)

PR #129, opened 2016-11-03
Rework PickupList

D0nPiano:
- [X] show display_name instead of image #120
- [X] remove duplicate heading #127
- [x] make pickupList Controller simpler #114
- [x] Remove Sticke headers when showDetail == "date" #86
- [x] Add option to hide sort & filter options #86
- [x] possibility to set the default filtering options (so that groupDetail only displays the ones you are part of, and storeDetail all) #86

### DIFF
--- a/client/app/components/_pickupList/pickupList.controller.js
+++ b/client/app/components/_pickupList/pickupList.controller.js
@@ -17,6 +17,7 @@ class PickupListController {
       showCreateButton: false,
       showDetail: "date",
       showTopbar: true,
+      showStickyHeaders: !(angular.isUndefined(this.options.showDetail) || this.options.showDetail === "date"),
       filter: {
         showJoined: false,
         showOpen: true,
@@ -26,7 +27,6 @@ class PickupListController {
     };
 
     this.options = angular.merge(this.defaultOptions, this.options);
-
     this.userId = -1;
 
     this.Authentication.update().then((data) => {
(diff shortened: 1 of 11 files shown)


tiltec:
I would remove the sorting option and only show the upcoming pickups first (ascending order). What do you think about it?

D0nPiano:
I'm pretty indifferent on removing the sorting functionality. It might be a bit useless, but it's also not bothering anything. So whatever you like better works.

Also, is there anything else that needs to be changed to make this mergeable?

client/app/components/_pickupList/pickupList.spec.js
```diff
@@ -239,7 +208,9 @@ describe("PickupList", () => {
         $scope: $rootScope.$new()
```
tiltec: What is the new rootScope good for?

client/app/components/_pickupList/pickupList.controller.js
```diff
@@ -2,22 +2,33 @@ class PickupListController {
...
-    this.reversed = false;
-    this.Authentication = Authentication;
-    this.PickupDate = PickupDate;
-    this.Store = Store;
-    this.userId = -1;
-    this.$filter = $filter;
 
-    this.$mdDialog = $mdDialog;
-    this.$document = $document;
+    Object.assign(this, {
```
tiltec: yeah! I like Object.assign() :)

PR #140, opened 2016-11-17
Editable fields

### LINKED ISSUE #123: Editable fields
Enable the user to change group/store/pickupdate fields (e.g. name, description).

tiltec:
Makes store and group details editable. I tried two different approaches, cause I didn't find a _best-looking_ way.

Possible tasks:

- [x] make it clearer that the user has to click there to edit
- [x] edit box should fit the original text
- [x] proper support for multiline strings

Closes #123.


client/app/app.js
```diff
@@ -8,6 +8,9 @@ import Common from "./common/common";
 import PageComponents from "./components/pages";
 import AppMaterial from "./app.material";
 import AppTranslate from "./app.translate";
+import AppXEditableConfig from "./app.xeditable";
+import "angular-xeditable";
```
tiltec: @nicksellen can you hint me how to do this better? I didn't see a way to import its name, seems not to support ES6 style.

tiltec:
Idea: have a edit button which makes the page editable (instead of clicking on the content to edit)
Reason: editing is not a very common operation, so it might confuse people who just want to select text and accidentally click.

client/app/app.js
tiltec: I've got no better solution for this, so I would just use this style. Similar to the mapintegration.

tiltec:
Situation: the angular-xeditable package.js includes jquery, but we don't use it (angular has it's own jqlite).

So - I would just merge this and try to remove the jquery dependency from https://github.com/vitalets/angular-xeditable/blob/master/package.json

Again, the snyk error is a false positive, jquery is not loaded.

tiltec:
I'll merge this in the next minutes :)

PR #145, opened 2016-11-20
Map integration

tiltec:
Using ui-leaflet, two map usages at storeList and createStore.

Resources:

- ui-leaflet docs: https://github.com/angular-ui/ui-leaflet/tree/master/doc
- ui-leaflet examples: https://angular-ui.github.io/ui-leaflet/#!/examples/simple-map
- leaflet reference: http://leafletjs.com/reference.html

Todo:

- [x] layout fix for "create new store" map
- [x] tests

Other maps can be added later on.


tiltec:
I'll just add the last test, then we can merge this!

PR #146, opened 2016-11-22
WIP: User page#128

D0nPiano:
ToDos:

- [ ] put more useful information on the page (joined groups?)
- [x] add controller test


tiltec:
I think it's ok to merge for now, it really just should exist.
Other things can be added when necessary.

PR #148, opened 2016-11-22
Design & Layout improvements

D0nPiano:
- [x] add shake animation to login form
- [x] loginContent shake instead of loginContainer
- [x] improve topbar layout
- [x] add margin between topbar and page
- [x] improve size of dialogs
- [x] groupmenu width
- [x] fixed login & signup width

Add tooltips for:
 - [x] groupDetail page for leave group button
 - [x] topBar for logout button

### DIFF
--- a/browserslist
+++ b/browserslist
@@ -1,0 +1,3 @@
+> 1%
+last 2 versions
+Firefox ESR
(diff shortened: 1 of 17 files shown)


alangecker:
login shaking results in a width >100%, so the horizontal scrollbar flashes multiple times
i fixed it by letting `loginContent` shake instead of `loginContainer`
just dont get it commited, but i guess you can change it on your own :D

everthing else seems fine ;)

PR #149, opened 2016-11-24
WIP: Conversations service

alangecker:
## ToDo's
- [x] create
- [ ] add methods `create()` & `writeMessage()`
- [ ] add tests

### DIFF
--- a/client/app/common/conversations/conversations.js
+++ b/client/app/common/conversations/conversations.js
@@ -1,0 +1,9 @@
+import Conversations from "./conversations.service";
+
+let conversationsModule = angular.module("Conversations", [])
+
+.service("Conversations", Conversations)
+
+.name;
+
+export default conversationsModule;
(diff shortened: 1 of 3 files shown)


tiltec:
Wow, thank you for tackling this @alangecker - I'm at your service if you need something.

Some thoughts: do you also plan to integrate real-time chat using socket.io? There is the `/socket/` endpoint configured at the server and it should also work in the backend.

tiltec:
We could merge the conversation service first, and then add the other components. Makes it faster to work on. What do you think?

alangecker:
good point! moved the ToDo's to an Issue #151, this PR is now only about the service.

i didn't planned it so far, because i'm still running into a lot of angular-related problems during just creating simple components...  :D

but i guess for somebody who knows how, it should not be that difficult to let the angular component listen for updates from the service

Issue #103, opened 2016-10-25
nest stores into groups?


alangecker:
How i imagine the hierarchy: `Page -> Group -> Store`
**shouldn't the url represent this (i.e. `/#!/Berlin/store/15`) and the components as well?**

![hierachy](https://cloud.githubusercontent.com/assets/10246027/19689888/251ec7d4-9acf-11e6-970b-2f58dbd64239.png)

in #84 there is currently a discussion about how to shape the data flow.
The reactway would be having the active groupId in _home_ and passing it down all the way to _groupMenu_ & _groupDetails_

right now the hierachy seems really chaotic and randomly, so we should create more mockups! :)

tiltec:
Looks good @alangecker!
I don't agree that what we have now is totally random and chaotic, but there surely can be many improvements!

To your points:
We have now `groupDetail`, which is essentially the homepage (it contains the topbar and the group details).

URL scheme could be `/<groupname>/store/<storename>` - it's currently not possible, as the groupname is not unique & urlsafe, and the storename is not unique per group, but I could implement this in the backend.

General URL scheme:

- `/<groupname>/` - group overview page (home)
- `/<groupname>/members` - detailed list of members
- `/<groupname>/stores/` - list of all stores in the group
- `/<groupname>/stores/<storename>` - store details
- `/login` and `/signup` for the usual stuff

D0nPiano:
Makes sense to me! I'm for the new URLs and making the groupName unique. ...is foodsharing.de allowing the same store name for two stores currently, @tiltec? e.g. when there are multiple REWE in one place?

I like the mochup and would be for implementing it this way in the desktop version - it'd only need to look good on mobile devices to (which might get a little messy)

> The reactway would be having the active groupId in home and passing it down all the way to groupMenu & groupDetails

I agree, I'd also really prefer it that way. Just having home have topbar & footer, and a view in the middle.

tiltec:
@D0nPiano
> is foodsharing.de allowing the same store name for two stores currently, @tiltec? e.g. when there are multiple REWE in one place?

Yes, it is allowed. But it's not so nice in lists when there are 5 stores with the same name.

The rationale for having unique store names per group could be something like this: users will use the name to refer to a store and not the ID, so make it easier for them to keep distinct names.
Rationale for not having globally unique store names: the groups have no interaction and users are mostly active in one group.

tiltec:
[...]

PR #167, opened 2016-12-02
Nest store into groups

tiltec:
Changes as discussed in #103. I did not nest the states. It doesn't seem fitting: we don't have a parent group page where all the subpages are nested in.

If I would create a parent group page, what should it have? Just a `ui-vieẁ` element to show the subpages?

What would be the subpages?

- storeDetail
- groupDetail
- members
- stores
- pickups


D0nPiano:
Could this already be merged? I'd rather do the ui-view stuff in a different pull request.

tiltec:
Yes, but I'd like to merge #168 first, because I added some tests there which might be helpful here as well. I would just go ahead, merge and resolve the conflicts. Then we can add this one, too.

PR #174, opened 2016-12-03
groupDetail Multiview

D0nPiano:
groupDetail now displays all information in 3 seperate boxes on desktop version. Mobile version still uses tabs.


D0nPiano:
@tiltec or whoever wants to review:
I'd like to get your opinion. I just changed the group page here, but I'm not quite sure if I like the result. What do you think? Should we leave it at the older version, or use this one?

[of course it's not yet ready, still got to be made more pretty etc.]

alangecker:
i do like the seperate boxes, uses the space much more on the desktop.
but we could still use tabs to hide not so often used features like the userList
...and imagine always that there will be much more in the future like a discussion board or events ;)

### improvement suggestions
- having the stores column always on the **left** side (also on storeDetail) as kind of a sub navigation
- Still tabs on desktop version to avoid having to much unnecessary informations at once
  ```Pickup | Discussion (later) | Members```

client/app/components/groupDetail/groupDetail.controller.js
```diff
@@ -1,6 +1,13 @@
 class GroupDetailController {
-  constructor(Group, CurrentGroup, $state, $stateParams) {
+  constructor(Group, CurrentGroup, $mdMedia, $scope, $state, $stateParams) {
     "ngInject";
+    $scope.screenIsSmall = !$mdMedia("gt-sm");
```
tiltec: Code style thing: we could outsource this to a service, if we want to have one application-wide definition of what _screen is small_ means. But it's also ok like this.

tiltec:
Do you think we could use the nested states from ui-router? https://ui-router.github.io/tutorial/ng1/hellogalaxy

There would be one parent `group` component, with children

- groupDetail
- storeDetail
- stores (maybe)
- members (maybe)

D0nPiano:
Like the ideas @alangecker!

@tiltec I agree that it makes sense. I think I'd merge this first though, just because otherwise it'd be a huge PR changing both at the same time.

client/app/components/groupDetail/groupDetail.controller.js
D0nPiano: Do you think it makes sense to create a global variable? Or do you mean creating a service that basically just returns $mdMedia("gt-sm")?

tiltec:
Looks amazing! :) Two more ideas/suggestions:

- use the full screen width (within reasonable amounts, let's say max-width 2000px?)
- make the headline with the group name bigger (the store list should visibly belong the the group)

D0nPiano:
Okay, never mind my previous comment. I stuck to your suggestions and moved everything to nested states. Only thing that needs to be improved is the storeDetail page (or also move it to groupDetail)

tiltec:
@D0nPiano do we have all the changes also in #237 and can delete this PR & branch?

D0nPiano:
Yes, storeMultiview is based on the most recent version here.

PR #178, opened 2016-12-07
Warn users on signup

tiltec:
Could be rewritten/styled more nicely, what do you think?
Also, should we put it on the login form as well?


alangecker:
should be a little bit more prettier now and yep, i think it should be on the login form as well, so that the people keep it in their mind.
Probably a hint how to contribute would still be great!

tiltec:
What about a link to our github repo with the github logo?
And a hint to https://github.com/yunity/foodsaving-frontend/blob/master/CONTRIBUTE.md ?

alangecker:
an github icon/link should be definitely always somewhere, but i think more about normal non-dev users having bugs or suggestions.

=> There is now the address `fstool@yunity.org` which gets forwarded into slack.

PR #184, opened 2016-12-07
Added contribute links

alangecker:
- [X] Github button & mail in splash
- [X] Github button & mail as footer

### DIFF
--- a/client/app/layouts/main.html
+++ b/client/app/layouts/main.html
@@ -6,3 +6,13 @@
     <ui-view></ui-view>
   </div>
 </div>
+<p class="footer">
+  made with &#9829; by <a href="https://yunity.org">yunity</a> |
+  contribute on <a href="https://github.com/yunity/foodsaving-frontend"><i class="fa fa-github"></i> Github</a> |
+  mail <a href="mail:fstool@yunity.org">
+    fstool@yunity.org
+    <md-tooltip md-direction="top">
+      {{ 'GLOBAL.DEVMAIL_NOTE' | translate }}
+    </md-tooltip>
+  </a>
+</p>
--- a/client/app/layouts/main.html
+++ b/client/app/layouts/main.html
@@ -7,9 +7,10 @@
   </div>
 </div>
 <p class="footer">
-  made with &#9829; by <a href="https://yunity.org">yunity</a> |
-  contribute on <a href="https://github.com/yunity/foodsaving-frontend"><i class="fa fa-github"></i> Github</a> |
-  mail <a href="mail:fstool@yunity.org">
+  made with <i class="fa fa-heart"></i> by <a href="https://yunity.org">yunity</a> |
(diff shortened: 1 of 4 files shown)


client/app/layouts/splash.html
```diff
@@ -1,4 +1,18 @@
-<language-chooser></language-chooser>
+<div>
+  <language-chooser></language-chooser>
+  <md-button href="https://github.com/yunity/foodsaving-frontend">
+    <i class="fa fa-github"></i>
+    Github
+    <md-tooltip md-direction="bottom">
```
tiltec: use the translate attribute instead of filters (less watches, less slowdown)
`translate="GLOBAL.GITHUB_NOTE"`

client/app/layouts/splash.html
```diff
@@ -1,4 +1,18 @@
...
+        {{ 'GLOBAL.GITHUB_NOTE' | translate }}
+    </md-tooltip>
+  </md-button>
+  <md-button style="text-transform:none"  ng-disabled="true">
+    <i class="fa fa-envelope"></i>
+    fstool@yunity.org <small translate="GLOBAL.DEVMAIL_NOTE">(only for Errors/Suggestions)</small>
```
tiltec: I'd remove the default text, it's duplicated in the locale-en.json

client/app/layouts/main.html
```diff
@@ -5,4 +5,14 @@
        class="md-whiteframe-2dp main-panel">
     <ui-view></ui-view>
   </div>
-</div>
\ No newline at end of file
+</div>
+<p class="footer">
+  made with &#9829; by <a href="https://yunity.org">yunity</a> |
+  contribute on <a href="https://github.com/yunity/foodsaving-frontend"><i class="fa fa-github"></i> Github</a> |
```
tiltec: should we leave this text english-only? I mean, devs will understand it..

alangecker: had troubles with angular-translate and HTML in the strings and it has not such an high priority, so i decided to leave it like that.
but feel free to change it if you want ;)

client/app/layouts/splash.html
alangecker: oh, that shouldn't be there! -> removed

client/app/layouts/splash.html
alangecker: done. 
but what about `<translate>GLOBAL.GITHUB_NOTE</translate>` ?

tiltec: Also possible! I was using this before, but I found that using one common method would be better.
The translate tag wasn't useful in some cases, so I would prefer to only use the translate attribute.

tiltec:
I'll do the following changes:

- make the mail link clickable
- remove the DEVMAIL note (or as tooltip)
- use fa-heart instead of entity
- use fa-envelope also in main.html
- use normal font size, it's better to read
- use opacity instead of color changes, it adapts better

Issue #151, opened 2016-11-26
Conversations


alangecker:
## ToDo's
- [ ] completing #191
- [ ] route `/conversation/with/:userId` for viewing/creating private conversations
- [ ] route `/conversation/new?participants=1,42,1337` for creating a new multichat
- [ ] participants selector
- [ ] 'Send Message' buttons on profil pages and pickup dates
- [ ] profil pictures

## Open Questions
- how to handle a conversation group, which got started for a specific pickup date, and somebody joins this date? (lower priority, but should be considered)

alangecker:
some progress.... ;)
![conversations](https://cloud.githubusercontent.com/assets/10246027/21027611/db5cd024-bd91-11e6-87e0-443b33f30ba0.png)

tiltec:
The backend needs to following changes:

- add participants to multichat
- leave multichat

Furthermore, the user should only be able to add participants that share a group with him.

tiltec:
We postponed the development of chat/conversations in our last development meeting. We had the idea of integrating an already existing chat system later (maybe Matrix/Riot), maybe something else.

tiltec:
Removing conversations in the backend? @alangecker 
https://github.com/yunity/foodsaving-backend/issues/254

PR #191, opened 2016-12-09
WIP: Conversations

alangecker:
### known issues: 
- [ ] Conversation does not get loaded: controller can not see the correct id on constructing since i merged the master locally
- [ ] Conversation list get reloaded on each thread change. how can we avoid that?
- [ ] date separator should only get showed when the date changes
- [X] and a lot of lint complaints^^

### ToDo's until merging
- [ ] tests
- [ ] new conversation
- [ ] send message
- [ ] i18n 
- [ ] use `$q` instead of native `Promise` with `$scope.$apply()`

further tasks under #151 

screenshot_2016-12-09_03-16-33


tiltec:
I would have a look over the code in the next days, if you feel like it. Tell me when you are ready ;)

D0nPiano:
> Conversation list get reloaded on each thread change. how can we avoid that?

Maybe like the nested views in storeMultiView?
So that `conversations` is the parent view
```
.state("conversation", {
      parent: "main",
      url: "/conversation/",                    // Removed ":id"
      component: "conversations"
}
```
of conversations/conversation, that has its own route:
```
.state("conversation.conversations", {
      url: "/conversation/:id",
      component: "conversation",
}
```

and instead of having the `<conversation></conversation>` component, you just include a `<ui-view>` element

D0nPiano:
> date separator should only get showed when the date changes

I actually had the exact same problem in the `_pickupList ` component. It was a little tricky, and in the end I just had a [groupByDate function](https://github.com/yunity/foodsaving-frontend/blob/conversations/client/app/components/_pickupList/pickupList.controller.js#L78) that get's called whenever the list is updated. I think the problem is quite similar here.

alangecker:
i'm gonna close this for now, because it has a pretty low priority and going be remade differently anyway.

PR #194, opened 2016-12-10
Upgrade angular to 1.6

tiltec:
Some improvements, however angular-material isn't ready for 1.6

# Improvements in AngularJS 1.6

>     Aligning jqLite with the latest version of jQuery (3.x).
>    Implementing long awaited features, such as support for inputs of type range and the ability to bind to any type of values using ngRepeat with select.
>    Disabling (by default) the pre-assignment of bindings on controller instances, which helps with support for native ES6 classes.
>    Changing the default $location hash-prefix to '!', as the previous empty string default was unconventional and confusing.
>    Reporting possibly unhandled promise rejections that would otherwise go unnoticed.

- https://github.com/angular/angular.js/blob/master/CHANGELOG.md
- https://docs.angularjs.org/guide/migration#migrating-from-1-5-to-1-6
- https://toddmotto.com/angular-1-6-is-here#component-and-oninit

I think we should change the constructor logic to always use $onInit or $onChanges.

# Problems in angular-material

https://github.com/angular/material/issues/10111

It might about the datepicker, that could be resolved quickly. So let's wait for some days.


lhrossi:
I don't know if helps, but in version 1.1.0 of angular material it works fine.

tiltec:
Thanks @lhrossi ! I downgraded material to 1.1.0 and fixed everything from our codebase to make it work with angular 1.6

The most important thing that we should keep in mind is that the constructor does not have the bindings available, therefore all the init stuff should be moved to the special lifecycle function $onInit.

Would like to merge this before any other PRs, maybe tomorrow? Resistance? @D0nPiano @nicksellen @alangecker

D0nPiano:
I didn't look at the branch myself yet, put upgrading sounds like a good idea.

lhrossi:
Hi @tiltec! acctually I'm not a contributor of the project, I just found your issue when searching to solve mine with the same error, so I shared here how I solve, I can't help you with the modifications. Good work!

tiltec:
@lhrossi that's how I understood it! Thanks very much for the hint :)

client/app/components/_pickupList/pickupListItem/pickupListItem.controller.js
```diff
@@ -1,12 +1,15 @@
...
-
-    this.collectors = [];
+    Object.assign(this, {
+      PickupDate,
+      User,
+      collectors: []
+    });
+  }
 
+  $onInit() {
```
tiltec: @D0nPiano just as a short example for the changes in angular 1.6: the setCollectors/setStoreInfo functions need to be called from $onInit instead of the constructor, because they need the data bindings.

(Another way could be: put them in a $onChanges hook, then they will get updated when the binding changes. It's generally the more flexible way :) )

PR #202, opened 2016-12-10
Markdown support for group description

tiltec:
I think that increases usefulness of the description field a lot

We could add a short HOWTO markdown as tooltip.


tiltec:
Using #203 seems better because markdown-it seems more fancy

PR #203, opened 2016-12-10
expandablePanel with markdown support

alangecker:
thought it could be useful to have a possibility to format text without getting length problems

### Todo's
- [x] fix animation on closing (stucks)
- [ ] check for XSS
- [x] content does not get updated after editing (probably angular-xeditable gets removed anyway) 

collapsed

expanded


tiltec:
Ok, I'm done for now.

XSS - I don't see a vulnerability, but I'm not an expert

Stuck animation... yeah, it's a bit slow. I would consider removing the "closing" functionality at all, I don't think will be used much.

PR #229, opened 2016-12-14
Show current user in topbar

tiltec:


tiltec:
@D0nPiano Would love to have your feedback on this :) It's just a small change, but seems quite important to have.

D0nPiano:
I like it! :+1: 
It's only a little problematic on the mobile version. I'm trying to work out something there...

PR #232, opened 2016-12-16
Show current user + Side Nav

D0nPiano:
Since in @tiltec  's request the buttons weren't shown in mobile, I thought of doing a little menu for the three buttons.

Well, and now I overdid it, and now it became a sidenav. :D

### DIFF
--- a/client/app/components/_topbar/topbar.controller.js
+++ b/client/app/components/_topbar/topbar.controller.js
@@ -1,8 +1,18 @@
 import logo from "./carrot-logo.svgimage";
 
 class TopbarController {
-  constructor() {
-    this.logo = logo;
+  constructor(Authentication) {
+    "ngInject";
+    Object.assign(this, {
+      logo,
+      Authentication
+    });
+  }
+
+  $onInit() {
+    this.Authentication.update().then((data) => {
+      this.loggedInUser = data;
+    });
   }
 }
 
--- a/client/app/components/_topbar/topbar.controller.js
+++ b/client/app/components/_topbar/topbar.controller.js
@@ -5,16 +5,16 @@ class TopbarController {
     "ngInject";
     $scope.screenIsSmall = !$mdMedia("gt-sm");
     $scope.$watch(() => {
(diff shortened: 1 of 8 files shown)


client/app/locales/locale-de.json
```diff
@@ -23,7 +23,8 @@
     "REDIRECT": "Umleitung zu deiner Gruppe..."
   },
   "TOPBAR": {
-    "LOGOUT": "Ausloggen"
+    "LOGOUT": "Ausloggen",
+    "GREETING": "Moin, {{username}}!"
```
tiltec: We need to enter this translation in transifex later, otherwise it gets lost on update.

tiltec:
Looks good! We might need something like this in future anyway.
Just add the missing test for toggleRight (and perhaps move the function to the controller), then we can merge it :)

client/app/locales/locale-de.json
D0nPiano: We always keep the english version though, right? So it just doesn't make sense to translate into other languages?

client/app/components/_topbar/topbar.controller.js
```diff
@@ -1,8 +1,30 @@
...
+  constructor(Authentication, $mdMedia, $scope, $mdSidenav) {
+    "ngInject";
+    $scope.screenIsSmall = !$mdMedia("gt-sm");
+    $scope.$watch(() => {
+      return $mdMedia("gt-sm");
+    }, (big) => {
+      $scope.screenIsSmall = !big;
+    });
+
+    $scope.toggleRight = () => {
```
D0nPiano: Ups, nope. No clue why I used scope there... just changed it (:

client/app/locales/locale-de.json
tiltec: yes, we use locale-en.json as source file for all translations. so the workflow is:
write key-value pair in `locale-en.json`, merge it to master, upload it to transifex, translate it, download the resulting `locale-**.json` files

Might put this in the README ;)

PR #237, opened 2016-12-18
Store detail multiview

D0nPiano:
# ToDos
## General
- [x] building routes for all items in the tabs
- [x] beeing more responsive.
- [x] fix storeList, storeList map etc..
- [x] use full screen width for mobile
- [x] fix duplication of stores in desktop view when opening `/group/1/stores`
- [x] rename Group -> GroupService

## Routes
- [x] change route names to:
```
 -> group
          -> groupDetail
                  -> description
                  -> members
                  -> stores
                  -> pickups
          -> store
```
- [x] change directory structure according to routes
- [x] add RouteTests

## Nice-to-have
- [x] highlight current store in StoreList
- [x] add whiteframe to userPage
- [x] group/*/description should only show one description on desktop
     -> what about editing the description?

## Fixes
- [x] fix tests
- [x] fix mdMedia use in groupDetail


D0nPiano:
## Group
screenshot from 2016-12-18 10-56-43
## Store
screenshot from 2016-12-18 10-57-57
## Mobile
screenshot from 2016-12-18 10-57-08

tiltec:
I'll fix the tests later tonight :)

tiltec:
I just have patchy online access, therefore I'll just dump my thoughts here while having a look at the code. Might do a more thorough review in the next days.

======

The route `/group/6/description` loads the description twice in desktop view.

Duplicated template code in group.html and its child components

Shouldn't the directory structure look more like this: (i.e. reflect the route)
```
group -> 
    groupDetail
    store
        -> storeDetail
```
        
We can have more than 1 component per module, that might make the module structure less bloated.

GroupDetailController: can `$state.current.name` change? Maybe this code belongs in a `$doCheck` or `$onChanges` function. To me, this code seems a bit confusing anyway, I don't understand why it is there.

groupDetail.html: there's a call to `$mdMedia` in the controller, but it's missing in the dependencies. Why doesn't the test fail? 

groupDetail.js: why `$state.groupData = group;`? Ah, it's used by many child components. Well, global state is easy, but will it bite us in future?

Route tests are missing or don't work (storeDetail.spec.js). BTW, also userDetail misses a route test.

client/app/components/group/groupDetail/groupDetail.html
```diff
@@ -25,7 +25,7 @@ <h1 flex editable-text="$ctrl.groupData.name">
     </div>
   </md-toolbar>
   <md-content layout-padding>
-    <div ng-if="$ctrl.$mdMedia('gt-sm')" editable-textarea="$ctrl.groupData.description">
+    <div ng-if="$ctrl.$mdMedia('gt-sm') && $ctrl.currentNavItem != 'description'" editable-textarea="$ctrl.groupData.description">
```
tiltec: this works if you open the URL, which is the common case I guess.
However, if I replace the default "pickups" in the URL with "description", ti still shows it multiple times...

I don't have an idea yet how to make it better though, just bluntly commenting ;)

tiltec:
`/group/1/stores` duplicates the stores list in desktop view.. obviously this is corner case (similar to the description route), but I wonder if we find a better solution than filling everything with triple conditionals... 

How do you generally feel about that? Merge now and fix later?

I'm a bit hesitant here, as this is already an enhancement to existing functionality, so I would make sure that it we don't introduce new problems.

D0nPiano:
I actually really don't know the best option there. So I am not even sure how I'd fix it. We could definitely default back to some other view, but on the other hand there might be a reason why someone would enter the URL group/*/description specifically. Really not quite sure though what's best.

nicksellen:
I'm sorry because I didn't really read the above text, but from the call, it seems it could have a client-side redirect (i.e. inside angular) so that when you visit `/group/1/stores` on a desktop browser, it'll send you to `/group/1` (as it already shows the store list).

nicksellen:
(and so in the routing definitions, they can be marked as `desktop: "/some/page"` and a hook could redirect to the desktop page if it's on desktop...)

client/app/components/group/group.controller.js
```diff
@@ -1,10 +1,18 @@
 class GroupController {
-  constructor($state, $mdMedia) {
+  constructor($state, $scope, $mdMedia) {
     "ngInject";
 
     if ($state.current.name === "group"){
       $state.go("group.groupDetail.pickups");
     }
+    
+    $scope.$watch(() => {
```
D0nPiano: Sadly I couldn't think of a better way to make the group breadcrumbs work. Happy if one of you guys could think of something without $scope.watch

tiltec: I'll look into this :)

tiltec:
I discovered some things that could/should be changed:

- `group` module name is duplicated (group service and group component) -> rename `group` service  module to `groupService`?
- group-specific stuff is in `group` as well as in `groupDetail` -> move all group specifics (incl whole template) to groupDetail?
- `createGroup` is child of `main` state, but the link is /group/create -> make `createGroup` a logical parent of `group`
- `storeDetail` is a child of `group`, seems to be more fitting as child of `group.groupDetail.stores`?
- the ui-view names `mainView` and `detail` are not really necessary, are they? I thought you need them when having multiple views per state

I did a lot of changes already, but right now nothing works anymore... so maybe I need to go back and start with smaller changes. What do you think about the points above?

D0nPiano:
> 1. `group` module name is duplicated (group service and group component) -> rename `group` service  module to `groupService`?

Sounds good to me!

> 2. group-specific stuff is in `group` as well as in `groupDetail` -> move all group specifics (incl whole template) to groupDetail and make `group` state abstract?
> 3. `createGroup` is child of `main` state, but the URL is /group/create -> make `createGroup` a logical parent of `group`

I think letting the group state resolve data for one specific group makes sense for URLs and data-wise (because both groupDetail and storeDetail need the data). `group.create` seems more clear, but really, what would the group state be doing then? It wouldn't resolve data anymore since `group.create` doesn't have data yet, and it also couldn't display a store list on the left anymore.

> 4. `storeDetail` is a child of `group`, seems to be more fitting as child of `group.groupDetail.stores`?

For me, group detail was more of a UI representation. So that:

- **group**: page with storeList on the left

- **groupDetail**: main page that shows the name & description
- and then a tabbed subviews of groupdetail (**stores, members, pickups**...)

Therefore it made more sense to me as group.store. But I think you also have a point. I think it just gets a little tricky with the ui-view elements, and it also makes the routes much longer. So I'd rather keep it the way it is. (probably goes along with the purposes of group and groupDetail generaly, so number 3 & 4)

> 5. the ui-view names `mainView` and `detail` are not really necessary, are they? I thought you need them when having multiple views per state

-> I'll look into that!

tiltec:
@D0nPiano thanks for you quick reply :)

It's ok for me to leave points 2-4 as they are and work on 1 and 5. I had a call with @nicksellen a few days a ago and we came to the conclusion that this PR is mostly about improving navigation (e.g. breadcrumbs and page titles), so we should only add as much hierarchy in the code as necessary. 

To me, the split between group and groupDetail is a bit confusing, but I could probably live with that.

tiltec:
Store
screen shot 2017-02-08 at 19 15 17

Group
screen shot 2017-02-08 at 19 14 57

Group on mobile
screen shot 2017-02-08 at 19 19 48

Amazing work @D0nPiano ! Can't wait to have this public!

Just some more comments - they might be too nitpicky, so if it's too hard, no need to solve them right now.

How to best navigate between stores on mobile view? Using the back button?

The name of the group is mentioned 3 times close to the top bar (navbar, breadcrumbs, content title bar).. could we reduce that by merging the content title bar with the breadcrumbs?

The store list in desktop view is quite prominent (good to have the map visible, but takes much space from the main content list). I wonder if it's better to move it to the right.

D0nPiano:
Thank you, glad you like it 🙂 

> ...might be too nitpicky

meh, I like any kind of feedback :)

> How to best navigate between stores on mobile view? Using the back button?

I think so. At least I couldn't think of a "clean way" yet. Do you have any ideas?

> The name of the group is mentioned 3 times close to the top bar (navbar, breadcrumbs, content title bar).. could we reduce that by merging the content title bar with the breadcrumbs?

Like a subtitle? But yes, agreed. I think it's also way too prominent still. And I don't like the linebreaks on mobile. I'll look into that!

> The store list in desktop view is quite prominent (good to have the map visible, but takes much space from the main content list). I wonder if it's better to move it to the right.

Agreed, especially on small desktop screens. Do you think just making it smaller also works for you? I'd like to keep it on the left somehow (I think sidebars are mostly on the left, and I always think that content-changing-buttons should be on the left of the changing content)

tiltec:
Re: store sidebar
Maybe reducing the distance between the list and the left screen border to zero would help? And letting the main view float in the middle.

PR #246, opened 2016-12-23
Breadcrumbs & Pagetitle

tiltec:
- [x] Design breadcrumbs
- [x] think of a nicer way to change document title (properly set the title on page load)
- [ ] test


tiltec:
I set the page title in a very easy way, without additional JS. so we don't need to add test, as far as I can see.

I hope there is no problem in moving ng-app from `body` to `html`.

tiltec:
Let's merge this first before changing things on #237  @D0nPiano

D0nPiano:
Breadcrumbs are now just underneath the topbar, small & grey:
screenshot from 2017-02-02 20-39-53

@tiltec Where are we going to put the tests (since it's not a component)?

tiltec:
@D0nPiano we can only test JS code for now, I don't really know how to test HTML+CSS and Angular statements inside HTML.
I will merge this now into the other branch, it seems to get complicated when having both in parallel.

Issue #231, opened 2016-12-15
Add confirmation Dialog to leave Group


D0nPiano:

tiltec:
Good idea, I think it makes sense, but it's not a high priority to me..

tiltec:
After getting feedback from Nantes, I would add the delete pickup confirmation soon.

tiltec:
Confirmation for delete pickup is implemented.

alangecker:
resolved with #273

PR #273, opened 2017-02-06
Add confirmation dialogue to leave group (RE: #231)

crhistianr:
The only LEAVE_TEXT I was confidently able to translate on my own was English and Spanish. 
Re #231


tiltec:
Hello @crhistianr and thank for your contribution. I'll have a closer look at it later, but it seems perfect - congrats!

Don't worry about the translations, actually it would have been enough to just add the English locale. We use transifex.com to manage the translations into other languages. I might update this on some readme file ;)

tiltec:
@crhistianr one test fails for me:
"should be able to leave a group" in `groupDetail.spec.js`

Can you reproduce this by running `npm run test`?

tiltec:
Since we've had big changes in master, I will merge this PR and fix the tests. Looking forward to hearing more from you!

PR #276, opened 2017-02-07
Added favicon

D0nPiano:
Can't believe we didn't have one before :laughing:


D0nPiano:
screenshot from 2017-02-07 14-48-38

PR #281, opened 2017-02-09
address autocomplete & map

### LINKED ISSUE #206: Show list of results when looking up addresses
Currently we use only the first result of a OSM Nominatim query. This is not always the result that the user was looking for.

Could either show more results as a list, or use a lookup field inside the map (https://github.com/perliedman/leaflet-control-geocoder).

alangecker:
### Components:
- [x] createGroup
- [ ] createStore
- [ ] storeDetail
- [x] groupDetails

closes #206


tiltec:
Thumbs up for using md-autocomplete! We just need to check that we are not sending too many requests to the server (I remember that the osm nominatim service defines fair use at 1 req/sec. I guess per user, not per page)

alangecker:
yeps, 1 req/s ([source](https://operations.osmfoundation.org/policies/nominatim/))
i've added a `md-delay` which waits for 500ms after the last key press, should be enough ;)

tiltec:
Maybe just do the group create & group edit here. I'm trying to merge both pages together & use ui-router for the transitions.

PR #282, opened 2017-02-09
Design changes

### LINKED ISSUE #265: Use expandable panel for group preview

alangecker:
### joinGroup
- [x] only one description at a time
- [x] separate password form _(in preparation for further options like a quiz)_
- [x] expandablePanel for group preview? _(Discussion: #265)_

### other changes
- [x] splash topbar buttons aligned
- [x] changed loading bar style
  _using the topbar as a loading bar is a nice idea, but resulted in a lot of problems, i.e. when page is scrolled down_
- [x] only one home button & aligned button
  _one common button to the groups home page_
- [x] replaced `...` with animated spinner for the initial loading
- [x] smaller footer
   _looks better with a smaller font size then the normal content_

closes #265


tiltec:
Can we merge this after #237 ? Might be easier to resolve the merge conflicts.
Or we merge it _into_ #237 ;) ... 

I think we need a separate _deploy_ branch soon separate from a _master_ for all the development, so we can merge together PRs more quickly... @nicksellen opinion?

client/app/components/_joinGroup/joinGroup.controller.js
```diff
@@ -21,11 +25,27 @@ class JoinGroupController {
...
-  joinGroup (group) {
-    return this.Group.join(group.id, { password: group.password })
+  joinGroup() {
+    return this.Group.join(this.active.id, { password: this.password })
     .then(() => {
-      this.$mdDialog.hide(group.id);
+      this.$mdDialog.hide(this.active.id);
+    })
+    .catch(() => {
+      // TODO: set form to invalid
```
tiltec: Let's fix this before going into master?

client/index.html
```diff
@@ -12,8 +12,51 @@
     <link rel="shortcut icon" href="app/icon/carrot_logo.ico">
     <link rel="icon" type="image/png" href="app/icon/carrot_logo.png" sizes="32x32">
     <base href="/">
+    <style type="text/css">
```
tiltec: I'll put the loading dots in a class called `angular-pre-loading`. Just to make it clear what the code is for.
Even better: put it into a separate file, but that requires some webpack magic...

client/app/components/_joinGroup/joinGroup.controller.js
alangecker: any idea how to access the form methods?
`this.$scope.form.password.$setValidity("check", false);`, which was used before, did not work somehow

tiltec: I can have a look into this... here's some more documentation: https://docs.angularjs.org/api/ng/type/ngModel.NgModelController

client/index.html
alangecker: i put in intentionally in the index.html, that the loading spinner is avaiable as soon as the first data is loaded ;)

Issue #291, opened 2017-02-14
Create Group not working


D0nPiano:
Create Group has to be updated to match the new backend fields:
- [x] Name
- [x] Description
- [x] Public description
- [x] Address
- [ ] Latitude
- [ ] Longitude
- [x] Password
- [ ] Timezone

PR #294, opened 2017-02-14
Fix: createGroup needs timezone

D0nPiano:
See #291


D0nPiano:
...might also just wait for address autocomlete and map #281

alangecker:
with #281 there will be a new address picker
unfortunately nominatim does not resolve the timezone, but we could fetch the timezone over a different API
As soon as the form sends the coordinates, we could also calculate the timezone out of them in the backend... @tiltec what do you think about that? 

----
further there's an error ;)
`Error: Param values not valid for state 'group.groupDetail.pickups'`

tiltec:
We can definitely calculate the timezone in the backend, but I'm not sure if we want to make the address mandatory (might be better for distributed groups/coordination groups). Then the timezone could be set by a user.

PR #297, opened 2017-02-15
PickupList: do a single request per store

tiltec:


D0nPiano:
How about
```
angular.forEach(pickups, (currentPickup) => {
        if (!angular.isDefined(stores[currentPickup.store])){
            stores[currentPickup.store] = this.Store.get(currentPickup.store);
        }
 });
```
in the first loop - then it only has to loop once.

tiltec:
We could also embed the lookup in the other loop, then it's only one!

PR #306, opened 2017-02-16
Show favicon

tiltec:


tiltec:
@alangecker can you delete the branches after merging? There's very convenient button ;)

PR #312, opened 2017-02-17
feat: add $mdToast for HTTP5xx and timeout

tiltec:
the parent container should not be scrollable, so I added `<md-content>` as child of `<body>` and set it to 100% height.


tiltec:
Where could I put a test for this? I'd also be fine without one ;)

D0nPiano:
Works, and code looks good to me. I would approve if I had a button :D

PR #313, opened 2017-02-17
refactor: don't nest md-content elements

tiltec:
It would only be suitable for locations where we want to have multiple scrollbars (e.g. in dialogs, where the outer scrollbar is disabled). I quickly checked and the site worked in the same way with this change. I think a md-content without a defined height is basically the same as a div.

Reference: https://material.angularjs.org/latest/api/directive/mdContent

Together with #312 it would also mean that we should only use md-content in md-dialog, if necessary. @D0nPiano what do you think about this?

### DIFF
--- a/client/app/components/_createStore/createStore.html
+++ b/client/app/components/_createStore/createStore.html
@@ -6,7 +6,7 @@
     </h2>
   </div>
 </md-toolbar>
-<md-content layout-padding>
+<div layout-padding>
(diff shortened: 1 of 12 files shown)


D0nPiano:
I think it sounds reasonable - and having nested scrollbars really is a pain in many cases, so I'm for your proposal. I'll look through the changes tonight.

D0nPiano:
The changes look good too (:

Issue #310, opened 2017-02-17
Add markdown short-help to store description


tiltec:
It is rendered with markdown as well.

Follow-up to #308

D0nPiano:
I think this goes along with replacing xeditable #205

Issue #305, opened 2017-02-16
StoreDetail: reload pickups when weeks_in_advance setting is changed


tiltec:
This setting controls how far in future the pickups should be generated.

Issue #205, opened 2016-12-11
Replace angular-xeditable with custom code


tiltec:
angular-xeditable has some problems, e.g. the inputs have to be predefined as directives and access to the temporary values is hard.

So I would suggest to replace it. Maybe with a very lightweight directive that supports the state transition (editable/noneditable). I already started in #198 with custom input fields. Let's discuss here about architecture and then replace the existing usages in storeDetail and groupDetail.

PR #331, opened 2017-02-20
Store edit #310

D0nPiano:
Issues fixed:
- Replacing xeditable #205
- adding markdown-input to storeEdit #310
- reloading PickupList in StoreDetail when settings changed #305


tiltec:
I'll add back the address. 
Also, the store-detail-map stopped working (missing $ctrl.mapData). I'll try to get working again.

Issue #213, opened 2016-12-12
Show history of field changes


tiltec:
I would track the changes of the following fields:

- store add
- store/group title/description/address/lat/long
- pickup dates add/delete

Needs nice ways to integrate it in the frontend and some intelligent backend changes...

tiltec:
@alangecker this is our tracking task the audit log.

PR #333, opened 2017-02-21
WIP: History of pick-ups & edits

### LINKED ISSUE #177: Show historic pickup dates
Pickup dates should only contain the buttons ```LEAVE``` / ```JOIN``` when they aren't in the past.

alangecker:
closes #177 and #213


tiltec:
Can you separate the profile picture into a separate PR? I think it could take a while until we have the backend ready for it and I would like to have the history in production before that.

alangecker:
the current solution doesn't require the backend implementation, it only generates  default custom avatars without requesting one from the backend.
Without the avatar we would have to touch the history design twice, or with removing it: thrice.
But now it is already done ;)

so far everthing is done here, except:
- [ ] review - maybe a better solution?
- [ ] `storeName` in payload, to save the name permanently and avoiding a unnecessary complex frontend
(would require a controller for each history-typus to resolve the missing data)
- [ ] adding all the missing typus strings

-

tiltec:
I will

- add the missing strings 
- add some tests
- add the store name to PICKUP_DONE payload

Issue #365, opened 2017-03-02
Use landing page title font also on splash pages


tiltec:
It looks nice, could also be used on splash pages.

Alternative: since we show the logo with text on the landing page, we could remove the text "Foodsaving worldwide" from splash pages and just use the logo.

devishi:
Do you aim to only include the logo (center aligned)?
Else, we could create a photoshop image with the logo as well as the text  "Foodsaving worldwide" to make it look consistent and explanatory?

tiltec:
@devishi I think a image logo+text could be nice there! Maybe having two versions, one with prominent text and one with small text? 

I don't have a strong opinion yet, except that I really like vector graphics, so keeping the branding as SVG seems worthwhile to me. :)

PR #369, opened 2017-03-04
Formatted splash page to match landing page text Issue #365

anne27:
Increased the logo size to make it look more prominent.
Also included text to make it consistent across pages.
Please suggest anything you'd like to be changed!
screenshot from 2017-03-05 03-28-11

### DIFF
--- a/client/app/app.styl
+++ b/client/app/app.styl
@@ -9,7 +9,7 @@
   logo
     padding-right 10px
     > img
-      height 60px
+      height 200px
       opacity 0.8
   @media screen and (max-width: 450px)
     h1.title
--- a/client/app/app.styl
+++ b/client/app/app.styl
@@ -1,7 +1,7 @@
 .splash
   background-image: url('yunityorg.jpg')
   background-size: cover
-  heading
+  h1
     font-family: 'Cabin Sketch', sans-serif
     color: #795548
     font-size: 50px
@@ -56,4 +56,3 @@ body > md-content
   .leaflet-top
   .leaflet-bottom
     z-index 75 !important
-    
--- a/client/app/app.styl
+++ b/client/app/app.styl
@@ -1,6 +1,11 @@
 .splash
   background-image: url('yunityorg.jpg')
   background-size: cover
+  heading
+    font-family: 'Cabin Sketch', sans-serif
+    color: #795548
+    font-size: 50px
+    
   min-height 100%
   > md-toolbar
     height 42px
@@ -51,3 +56,4 @@ body > md-content
   .leaflet-top
   .leaflet-bottom
     z-index 75 !important
+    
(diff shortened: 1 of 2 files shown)


client/app/layouts/splash.html
```diff
@@ -1,3 +1,10 @@
+<style>
+md-heading-splash{
+    font-family: 'Cabin Sketch', sans-serif;
+    color: #795548;
+    font-size: 50px;
+  }</style>
+  <!-Can be added to angular-material.css>
```
tiltec: angular-material is not written by us, so we can't change it. I would recommend to put this style in the `app.styl` (supports CSS, but also the simplified [Stylus language](http://stylus-lang.com/). And maybe drop the `md-` prefix to make clear that it's our style and not from Material design :)

What about a structure like this:
```styl
.splash
  heading
    font-family: 'Cabin Sketch'
     // ... styles from splash.html
  // other styles 
```

(Further thought: maybe we should put the splash page style in a `splash.styl` next to `splash.html`... but I guess that's another task)

client/app/layouts/splash.html
```diff
@@ -21,7 +28,7 @@
   <div layout="column" layout-align="center center" flex="noshrink" layout-margin>
     <div layout layout-align="center center">
       <logo></logo>
-      <h1 class="title">Foodsaving Worldwide</h1>
+      <md-heading-splash align="center">Foodsaving<br>Worldwide</md-heading-splash>
```
nicksellen: custom element names are intended for angular directives or components, if its just a normal element then using the most semantic element is best, such as h1

anne27:
Updated splash.html and app.styl.
Now the formatting and the HTML code are in different files. :)

client/app/layouts/splash.html
```diff
@@ -21,7 +21,7 @@
   <div layout="column" layout-align="center center" flex="noshrink" layout-margin>
     <div layout layout-align="center center">
       <logo></logo>
-      <h1 class="title">Foodsaving Worldwide</h1>
+      <heading align="center">Foodsaving<br>Worldwide</heading>
```
nicksellen: heading isn't a proper element either. you can see them at https://developer.mozilla.org/en/docs/Web/HTML/Element. you can pick random words and they will behave like div elements but there should be good reason (like angular directives/components)

tiltec: Agreed - we should reserve custom names for angular. Maybe something for the `contribute.md` (which I didn't have a look inside for quite some time)

anne27: Thanks, that was a great article.  Tere are only two options- <h1-h6> or <header>. <header> might sound a tad ambiguous. I guess <h1> makes more sense here?

client/app/app.styl
```diff
@@ -1,6 +1,11 @@
 .splash
   background-image: url('yunityorg.jpg')
   background-size: cover
+  heading
```
tiltec: I think we're safe to use `h1` here, the splash page content uses `h2`

anne27: Updated!

PR #378, opened 2017-03-05
added function isFull

Rian0702:


client/app/components/group/_pickupList/pickupListItem/pickupListItem.component.js
```diff
@@ -4,7 +4,7 @@ import "./pickupListItem.styl";
 
 let pickupListItemComponent = {
   bindings: {
-    data: "<",              // pickup info, including .isFull .isUserMember etc..
+    data: "<",              // pickup info, including  .isUserMember etc..
```
tiltec: I would include isFull in the data binding, to avoid using `parentCtrl` (I think that is bad style).
We could pass a `config` object that has all the settings for `pickupListItem`, e.g. isFull, isMember etc

client/app/components/group/_pickupList/pickupList.controller.js
```diff
@@ -59,6 +57,15 @@ class PickupListController {
     this.filterAndDisplayPickups();
   }
 
+/**
+ * checks if a pickup is already full
```
tiltec: nice documentation! We didn't do it so far, but maybe we should. Do you know how this style is called?

Rian0702: honestly this was auto completed by visual studio code .. but this is jsdoc i think. look here http://usejsdoc.org/tags-param.html

tiltec:
Example of a test. Can be added below `it("addPickupInfo get Store Info functionality"`

```js
    it("returns if pickup is full", () => {
      expect($ctrl.isFull({
        "collector_ids": [1,2,3],
        "max_collectors": 3
      })).to.equal.true;
      expect($ctrl.isFull({
        "collector_ids": [1,2],
        "max_collectors": 3
      })).to.equal.false;

      // can also be unset, then unlimited people can join
      expect($ctrl.isFull({
        "collector_ids": [1,2]
      })).to.equal.false;
    });
```

Issue #311, opened 2017-02-17
Use 'today' and 'tomorrow' instead of dates


alangecker:

Sliverriver:
is this meant for the pickup overview ?

tiltec:
I think yes. Or other locations where you want to have easy-to-read dates (e.g. history)

anne27:
Hey, so in "History", dates are currently printed as 
**Monday, March 6, 2017**
Should they be printed as
**Today**
**Yesterday**
or
**Today, March 6, 2017**
**Yesterday, March 5, 2017**

tiltec:
I would prefer 
**Today**
**Yesterday**
and then a tooltip (using md-tooltip) showing the full date. Does it seem good to you?

There's a nice library called moment.js that we already use for date formatting. It can also print the human-readable relative date in many languages: https://momentjs.com/docs/#/displaying/fromnow/

Maybe it makes sense to introduce this directive: https://github.com/urish/angular-moment
But it's not really active (last commit Dec 2016), so I would actually prefer using a moment.js-only solution.

D0nPiano:
I'm for the second version ("today, March 6") just because it's more precise and might prevent timezone confusions

anne27:
Yes, I guess time zones might be a problem. Team Starbucks creating a pull request to add "Today" and "Tomorrow" for ease of readability. :)

nicksellen:
I would use https://date-fns.org/ for this now :)

nicksellen:
Places where it could be implemented:
- wall
- history

PR #386, opened 2017-03-06
Using "Today" and "Tomorrow" Issue  #311

devishi:
Hey,
We have added "Today" and "Tomorrow" in addition to the dates for the pickup group updates (history).
:)


client/app/services/history/history.service.js
```diff
@@ -5,12 +5,17 @@ class HistoryService {
...
   list(filter) {
     return this.$http.get("/api/history/", { params: filter })
       .then((res) => {
         return res.data.results.map( (entry) => {
           entry.date = new Date(entry.date);
+           var utc = new Date().toISOString().substr(0,10); 
+          if (entry.date.toISOString().substr(0,10)==utc)
+            entry.status="Today, ";
+          else if (entry.date.toISOString().substr(0,2)==utc.substr(0,2) && entry.date.toISOString().substr(6,10)==utc.substr(6,10) && parseInt(entry.date.toISOString().substr(3,5),10)==parseInt(utc.substr(3,5),10)-1) 
+            entry.status="Yesterday, ";
```
D0nPiano: It'd be good to include translation support here with using something like `$translate('HISTORY.TODAY')` and adding it to the english translation file.

nicksellen:
I would suggest to use momentjs fromNow() for this https://momentjs.com/docs/#/displaying/fromnow/. And it already supports i18n. Why just today/tomorrow as special?

Also please keep codestyle the same - it should look like one person wrote it. You can run eslint locally as a standalone command or there are many editor plugins for it. It is your friend :)

anne27:
@nicksellen I guess what we're wanting to do here is have "Today" and "Tomorrow", and have the rest of the days written as they are right now (i.e. fullDateFormat). That is why we simply used the condition.

tiltec:
Feedback:

1. Translation needs to be possible 
2. Code location could be improved

## Moment.js with built-in translations

I agree with Lars and Nick, the messages should be translatable before I would merge.
The easiest option would be using the great `moment.js` library. I did a bit of research and there seem to be two built-in options, which are not exactly fitting.

### Calendar

```
moment(entry.date).calendar()
```
screen shot 2017-03-08 at 13 31 34

### fromNow

```
moment(entry.date).fromNow()
```
screen shot 2017-03-08 at 13 33 49

You can also customize those, read more in the documentation of moment.js. Maybe you can find a nice option? :)

## Code location

I would prefer to have the code in the history component controller (components/group/_history). We currently have almost no code in the backend-facing data services, so it would be unexpected to find it there. But would also listen for reasons why it should be otherwise :)

nicksellen:
Personally I like the fromNow() or calendar() (without the time bit). I only ever understand dates when specified relative to now. My mum always uses proper dates but I have to translate them first "aaah you mean in a week from now..."

nicksellen:
I think this is a good example of a task where it was too precisely specified. "Using today and tomorrow" is one way to implement something. The real task is something like "use human friendly relative dates" possibly with a justification too.

Never be afraid to widen the scope of a task if it seems appropriate.

tiltec:
You can remove the time bit from calendar by giving moment.js a custom format, but then you'd have to provide translations as well.

Maybe play with `fromNow()` and `from()` until it's good. I think we don't need to show a headline for today at all.

tiltec:
@devishi Do you know when you will have time to continue working on this?

devishi:
Hi, 
I am really sorry for delaying this. My exams were going on. I will start working on it again in a week's time.
Thanks

tiltec:
No worries, good luck for your exams!

tiltec:
I guess we can close this?

nicksellen:
Yes, I agree. Feel free to continue if you have some time again @devishi - there will be a lot of changes in master you'll have rebase onto though.

PR #390, opened 2017-03-07
fix: reverse default pickup sorting

### LINKED ISSUE #388: Pickup list: sorting isn't working correctly
Sometimes the pickup dates are shown in the wrong order. 6PM should appear before 8PM.

image

tiltec:
@D0nPiano it looks like the original code was intended. Do you remember why we had `-date` there? Does something break when we change this?

Closes #388


D0nPiano:
Does removing it reverse the pickuplist? At the moment, it's present -> future for me when you scroll down which makes sense to me.

tiltec:
`-date` means showing the pickups first that are far in future (see the linked issue, in this case showing pickups first that are later at a day). So it seems fine to change it :)

Issue #372, opened 2017-03-05
Refactor: pickupList


tiltec:
Main goal: make pickupList code simpler and make the page feel more snappy. 
How: reduce usage of the updatePickups() function, usually a reload from server is not necessary.

TODO:

- [x] use function to get pickup status instead of object keys, e.g. `isFull(pickup)`, as mentioned in #361 
- [x] use Promise feedback from $mdDialog and Pickup service to update list
- [x] call `filterAndDisplayPickups` from template (ng-repeat), rename appropriately
- [x] pickupListItem join/leave: use promise to update list (with output bindings instead of parentCtrl)
- [x] pickupListItem: load user details from a yet-to-create service that holds all users (or just the current groups' users)

PR #393, opened 2017-03-07
#372 - refactored dialog promise to not call update every time, added…

Rian0702:
… isMemberOf function

 !!!please check test "is called and updates pickup list"!!! 

#372

### DIFF
--- a/client/app/components/group/_pickupList/pickupList.controller.js
+++ b/client/app/components/group/_pickupList/pickupList.controller.js
@@ -182,8 +182,13 @@ class PickupListController {
       controller: DialogController,
       controllerAs: "$ctrl"
     }).then((data) => {
-      this.allPickups.push(data);
-      this.filterAndDisplayPickups();
+      if (data.start_date) {
+        // workaround: reload complete list if series was created
+        this.updatePickups();
+      } else {
+        this.allPickups.push(data);
+        this.filterAndDisplayPickups();
+      }
     });
   }
 }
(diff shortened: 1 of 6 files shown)


client/app/components/group/_pickupList/pickupList.spec.js
```diff
@@ -252,6 +242,16 @@ describe("PickupList", () => {
         "collector_ids": [1,2]
       })).to.equal.false;
     });
+
+    it("isUserMember functionality", () => {
```
tiltec: I love those simple tests :)

client/app/components/group/_pickupList/pickupList.controller.js
```diff
@@ -141,8 +147,19 @@ class PickupListController {
...
+      if (this.isDeleteSeries) {
+        this.allPickups = this.allPickups.filter((pickup) => {
+          return pickup.series !== this.pickupToDelete.series;
+        });
+      } else {
+        let index = this.allPickups.indexOf(this.pickupToDelete);
+        this.allPickups.splice(index, 1);
+      }
+      this.filterAndDisplayPickups();
+      this.isDeleteSeries = false;
```
tiltec: Good! It makes sense to restore this to default.

client/app/components/group/_pickupList/pickupList.controller.js
```diff
@@ -164,8 +181,9 @@ class PickupListController {
       },
       controller: DialogController,
       controllerAs: "$ctrl"
-    }).then(() => {
-      this.updatePickups();
+    }).then((data) => {
+      this.allPickups.push(data);
```
tiltec: I'll add a workaround for series.

PR #402, opened 2017-03-10
Feature/resend verification

andela-tisrael:
Hello, I have completed the task, I need you to review it so as to make the necessary correction to the PR. :)

### DIFF
--- a/client/app/components/_topbar/topbar.html
+++ b/client/app/components/_topbar/topbar.html
@@ -12,8 +12,8 @@
       </md-tooltip>
     </md-button>
 
-    <md-button ng-if="!screenIsSmall" class="userProfileLink" ui-sref="notification({id: $ctrl.loggedInUser.id, email: $ctrl.loggedInUser.email})">
-      <i class="fa fa-envelope-o"></i>
+    <md-button ng-if="!screenIsSmall" class="userProfileLink" ui-sref="notification">
+      <i class="fa fa-bell-o"></i>
       <md-tooltip>
           <span translate="TOPBAR.NOTIFICATION"></span>
       </md-tooltip>
--- a/client/app/components/_topbar/topbar.html
+++ b/client/app/components/_topbar/topbar.html
@@ -11,6 +11,14 @@
           <span translate="TOPBAR.USERPROFILE"></span>
       </md-tooltip>
     </md-button>
+
+    <md-button ng-if="!screenIsSmall" class="userProfileLink" ui-sref="notification({id: $ctrl.loggedInUser.id, email: $ctrl.loggedInUser.email})">
+      <i class="fa fa-envelope-o"></i>
+      <md-tooltip>
+          <span translate="TOPBAR.NOTIFICATION"></span>
+      </md-tooltip>
+    </md-button>
+
     <md-button ng-if="!screenIsSmall" class="md-icon-button" aria-label="{{ 'TOPBAR.LOGOUT' | translate }}" logout>
       <i class="fa fa-fw fa-sign-out"></i>
       <md-tooltip>
@@ -37,6 +45,12 @@
         <md-button ng-click="$ctrl.toggleRight()" class="userProfileLink" ui-sref="userDetail({id: $ctrl.loggedInUser.id})" aria-label="{{ 'TOPBAR.USERPROFILE' | translate }}">
           <i class="fa fa-user"></i>
           <span translate="TOPBAR.USERPROFILE"></span>
+        </md-button>
+      </md-menu-item>
+      <md-menu-item>
+        <md-button ng-click="$ctrl.toggleRight()" class="userProfileLink">
+          <i class="fa fa-envelope-o"></i>
+          <span translate="TOPBAR.NOTIFICATION"></span>
         </md-button>
       </md-menu-item>
       <md-menu-item>
--- a/client/app/components/notification/notification.controller.js
+++ b/client/app/components/notification/notification.controller.js
@@ -1,5 +1,5 @@
 class NotificationController {
-  constructor(User, Authentication, $state, $injector) {
+  constructor(User, Authentication, $state, $injector, $translate) {
     "ngInject";
     Object.assign(this, {
       User,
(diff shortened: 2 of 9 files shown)


client/app/services/authentication/authentication.service.js
```diff
@@ -21,6 +21,12 @@ class AuthenticationService {
       .catch((res) => this.$q.reject(res.data));
   }
 
+  verify(userEmail) {
```
tiltec: This should be renamed appropriately, to `resendVerificationRequest` or similar.

client/app/components/notification/notification.html
```diff
@@ -0,0 +1,27 @@
+<div layout="row">
+  <div flex></div>
+
+<div id="notification" class="md-whiteframe-2dp" style="margin: 0; padding: 0">
+  <md-toolbar class="md-hue-1">
+    <div class="md-toolbar-tools">
+      <h1 flex>
+        <i class="fa fa-user" aria-hidden="true"></i>&nbsp;
+        {{ $ctrl.userdata.display_name }}
```
tiltec: Maybe add a message here to make clear that the user is on the notifications page.

client/app/components/notification/notification.js
```diff
@@ -0,0 +1,36 @@
...
+      parent: "main",
+      url: "/notification/user/{id:int}",
+      component: "notification",
+      resolve: {
+        userdata: (User, $stateParams) => {
+          return User.get($stateParams.id).then((user) => user);
+        }
+      },
+      ncyBreadcrumb: {
+        label: "{{$$childHead.$ctrl.userdata.display_name}}"
```
tiltec: This is what will show up in the browser title bar. I would set it "Notifications" (but translated).

client/app/components/notification/notification.js
```diff
@@ -0,0 +1,36 @@
...
+])
+
+.component("notification", notificationComponent)
+
+.config(($stateProvider, hookProvider) => {
+  "ngInject";
+  $stateProvider
+    .state("notification", {
+      parent: "main",
+      url: "/notification/user/{id:int}",
```
tiltec: Could you nest the notifications page in the user state?
So that the URL looks like `/user/{id:int}/notifications`? Then you also don't need to resolve the user data again.

You can do this by setting `parent` appropriately.

client/app/components/notification/notification.html
```diff
@@ -0,0 +1,27 @@
...
+      </h1>
+    </div>
+  </md-toolbar>
+  <div class="content">
```
tiltec: It would be nice to have uniform indentation. Do you have `editorconfig` plugin for your editor/IDE? It will set 2 spaces of indentation per opened tag and other settings (more info in CONTRIBUTE.md)

client/app/services/authentication/authentication.service.js
```diff
@@ -21,6 +21,12 @@ class AuthenticationService {
...
+    return this.$http.post("/api/users/resend_verification/", { email: userEmail })
```
tiltec: You don't need to provide the users' mail or any other data. (You can also remove that part in the controller)

client/app/components/notification/notification.html
```diff
@@ -0,0 +1,27 @@
...
+      </span>
+      <span ng-show="$ctrl.isVerified()">
+      <span style="float: right;">Verified</span>
+      <i class="fa fa-check-circle fa-2" style ="float: right; color: green;" aria-hidden="true"></i> 
+      </span>
+      </h3>
+  </div>
+</div>
+  <div flex></div>
+  <!--ng-click="$ctrl.sendVerification($ctrl.userdata.email)"-->
```
tiltec: You can remove that comment ;)

client/app/components/notification/notification.controller.js
```diff
@@ -0,0 +1,36 @@
...
+  isVerified() {
+    return this.userdata.mail_verified;
+  }
+  sendVerification(email) {
+    const $mdToast = this.$injector.get("$mdToast");
+    this.Authentication.update().then((data) => {
+      if (data.mail_verified === false) {
+        this.Authentication.verify(email).then(() => {
+          $mdToast.show($mdToast.simple()
+            .textContent("Verification Email Sent")
```
tiltec: Make all strings translatable (probably using the translate service or filter here).

client/app/components/notification/notification.controller.js
```diff
@@ -0,0 +1,36 @@
...
+            .highlightAction(false));
+          return false;
+        }).catch(() => {
+          $mdToast.show($mdToast.simple()
+            .textContent("Email does not Exist")
```
tiltec: There are three cases where this could happen:

- connection interrupted
- server error
- mail already verified

The first two are already handled in a $http interceptor (see `app.HTTPErrorHandler.js` and the last one is not expected.

I would either leave this away or just forward the error message of the server:

```
catch((err) => {
...
.textContent(err)
...
}
```

It will probably be a ugly message, but it shouldn't happen at all ;)

client/app/components/notification/notification.controller.js
```diff
@@ -0,0 +1,36 @@
...
+    });
+  }
+
+  isVerified() {
+    return this.userdata.mail_verified;
+  }
+  sendVerification(email) {
+    const $mdToast = this.$injector.get("$mdToast");
+    this.Authentication.update().then((data) => {
+      if (data.mail_verified === false) {
```
tiltec: I think you can leave away this check. The button should not be clickable in this case, so checking again seems to be too much.

client/app/components/notification/notification.controller.js
```diff
@@ -0,0 +1,36 @@
...
+      $state,
+      $injector
+    });
+  }
+
+  isVerified() {
+    return this.userdata.mail_verified;
+  }
+  sendVerification(email) {
+    const $mdToast = this.$injector.get("$mdToast");
```
tiltec: Is there a reason why you inject $mdToast here? I would just inject it in the constructor (analog to `$state` etc)

client/app/components/_topbar/topbar.html
```diff
@@ -40,6 +48,12 @@ <h1 class="md-toolbar-tools"><span translate="TOPBAR.GREETING" translate-values=
         </md-button>
       </md-menu-item>
       <md-menu-item>
+        <md-button ng-click="$ctrl.toggleRight()" class="userProfileLink">
+          <i class="fa fa-envelope-o"></i>
+          <span translate="TOPBAR.NOTIFICATION"></span>
```
tiltec: Great!

client/app/services/authentication/authentication.service.js
```diff
@@ -21,6 +21,12 @@ class AuthenticationService {
...
+      .then((res) => res.data)
+      .then(() => this.data = "Email Sent");
```
tiltec: Is this used? I wouldn't set (translatable) messages in this backend-facing service, but instead in the controller.

PR #403, opened 2017-03-11
Update store map by equality watching

tiltec:
Problem: store map didn't update after an added store or address change.

I tried several patterns, but this seems to be the best compromise right now.

- $state.go with `reload: true` showed an empty map
- $doCheck needs some dirty checking, needs much code or workaround (e.d. using a version number)
- other notify patterns are quite obstructive (listeners, event broadcasting)

This should mostly work, we just need to remember this when we get into frontend performance troubles.

### DIFF
--- a/client/app/components/group/_storeList/storeList.controller.js
+++ b/client/app/components/group/_storeList/storeList.controller.js
@@ -1,10 +1,9 @@
 class StoreListController {
-  constructor(Store, $state, $document, $mdMedia, CurrentStores, $scope) {
+  constructor(Store, $state, $document, $mdMedia, CurrentStores) {
     "ngInject";
     Object.assign(this, {
       Store,
       CurrentStores,
-      $scope,
       storeList: CurrentStores.list,
       $state,
       $document,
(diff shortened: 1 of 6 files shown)


client/app/components/group/_storeList/storeList.controller.js
```diff
@@ -17,7 +17,6 @@ class StoreListController {
     if (changes.groupId && angular.isDefined(changes.groupId.currentValue)) {
       this.Store.listByGroupId(changes.groupId.currentValue).then((data) => {
         this.CurrentStores.set(data);
-        this.storeList = angular.copy(this.CurrentStores.list);
```
tiltec: @nicksellen I removed this line already...

nicksellen: Right, although I was just looking at the store list not updating, not the map bit. Why not just get the `CurrentStores` service in the store map rather than pass as binding?

tiltec: The only reason I could imagine is to keep storeMap reusable, but that's not important at all.
So yes, I would remove the binding. The watch is still necessary, though.

Issue #163, opened 2016-12-02
Testing of templates


tiltec:
Currently we only test controllers, but we should also check templates.

- Do they get compiled with all directives/components?
- No warnings should appear in the JS console
- test if some important elements exist in the page

Some ideas about this: http://daginge.com/technology/2013/12/14/testing-angular-templates-with-jasmine-and-karma/

How to get the template compiled: https://github.com/yunity/foodsaving-frontend/blob/master/client/app/components/_storeList/_storeMap/storeMap.spec.js#L21

tiltec:
I tried it yesterday, it didn't work. When compiling the template, we need angular-translate incl config in each of the components, and maybe other stuff that we rely on being loaded in grandparent app.js

Of course this means that our components are not really decoupled, but it seems a bit too complex to do...

PR #407, opened 2017-03-12
Fix missing aria-label in storeList component

### LINKED ISSUE #406: aria-label warning causes test failure
Maybe just because I have super new chromium my tests fail with:

```
FAILED TESTS:
  StoreList
    ✖ "after each" hook for "gets store data"
      Chrome 56.0.2924 (Linux 0.0.0)
    Error: Expected $log to be empty! Either a message was logged unexpectedly, or an expected log message was not checked and removed:
    ---------
    MOCK $log (warn): ARIA: Attribute "
    
    ---------
    MOCK $log (warn): aria-label
    
    ---------
    MOCK $log (warn): ", required for accessibility, is missing on node:
    
    ---------
    MOCK $log (warn): [object HTMLButtonElement]
    
    ---------
    
        at Object.$log.assertEmpty (webpack:///~/angular-mocks/angular-mocks.js:420:0 <- spec.bundle.js:33410:16)
        at Context.<anonymous> (webpack:///client/app/components/group/_storeList/storeList.spec.js:13:9 <- spec.bundle.js:140638:11)
```

... fix incoming shortly ...

nicksellen:
Was causing test to fail

Fixes #406

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -33,6 +33,7 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/).
 - Disable autoformatting features on login and signup page
 - Disable buttons when a request is in progress
 - Mail change now happens after the new mail is verified
+- Fix missing aria-label causing test to fail
 
 
 ## [Release 2] - 2017-02-21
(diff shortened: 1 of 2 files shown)


tiltec:
Good, then the assertion did what it should do :)
The test runs `$compile`, which triggers the log warnings. I'd really like to compile all templates on testing #163 but we need to find a solution for the translations first (more in the linked issue).

PR #410, opened 2017-03-12
Fix store list not updating

nicksellen:
First addition is a failing test...

### DIFF
--- a/client/app/components/group/_storeList/storeList.spec.js
+++ b/client/app/components/group/_storeList/storeList.spec.js
@@ -34,13 +34,12 @@ describe("StoreList", () => {
   };
 
   describe("Component", () => {
-    let $scope, $ctrl, $httpBackend, CurrentStores;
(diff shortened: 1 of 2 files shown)


client/app/components/group/_storeList/storeList.spec.js
```diff
@@ -23,13 +23,24 @@ describe("StoreList", () => {
...
+    "description": "all the good stuff again",
+    "group": 1,
+    "address": null,
+    "latitude": null,
+    "longitude": null
+  };
+
   describe("Component", () => {
-    let $scope, $ctrl, $httpBackend;
+    let $scope, $ctrl, $httpBackend, CurrentStores;
```
tiltec: In #405 I changed all tests to use injected services from the controller if possible.
In line 67, you can just append a `$ctrl` before CurrentStores and remove the injection here.

PR #420, opened 2017-03-21
372 refactor pickup list

### LINKED ISSUE #372: Refactor: pickupList

Rian0702:
Closes #372 

had to insert ng-if for md-button to make hide and show working.


client/app/components/group/_pickupList/pickupList.controller.js
```diff
@@ -191,6 +191,23 @@ class PickupListController {
       }
     });
   }
+
+  joinPickup(pickupId) {
```
tiltec: Can you move this logic completely into pickupListItem? The `collector_ids` can also be changed from there.
Also, have a look at the JS Array.find() function, it can simplify this.

client/app/components/group/_pickupList/pickupList.controller.js
```diff
@@ -191,6 +191,23 @@ class PickupListController {
...
+    let pickupToJoin = this.$filter("filter")(this.allPickups, { id: pickupId });
```
tiltec: The "filter" filter will go away in Angular2+, I wonder if we should avoid using it...

client/app/components/group/_pickupList/pickupList.controller.js
tiltec: I came to the conclusion that it's good to have this here (and not move it to pickupListItem). So we can leave it.

tiltec:
@Rian0702 Are you still interested in working on this?

tiltec:
I will finish this PR now

PR #441, opened 2017-04-10
Change Store/Group location by clicking on map #391

### LINKED ISSUE #391: Map: Add store by clicking and improve zoom
Right now one has to enter an address to add a store location. It would be much more intuitive and flexible if adding a store location by clicking on the map was possible.
When opening the map to enter a store location by clicking, the default zoom should be relative to the groups location.

- [ ] enable adding store locations by clicking
- [ ] localize default zoom level

D0nPiano:
Closes #391


client/app/app.js
```diff
@@ -70,6 +70,10 @@ angular.module("app", [
 .run(AppLocalizeRun)
 .config(AppMaterial)
 .config(AppHTTPErrorHandler)
+.config(( $mdGestureProvider ) => {
```
D0nPiano: I'm not quite sure if tinkering with the mdGestureProvider is a good idea - but otherwise clicks are not working on mobile maps.

djahnie:
Ooooh amazing! 😃 
Thx @D0nPiano for tackling this! 😗

client/app/app.js
tiltec: Is there documentation on this? I found https://github.com/angular/code.material.angularjs.org/blob/master/HEAD/partials/api/material.core.gestures/service/$mdGestureProvider.html

I guess the touch event is intercepted by $mdGesture and does not reach leaflet. But is there another event emitted that could be used by leaflet?

D0nPiano: Well, I found https://github.com/angular/material/issues/1300#issuecomment-98498251  - so apperently this destorys certain swipe features of angular material, but I think having a working map is more important. Will not only be interesting for this issue, but also for clickable popups etc.

tiltec: I agree, we can think about this more when we need md-swipe.

client/app/components/group/store/_storeEditCreateForm/storeEditCreateForm.controller.js
```diff
@@ -1,22 +1,45 @@
...
     });
+
+    $scope.mapController = this;
+    $scope.$on("leafletDirectiveMap.click", (event, e) => {
+      let item = {
+        latitude: e.leafletEvent.latlng.lat,
+        longitude: e.leafletEvent.latlng.lng
+      };
+      $scope.mapController.marker = {
+        p: {
```
tiltec: Repeated code, maybe move it into a function?

client/app/components/group/store/_storeEditCreateForm/storeEditCreateForm.controller.js
```diff
@@ -1,22 +1,45 @@
...
       mapDefaults: {
         scrollWheelZoom: false,
-        zoomControl: false,
-        dragging: false
+        zoomControl: true,
+        dragging: true
       }
     });
+
+    $scope.mapController = this;
```
tiltec: Is this necessary? `this` is already bound to `$scope.$ctrl`.

PR #446, opened 2017-04-10
Start testing compilation of templates

tiltec:
I started integrating the compilation of templates into our tests. It's not done yet, but would be happy about feedback.

Advantages: it should be possible to discover missing directives, aria-labels...


tiltec:
I will try to merge this branch later today, even if the test coverage measured in angular templates is not complete yet.

Issue #382, opened 2017-03-05
History filtering/date-browsing


tiltec:
The history API returns the latest 50 results per default: https://github.com/yunity/foodsaving-backend/pull/260

We should give the user the opportunity to browse through the history and filter by action type (`typus`).
Use swagger ( https://foodsaving.world/docs/#!/history/history_list ) to see the filtering options and the API results.

Update: we now can filter by user and store.

To do:
- attach history to store view (only showing store-related events)
- enable browsing by date (e.g. load history events before a point in time)

Sliverriver:
does this also include filter by date ?

tiltec:
Yes, that's what I mean with browsing. Maybe a date picker to jump to a certain point in time.
The backend could be changed to offer date filters, that could make it easier.

Sliverriver:
i am a bit lost where to comment for this request - if it's a new issue or improvement for this one. The history looks really nice on the frontend. I worry that it can get quite cluttered where it is now. has there been discussion if the history could be attached to a store ? Like the same tab just on a store level to switch to ?

tiltec:
Yes, that would also be possible. Note to implementers: history is a stateless component, you can just feed it any list of history items.

tiltec:
To do: jump to a point in time.

tiltec:
The new history lost all options for filtering currently, should we add it again? Which ones are the most useful?

![image](https://user-images.githubusercontent.com/4410802/32144349-cdc6bfd0-bcb7-11e7-91dc-cea7842b6511.png)

(Will move it back to Discussion)

tiltec:
It would be nice to have/cool, but there wasn't a _real_ need for it so far. You can visit group/store/user history right now, which is fairly powerful already.

tiltec:
Can be reopened if requested.

PR #449, opened 2017-04-10
History filtering #382

D0nPiano:
ToDos:
- [ ] Test

Preview:
screenshot from 2017-04-10 21-29-45


tiltec:
Cool! Should we add a history tab to the stores as well?

client/app/components/group/_history/history.controller.js
```diff
@@ -1,9 +1,112 @@
...
+  }
+
+  showAllStores(bool){
+    angular.forEach(this.stores, (store) => {
+      store.selected = bool;
+    });
+    this.updateFilteredData();
+  }
+
+  getAllStores(){
```
tiltec: You could get rid of this function by using the values from the CurrentStores service.

client/app/components/group/_history/history.controller.js
```diff
@@ -1,9 +1,112 @@
...
     });
   }
+
+  $onInit(){
+    this.getAllStores();
+    this.getAllUsers();
+    this.updateFilteredData();
+  }
+
+  update(){
```
tiltec: I would try use the angular digest cycle more, e.g. having a function that gets called in the template `getEntries()` and that function returns filtered history data. Advantage: less explicit data management code.

client/app/components/group/_history/history.html
```diff
@@ -1,5 +1,48 @@
...
+            <md-menu-divider></md-menu-divider>
+            <md-menu-item type="checkbox" ng-click="$ctrl.update()" ng-model="user.selected" ng-repeat="user in $ctrl.users">{{user.display_name}}</md-menu-item>
+          </md-menu-content>
+        </md-menu>
+        <md-menu>
+          <button ng-click="$mdMenu.open()">
+            Type
+          </button>
+          <md-menu-content>
+            <md-menu-item type="checkbox" ng-click="$ctrl.update()" ng-model="$ctrl.types.groups">{{'HISTORY.TYPES.GROUPS' | translate}}</md-menu-item>
```
tiltec: It is recommended to use the `translate` directive over filters to reduce the watches.
`<span translate="HISTORY.TYPES.GROUPS"></span>`

tiltec:
I'll work on this now

tiltec:
TODO:
(in addition to tests)

- make user/store list scrollable
- make user/store list searchable (it can get long)

client/app/components/group/_history/history.controller.js
```diff
@@ -49,6 +60,77 @@ class HistoryController {
...
+      if (angular.isUndefined(user._selected)) {
+        user._selected = true;
+      }
+      return user;
+    });
+  }
+
+  showAllStores(bool) {
+    angular.forEach(this.getStores(), (store) => {
+      store._selected = bool;
```
tiltec: I don't like this, it changes the values in CurrentStores. But everything else seems to involve much more code (e.g. merging together two lists)

nicksellen: I would keep selected state in it's own object, something like:

```js
var selectedStores = {};

// select a store
selectedStores[id] = true;

// unselect a store
delete selectedStores[id]

// check if a store is selected
selectedStores[id]
```

tiltec: How does that go together with `ng-model`? E.g. here https://github.com/yunity/foodsaving-frontend/pull/449/files#diff-1491a792156b8840cb4c659ffb0de4ddR14

I'm not a huge fan of using getters/setters there, too much boilerplate.

nicksellen:
```html
ng-model="selectedStores[store.id]"
```
Maybe?

Sometimes have to use the `ng-click` on a checkbox instead.

nicksellen: There are much worse things than boilerplate too.

tiltec: Good idea, will refactor now.
(this feature here is not urgent at all, but I wanted to get it merged before the code around it changes too much.. might need to rething on my priorities though)

tiltec:
Didn't manage to enable scrolling in the user list, but otherwise it's ready for merging.

PR #450, opened 2017-04-12
rename history to activity

varu038:


tiltec:
I would close this, as most of the refactoring work is actually on the back-end side and there's currently no one interested in doing this, right?

nicksellen:
I agree, has no activity for a long time. I still like the idea, but should track it in an issue and see when someone has time to work on it.

Issue #218, opened 2016-12-12
Persist currentGroup


tiltec:
## Problem

when opening a user page directly, the group is not set.

## Solution

Let's store the currentGroup on the server.

Retrieve it when the page opens. Save it when you open another group. Could be embedded in the currentGroupService.

## Other solution

Simpler would be to just use localstorage/cookie.

tiltec:
Associated backend task: https://github.com/yunity/foodsaving-backend/issues/277

ShababShahriar:
Is this still open for contribution? Would like to work on it if no one else is.

tiltec:
Backend is already done https://github.com/yunity/foodsaving-backend/pull/286 , needs frontend implementation.

ShababShahriar:
Up for it.

tiltec:
Is this information sufficient for you?

> Retrieve it when the page opens. Save it when you open another group. Could be embedded in the currentGroupService.

When using the currentGroup service, I assume it will only be a few lines of additional code there.

tiltec:
Code locations:

Setting the current group: https://github.com/yunity/foodsaving-frontend/blob/master/client/app/components/group/group.js#L60

Switching to the first group after login: https://github.com/yunity/foodsaving-frontend/blob/master/client/app/components/home/home.controller.js#L15

Call the backend to modify user details: https://github.com/yunity/foodsaving-frontend/blob/master/client/app/services/user/user.service.js#L29
(is using PATCH, so you only need to send the changed data: the user id and the current_group id)

Service to save the current group in the browser
https://github.com/yunity/foodsaving-frontend/blob/master/client/app/services/group/currentGroup.service.js
could be used to call the User.save function when the set() function is called (needs to get the user id from the Authentication.update method)

Auth update
https://github.com/yunity/foodsaving-frontend/blob/master/client/app/services/authentication/authentication.service.js#L17
returns the user details if the user is logged in

tiltec:
It works, we should refactor the code locations in future though.

PR #471, opened 2017-04-23
Persist current group of user

ShababShahriar:
Connected to Issue #218.

A few (date/numbers related) tests are failing locally but I can't figure out how they may be affected by my changes. Also, lint considers snake-cased variable `current_group` as an error but this is necessary for the backend code to work. Bypassing linting in this case could be an option but open to a second opinion.

### DIFF
--- a/client/app/components/group/group.spec.js
+++ b/client/app/components/group/group.spec.js
@@ -2,6 +2,7 @@ import GroupDetailModule from "./group";
 import GroupDetailController from "./group.controller";
 import GroupDetailComponent from "./group.component";
 import GroupDetailTemplate from "./group.html";
+import CurrentGroup from "../../services/group/group";
 
 const { module } = angular.mock;
 
@@ -9,6 +10,7 @@ describe("Group", () => {
   let $httpBackend, $state;
 
   beforeEach(module(GroupDetailModule));
+  beforeEach(module(CurrentGroup));
   beforeEach(module({ $translate: sinon.stub() }));
   beforeEach(module(($stateProvider) => {
     $stateProvider
@@ -65,8 +67,10 @@ describe("Group", () => {
   describe("Route", () => {
     beforeEach(() => {
       $httpBackend.whenGET("/api/auth/status/").respond({ id: 43 });
-      inject(($translate, $q) => {
+      inject(($translate, $q, $injector) => {
         $translate.returns($q.resolve());
+        let CurrentGroup = $injector.get("CurrentGroup");
+        sinon.stub(CurrentGroup, "set");
       });
     });
 
--- a/client/app/components/group/group.spec.js
+++ b/client/app/components/group/group.spec.js
@@ -65,9 +65,8 @@ describe("Group", () => {
   describe("Route", () => {
     beforeEach(() => {
       $httpBackend.whenGET("/api/auth/status/").respond({ id: 43 });
-      inject(($translate, $q, $injector) => {
+      inject(($translate, $q, CurrentGroup) => {
         $translate.returns($q.resolve());
-        let CurrentGroup = $injector.get("CurrentGroup");
         sinon.stub(CurrentGroup, "set");
       });
     });
--- a/client/app/components/group/group.spec.js
+++ b/client/app/components/group/group.spec.js
@@ -2,7 +2,6 @@ import GroupDetailModule from "./group";
 import GroupDetailController from "./group.controller";
 import GroupDetailComponent from "./group.component";
 import GroupDetailTemplate from "./group.html";
-import CurrentGroup from "../../services/group/group";
 
 const { module } = angular.mock;
 
@@ -10,7 +9,6 @@ describe("Group", () => {
   let $httpBackend, $state;
 
   beforeEach(module(GroupDetailModule));
-  beforeEach(module(CurrentGroup));
   beforeEach(module({ $translate: sinon.stub() }));
   beforeEach(module(($stateProvider) => {
     $stateProvider
--- a/client/app/components/home/home.controller.js
+++ b/client/app/components/home/home.controller.js
@@ -1,22 +1,32 @@
 class HomeController {
-  constructor($state, $document, $mdDialog, GroupService) {
+  constructor($state, $document, $mdDialog, GroupService, Authentication) {
     "ngInject";
     Object.assign(this, {
       $state,
       $document,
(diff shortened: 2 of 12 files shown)


client/app/services/group/currentGroup.service.js
```diff
@@ -9,18 +9,39 @@
...
+      value: {},
+      User,
+      Authentication
     });
   }
 
   set(value) {
     angular.copy(value, this.value);
+    if (angular.isUndefined(this.Authentication.data)) {
+      this.Authentication.update();
```
tiltec: This is wrong. The update() method sets `Authentication.data` asynchronously. If the authentication data does not exist, `persistCurrentGroup` will use the undefined value before it gets updated.

I think we can assume that setting the current group is only possible when a user is logged in. Therefore, the call to Authentication.update() should not be necessary.
At the maximum, I would check in persistCurrentGroup that `Authentication.data` is defined and throw an error if not. But I'm also happy to leave this check away.

client/app/components/home/home.controller.js
```diff
@@ -1,22 +1,32 @@
...
   $onInit() {
-    this.GroupService.listMy().then((data) => {
-      if (data.length > 0) {
-        this.$state.go("group", { groupId: data[0].id });
-      } else {
-        this.openJoinGroupDialog();
-      }
-    });
+    if (angular.isUndefined(this.Authentication.data)) {
+      this.Authentication.update();
```
tiltec: Same problem here - remember that Authentication.update() returns a promise. The code that follows would need to be run in a `then` function.

But we can safely assume that `Authentication.data` is defined - the `hookProvider` checks that (and redirects to the login screen if the user is not logged in). There is no race condition there: the state only gets loaded when the [hookProvider did its job](https://github.com/yunity/foodsaving-frontend/blob/b840c3dfc6f06181603f676d4459ca9881aad71e/client/app/services/authentication/hook.js#L20)

tiltec:
That is a great start in frontend coding! I guess you've seen many parts of the codebase and I'm impressed by the number of test modifications :)
I think you can code it a bit different, though - usually, you don't need to load modules in the test, but you should add them to the module requirements (e.g. in `groupMenu.js`)

For the snake_case variable: that's quite unfortunate. This problem persists throughout our codebase and the best solution right now is to ignore the linter errors with `eslint-ignore-line`. As a better solution, I would consider switching the API to camelCase. We had a discussion about this somewhere in an issue.

client/app/services/group/currentGroup.service.js
ShababShahriar: I wasn't sure if I should put this check since it's handled during the routing, and was hoping you'd comment on this. Thanks, removed it in the new commit.

client/app/components/home/home.controller.js
ShababShahriar: This is good learning though I'll admit that the hookProvider coding seems a bit complex to wrap my head around.

client/app/components/group/group.spec.js
```diff
@@ -65,8 +65,10 @@ describe("Group", () => {
   describe("Route", () => {
     beforeEach(() => {
       $httpBackend.whenGET("/api/auth/status/").respond({ id: 43 });
-      inject(($translate, $q) => {
+      inject(($translate, $q, $injector) => {
```
tiltec: Can't you simply write `inject(($translate, $q, CurrentGroup) => {` here instead of the verbose `injector.get` syntax?

client/app/services/group/currentGroup.spec.js
```diff
@@ -14,36 +14,61 @@ describe("CurrentGroup service", () => {
...
 
   it("can be set", () => {
     let newGroup = {
+      id: 1,
       anything: "can",
       go: "here",
       including: ["lists", "of", "stuff"]
     };
     CurrentGroup.set(newGroup);
+    assert(stub.withArgs(newGroup.id).calledOnce);
```
tiltec: Style thing: we only have `expect` in our codebase, maybe should stick to it for consistence? [using chai](https://github.com/domenic/sinon-chai)
`expect(CurrentGroup.persistCurrentGroup).to.have.been.calledOnce`

Also, I find it better to not introduce new references for the stub: `CurrentGroup.persistCurrentGroup` instead of `stub`

tiltec:
There's one failing test in `storeDetail.spec.js`, which is caused by the `CurrentGroup.set` in `group.js` - bad decoupling of those "unit" tests. I think the `$state.go` first loads the parent states (-> `group`), which calls `CurrentGroup.set`, which fails in turn because of missing authentication data.

I'll push a simple stub of the set method.

nicksellen:
There might be some interplay between the tests because you are not using sinon sandboxes [1] - they ensure that after the test is complete, all the sinon stuff is undone. But on the other hand, maybe this is not a problem in angular as the dependencies get injected fresh each time?

[1] http://sinonjs.org/releases/v2.2.0/sandbox/

nicksellen:
I'm very tired right now, so just having a quick flick through, a couple of points:
- the idea for `CurrentGroup` was to have it as a data-only service, this adds behaviour to it, not clear how it differs from `GroupService` now
- seems easy for server/client to get out of step, I switch to group A (server==client), then in another tab/window/device select group B, my original client is out of step now (although it doesn't currently seem to update the data in `Authentication` anyway) - seems it's more a _default_ group, or _last_ group or something...

I don't think I understood the initial requirement though:
> when opening a user page directly, the group is not set.
What is a user page?

client/app/services/group/currentGroup.spec.js
ShababShahriar: Valid points.

client/app/components/group/group.spec.js
ShababShahriar: Wasn't really familiar with the syntax, makes more sense now.

tiltec:
> the idea for CurrentGroup was to have it as a data-only service, this adds behaviour to it, not clear how it differs from GroupService now

Agree, maybe we should move the persist method to the User service and call it explicitly from the locations where `CurrentGroup.set` is called?

> What is a user page?

userDetail. the main problem is that the home screen chooses the first response of the server as the users' selected group.

tiltec:
I'd be in favour of merging now and leave the mentioned changes for a later refactor. Awaiting your feedback on this @nicksellen

PR #482, opened 2017-05-07
Fixes #462 Refactor logout directive into controller

### LINKED ISSUE #462: Refactor logout directive into controller
The `logout` directive is a leftover from early development. It's defined in [home.logout.js](https://github.com/yunity/foodsaving-frontend/blob/master/client/app/components/home/home.logout.js) and used in [topbar.html](https://github.com/yunity/foodsaving-frontend/blob/master/client/app/components/_topbar/topbar.html#L61). This is confusing and differs from our codestyle.

Move the functionality into the `topbar` controller.

tsmrachel:
https://github.com/yunity/foodsaving-frontend/issues/462


client/app/components/_topbar/topbar.controller.js
```diff
@@ -24,6 +25,15 @@ class TopbarController {
     this.$mdSidenav("right").toggle();
   }
 
+  logOut(){
+    this.Authentication.logout()
+      .then(() => {
+        this.$location.hash("/login");
```
tiltec: Code style: we always used ui-router for state transitions. Is there a reason why you chose this instead of `$state.go("login")`?

tsmrachel: Actually, i'm just not very familiar with angularJS. Appreciate the edit to include the suggested method 😄

tiltec:
Are you interested in adding a test as well? There wasn't one before, but now it's easier to write one :)

I would stub the Authentication.update method and test if the logout method successfully redirects to the login page. A similar test is in the `login.spec.js` in line 65, but this one is using a mock http backend instead of a stub.

tiltec:
Perfect! Will merge now!

PR #490, opened 2017-05-09
Disable ability to select a date that is in the past

jbforth4:
Added js function to only allow the current date and future dates to be selected in the pickupEditCreate folder.  Also updated the html file to use the js function.

### DIFF
--- a/client/app/components/group/_pickupEditCreate/pickupEditCreate.controller.js
+++ b/client/app/components/group/_pickupEditCreate/pickupEditCreate.controller.js
@@ -21,6 +21,16 @@ class pickupEditCreateController {
       isSeries: this.data.series,
       mode: this.data.series ? "series" : "single"
     });
+    
+     setMiniDate() {
+       myDate = new Date();
+
+       this.minDate = new Date(
+       this.myDate.getFullYear(),
+       this.myDate.getMonth(),
+       this.myDate.getDate()
+       );
+     }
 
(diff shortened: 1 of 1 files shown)


client/app/components/group/_pickupEditCreate/pickupEditCreate.controller.js
```diff
@@ -137,4 +137,14 @@ class pickupEditCreateController {
   }
 }
 
+  noPastDateCalender() {
+  this.myDate = new Date();
```
tiltec: better call this `now`?
Also, you just use this in local scope, so you don't need to define it on `this`.

client/app/components/group/_pickupEditCreate/pickupEditCreate.controller.js
```diff
@@ -137,4 +137,14 @@ class pickupEditCreateController {
   }
 }
 
+  noPastDateCalender() {
```
tiltec: maybe rather 'setMinDate'? It describes more accurately what the code is doing.

Also, you need to invoke this method somewhere. I'd recommend the $onInit method (just append it)

tiltec:
Does it make sense to you? It doesn't work, can you guess why?

jbforth4:
After the changes you made I was able to compile but the test environment
still won't launch. I was up messing around with it last night so without
being able to test and see the errors it it's hard to guess

tiltec:
Solved via #517

PR #495, opened 2017-05-15
Fixes #377 Group create/edit & store create/edit: check if name alrea…

### LINKED ISSUE #377: Group create/edit & store create/edit: check if name already exists
Groups have unique names. Stores have unique names within their groups. The editCreate forms should prevent form submission if the name already exists and show an error message.

To be added to the Group and Store services:

- Filter groups by name via `/api/groups/?name=abc`
- Filter stores in group 45 by name via `/api/stores/?group=45&name=abc`
- tests

To be added to the editCreate forms and controllers:

- asynchronous validator (maybe as directive for ngInput)
- translatable error message with `ng-message`
- text string to `locale-en.json`
- tests

Maybe in two Pull Requests: first the service, then the form/controller modification.

tsmrachel:
…dy exists

https://github.com/yunity/foodsaving-frontend/issues


tsmrachel:
@tiltec what's the best way to resolve the below? The logic works, but i'd like to get a preliminary review and some guidance on how to resolve this issue before i continue working on the rest of the tests. 

If I use the array syntax to inject dependencies, the linter complains 

[`.directive("groupnameValidator", [GroupService, "$q", (GroupService, $q) => {`](https://github.com/yunity/foodsaving-frontend/pull/495/files#diff-2d3107a266af6a260ca9922a46026704R21) - 'You should use the function syntax for DI'


If I use the function syntax for DI, i get an error

`.directive("groupnameValidator", (GroupService, $q) => {` - 'Error: [$injector:strictdi] function(GroupService, $q) is not using explicit annotation and cannot be invoked in strict mode'

tiltec:
Good question! We are using ng-annotate, have a look around in our code base for the `ngInject` strings that appear to have no functionality.

client/app/components/group/_groupEditCreateForm/groupEditCreateForm.spec.js
```diff
@@ -86,6 +86,60 @@ describe("GroupEditCreateForm", () => {
...
+        el = angular.element(
+        "<input type='text' name='name' ng-model='$ctrl.data.name' groupname-Validator/>"
+        );
+        $compile(el)(scope);
+      });
+      ngModel = el.controller("ngModel");
+    });
 
+    it("should reject if group name already exists", () => {
+      sinon.stub(GroupService, "listByGroupName").returns($q((resolve) => {
```
tiltec: I like the command chain here, seems slightly more readable than what I wrote in the other tests. Will use this style in future.

Issue #474, opened 2017-04-28
Make pickup manage page easier to discover


tiltec:
There were some reports that it's not intuitive to change the pickup settings, which is only possible on the manage pickup page.

As a solution, the prominent "create pickup" button could be replaced with a "manage pickup dates" button.

Also, I would recommend to allow deleting pickup dates only on the manage pickup page, not on the store page. It's not a commonly needed action.

MonsieurShaik:
Hello @tiltec . I have understood the requirement. You want the Create Pickup button to be replaced with Manage pickup dates button which will redirect you to Manage pickups page.
![scrshot](https://cloud.githubusercontent.com/assets/8210079/25558060/8b35b316-2d3c-11e7-8fdf-e4b650b49bbc.PNG)

tiltec:
Yes, very much so! Would you be happy to take this task?

MonsieurShaik:
Sure, thanks. :)

MonsieurShaik:
Hi @tiltec . Sorry for the delay. I have created a pull request. Here is the link.
https://github.com/yunity/foodsaving-frontend/pull/512

PR #512, opened 2017-06-05
Issue #474: Replaced Create Pickup button with Manage Pickup Dates button. Removed Delete Pickup Dates button from pickup-list-item,

MonsieurShaik:
I have also removed 'Check and Manage Pickup Dates' from the top bar/header of store details as we now have a more prominent 'Manage Pickup Dates' button.
I have added a new phrase in locale files in English. I have not updated the phrases in other languages as I am not familiar with them. Please check the attached screenshot.
Thanks.
scrshot 1


tiltec:
Cool stuff! That will allow people to find the manage-pickups page much easier, especially considering "headline blindness" (users tend to skip over the top part while looking for buttons).
Merged! :D

PR #538, opened 2017-07-01
use package-lock.json to pin versions

tiltec:
With the package-lock.json file we can now use wildcard versions in package.json and upgrade with `npm upgrade`. Greenkeeper isn't working so well anymore because it doesn't generate package-lock.json.

Pitfalls: npm install and npm upgrade actually change package.json and replace the wildcard with the installed version. `--no-save` prevents this, but also prevents package-lock.json to be updated.

The alternative is to keep versions in `package.json` and run `npm outdated` for detecting upgrades.


tiltec:
@nicksellen and me decided that #541 is a better solutin.

PR #557, opened 2017-07-05
Group join Improvements

D0nPiano:
I thought it'd be a good idea if groupList is a list like storeList, so I created a joinGroupList page which displays only the groups you are not part of.


client/app/components/pages.js
```diff
@@ -10,7 +10,8 @@ import topbar from "./_topbar/topbar";
 import landingPage from "./landingPage/landingPage";
 import notifications from "./notifications/notifications";
 import groupInfo from "./groupInfo/groupInfo";
-import groupList from "./groupList/groupList";
+import groupList from "./_groupList/groupList";
```
tiltec: groupList should not be imported here, but where it is used (e.g. joinGroupList)

client/app/components/joinGroupList/joinGroupList.js
```diff
@@ -0,0 +1,36 @@
...
+.config(($stateProvider, hookProvider) => {
+  "ngInject";
+  $stateProvider
+    .state("joinGroupList", {
+      parent: "main",
+      url: "/joinGroupList",
+      component: "joinGroupList",
+      resolve: {
+        groups: (GroupService) => {
+          return GroupService.list().then((allGroups) => {
```
tiltec: you don't need the then part, it does nothing

client/app/components/joinGroupList/joinGroupList.js
```diff
@@ -0,0 +1,36 @@
...
+])
+
+.component("joinGroupList", joinGroupListComponent)
+
+.config(($stateProvider, hookProvider) => {
+  "ngInject";
+  $stateProvider
+    .state("joinGroupList", {
+      parent: "main",
+      url: "/joinGroupList",
```
tiltec: I'd like to avoid camel-case urls, I would prefer something like /join-group or /group/join

tiltec:
I think a group list should also be available for anon visitors, complementary to the map on the landing page.

client/app/components/pages.js
D0nPiano: Yeah, makes sense.
I thought I might include it on a userPage soon, so that you can see what groups a user is part of. Do you think that's a good idea? Will change it otherwise

D0nPiano:
> I think a group list should also be available for anon visitors, complementary to the map on the landing page.

@tiltec  Didn't quite get it. You mean we should have a page with the list of all groups, for logged out users?

tiltec:
> You mean we should have a page with the list of all groups, for logged out users?

Yes. Groupinfo shows public infos for all kinds of users, and as a user I expect to be able to go "back" to a group list. That's what I implemented here.

tiltec:
> You mean we should have a page with the list of all groups, for logged out users?

Yes. Groupinfo shows public infos for all kinds of users, and as a user I expect to be able to go "back" to a group list. That's what I implemented here.

D0nPiano:
@tiltec So, should we create a seperate list of all groups, and link to it on the landing page then? Or just add a grouplist to the landing page? I just thought it was odd that it's a mixture of join group (with the url and the header "Which group would you like to join?") and a group list that just lists **all** groups on FS.world.

tiltec:
Maybe show different content, depending if the user is logged in or not? It could make sense to show the full list if the user is not logged in. For logged-in users it makes sense to filter out joined groups.

tiltec:
Oops, I didn't mean to merge this yet. But it's fine, we can just add the mentioned changes on top of master.
@D0nPiano I can do the changes that I proposed in my last comment if you are happy with it.

D0nPiano:
From my point of view a seperate, full list makes more sense somehow (then logged-in users could also access the complete list, and we would have the "join-group" URL for logged out users) - but I would be fine with it either way. So if you have something in mind, I'd definitely be happy with you changing it 🙂

tiltec:
I think that having two lists (which show similar content) could be confusing. I sometimes confuse group and store views because they look so similar.
Maybe we just mark the "already joined" groups differently, e.g. different color

D0nPiano:
Yeah, makes Sense. Different colors sound good, too 👍
And should it be able to be accessed by logged out users? And if so, should we put a button on the landing page then?

tiltec:
Yes logged out users should be able to access it. A simple button on the landing page could be good, maybe below the map.

Issue #567, opened 2017-07-23
Desktop: remove storeDetail map / zoom in on sidenav


D0nPiano:
The map on the left could easily replace the store map on desktop. Would try to implement something like this:

![screenshot from 2017-07-23 22-23-24](https://user-images.githubusercontent.com/16825880/28502573-9db07b38-6ff5-11e7-8a70-7fc57e998f02.png)

tiltec:
Quite a good idea!

PR #569, opened 2017-07-23
User map edit

D0nPiano:
## Changes
- created mapPicker component
- added mapPicker to storeEditCreate Form and UserDetail edit form
- added map to UserDetail

Will also be useful for displaying users on new Group map when building #567

### DIFF
--- a/client/app/components/_mapPicker/mapPicker.component.js
+++ b/client/app/components/_mapPicker/mapPicker.component.js
@@ -1,0 +1,14 @@
+import template from "./mapPicker.html";
+import controller from "./mapPicker.controller";
+import "./mapPicker.styl";
+
+let mapPickerComponent = {
+  bindings: {
+    placeholder: "@",
+    data: "<" // item with latitude, longitude and address
+  },
+  template,
+  controller
+};
+
+export default mapPickerComponent;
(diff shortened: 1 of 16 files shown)


client/app/components/_mapPicker/mapPicker.component.js
```diff
@@ -0,0 +1,14 @@
+import template from "./mapPicker.html";
+import controller from "./mapPicker.controller";
+import "./mapPicker.styl";
+
+let mapPickerComponent = {
+  bindings: {
+    placeholder: "@",
+    data: "<" // item with latitude, longitude and address
```
tiltec: You're using this as a two-way binding in the code, aren't you? E.g. when you do `$ctrl.data.something = "bla"` in mapPicker, the parent would see the changes.
So it should be "=" instead of "<". It would still work with "<" (one-way binding) but it violates general angular style guidelines.

Even better than using two-way bindings would be to use output bindings "&", because that is compatible with angular2+

@nicksellen do you have more input on this topic?

D0nPiano: Ah, yeah, forgot to change it. I tried to implement something like:
```
data: "<",
getData: "&"
```
where getData is called from the parent scope and evaluated in the child scope - the "&" operator does exactly the opposite though (calling from child, evaluate in parent scope).

Therefore I just stuck to two way binding like _markdownInput is also using. Just forgot to change "<" to "=" .

PR #572, opened 2017-07-24
GroupMap for desktop

D0nPiano:
## Changes
- divide sidenav in map and stores #567 
   - zoom into stores on storeview
   - remove storemap
- add invitations to group menu
- made font-size of toolbars a little smaller on mobile
- added users to groupMap, only really makes sense after merging #569
- create `map` var in CurrentGroup
- added awesomeMarkers

## ToDos
- [x] Replace user and store icons with nicer ones
- [x] Think of a better layout for groupMap 
- [x] figure out if there should be an overview button
- [x] add tooltips


client/app/components/group/_groupMap/groupMap.controller.js
```diff
@@ -0,0 +1,127 @@
...
+      options: {
+        showUsers: false
+      }
+    });
+  }
+
+  $onInit() {
+    // deep watching
+    this.destroy = this.$scope.$watch(() => this.CurrentStores.list, () => {
+      this.$timeout(() => this.update(), 200);
```
D0nPiano: I'm really not too happy with having these timeouts - but sadly the icons didn't always show up otherwise. Not quite sure what to do...

client/app/components/group/groupDetail/_groupOptions/groupOptions.html
```diff
@@ -54,6 +54,12 @@
     </md-button>
     <md-menu-content>
       <md-menu-item>
+          <md-button ui-sref="groupInvites" aria-label="{{ 'GROUP.INVITE_TITLE' | translate }}">
```
tiltec: This has 4 spaces indentation, while we usually have 2 spaces. Can you check your editor setting?

client/app/components/group/_groupMap/groupMap.controller.js
tiltec: It could have to do with the watch settings in ui-leaflet. Maybe play around with the `watchOptions` setting? Although `markers` always gets replaced with a new object, so even the most shallow watching should work.
https://github.com/angular-ui/ui-leaflet/blob/master/CHANGELOG.md#breaking-changes

client/app/components/group/groupDetail/_groupOptions/groupOptions.html
D0nPiano: Hm, odd. Don't know why it didn't fix it automatically....

client/app/services/group/currentGroup.service.js
```diff
@@ -14,7 +14,18 @@ export default class CurrentGroup {
     Object.assign(this, {
       value: {},
       User,
-      SessionUser
+      SessionUser,
+      map: {
```
tiltec: How is this related to the current group? These seem to be proxy methods that call down to the groupMap component. Seems unexpected to me.

D0nPiano: Yes, they are. I thought it's the best place to store them, since all of the map is group related and I didn't want to create another "currentGroupMap" service. Open for ideas though.

tiltec: The interface of the map proxy methods is quite different from the `Current*` services, which essentially are just data caches.
Somehow this looks to me like an "observer" scheme. The "groupMap" component registers an observer by setting the functions. So there are typical corner cases, e.g. what happens when the `groupMap` component gets destroyed? Will the proxy methods get set to doing nothing again?

As far as I understood, developers in Angular1 should avoid custom observer patterns and prefer watchable data bindings. It should be easier to understand when everything follows a similar pattern.

In this case, we could try to add a `mapOptions` object (e.g. to the currentGroup service)
It can get watched in `groupMap` like this:

```
this.$scope.$watch(() => this.CurrentGroup.mapOptions, () => {
    // run update methods
});
```

From other places in the code, you could set `currentGroup.mapOptions.overview = true`.

Does this make sense?

tiltec: Although even watches should be deregistered:

```
$onInit() {
this.destroy = this.$scope.$watch(...);
}

$onDestroy() {
this.destroy();
}
```

I skipped doing this in at least one case, shame on me ;) You can see that watches are some kind of observer pattern, thy both come with registering and deregistering.

D0nPiano: So you basically mean not storing methods, but only storing data? Basically for the case that the parent scope is destroyed and the methods wouldn't do what they should anymore?
I don't really think it's an observer pattern. As far as I understood, this would mean something like having an `observers` array in our `groupMap` or `currentGroup`, that all other controllers could be part of / subscribe to, and our groupMap would alert all methods in this array on changes.
But since we don't have subscriptions to messages here, but only an interface to a function that's called in it's own scope, I'd also be fine with having a couple of watch patterns less, and keeping it like this. ...will read myself a little more into it though to really understand the problem. Maybe I can find something on that topic.

tiltec: Based on the design principle that controller actions should only be triggered by angular's digest cycle, it makes sense to use watches.

Re observer: Obviously this is a special case with only one observer allowed. So in this regard, the watches would be better since they allow any number :)

I agree that if it would only be about the deregistering, it wouldn't be a huge deal. It's more about having similar interfaces (if possible) to reduce complexity. I would find a data store that is being watched easier to understand than the proxy methods.

D0nPiano: Yeah, I think I get what you are saying with "single observer". And having data (vs. methods) also makes it easier if somehow storeDetail loads faster than groupMap. So I just changed it to use only data in the CurrentGroup now, and also removed the timeouts (not quite sure why it didn't even display the markers before, should see if we report the problem again sometime...)

client/app/app.styl
```diff
@@ -22,12 +22,14 @@ md-toolbar
   height 42px
   max-height 42px
   min-height 42px
+  .md-toolbar-tools .md-button.md-icon-button
+        margin 0
```
tiltec: can you check the indentation here as well? seems to be 4 spaces

client/app/components/group/_groupMap/groupMap.controller.js
```diff
@@ -0,0 +1,131 @@
...
+        lat: this.CurrentGroup.value.latitude,
+        lng: this.CurrentGroup.value.longitude,
+        zoom: 12
+      };
+    }
+  }
+
+  $onDestroy() {
+    angular.forEach(this.watchers, (watcher) => {
+      watcher();
```
tiltec: maybe rename to `deregister()`?

Issue #159, opened 2016-12-01
Users can enter how much food they saved (feedback)


tiltec:
After doing the pickup, users can give a feedback:

- kilograms
- text field for comments

Should be separate for every user that participated in the pickup. Could be counted and used as statistics.

D0nPiano:
I like the idea. Maybe there could be something reminder-ish / bells / popup, that appears the next time you are online after a pickup is done?

djahnie:
Sounds nice!
Even though people probably tend to exaggerate a bit, when asked to enter the weight themselves each time. ;)
Some stats are definitely good for motivation! (Even Tais said it worked on her on foodsharing.de...^^)

djahnie:
The data structure in [yunity/foodsaving-backend#253](https://github.com/yunity/foodsaving-backend/issues/253) lets every user who took part in a pick-up enter a comment and weight. 
So when entering the weight, should the users just estimate what _they themselves_ carried away or the weight of the complete pick-up? I guess it's the latter, so we should tell them somewhere to avoid confusion.
Another thing: Should we noticably limit the amount of characters in the comment fields, to force people to give concise feedback (like e.g. 100 characters max) or allow for verbose narratives?

brnsolikyl:
I've been thinking about this issue here in our group and what I wanted to do regarding weight statistic is the same thing as Foodsharing.de did. Kristjan said the amount of food saved is based on an average that you take from weighing a limited number of pickups, say 10, and then you multiply this average by the total number of pickups. It seems much more reasonable than second-guessing the weight of each and every pickup.

It's a tough question, ideally we'd like to know more exactly the amount of food saved in each pickup and expect that foodsavers will be able to give this input. I'm a bit skeptical about it, but we could test it and see how people react to it. In my opinion I think the statistics https://github.com/yunity/foodsaving-frontend/issues/355 should be prioritized over this one. Then we'd be able to go back in history to see the total number of pickups and present the amount of food saved based on an average.

tiltec:
The weight estimates on foodsharing are far off the truth. I did it myself when I was team leader in Munich, setting the weight to the upper limit that was expect. I did save tons of food, but the website showed easily double the amount.

As a base for the total amount of saved food per store, I would do the following:
- collect feedback from people who actually picked up
- calculate the average
- multiply the average with the total amount of pickups

[...]

PR #575, opened 2017-07-28
Pickup-Feedback

D0nPiano:
Has no functionality yet, so services etc. all still have to be created... will solve (2) from #159


djahnie:
I love how you added the little guy with the trailer! 😃 
But I'd change the text above the weight scale to 'Amount of food the store gave to you (in kg)', so that it's clear that this number should be an estimate of _the whole_ pick-up, before sorting out and dividing between foodsavers.

tiltec:
I think it would be better that every user just reports what they saved (took home) themselves. This could need more discussion, I'd be happy to continue in #159

tiltec:
Motto: merge now, adjust later ;)

PR #579, opened 2017-07-30
add md-floating-label to mapPicker

djahnie:


tiltec:
This makes the address autocomplete in store edit/create look like an input field, same as in group edit/create.
I think that gives better UX. @D0nPiano what do you think?

PR #581, opened 2017-07-31
Added illustration for feedback amount

D0nPiano:
Felt like drawing something tonight... so we now have an illustration for the amount of food the user picked up:


djahnie:
Okay, obviously I was too late. Ignore my question in #159...^^

tiltec:
Some test cases for the setFeedbackImages method could be nice :)

client/app/components/group/pickupFeedback/pickupFeedback.spec.js
```diff
@@ -1,4 +1,6 @@
 import PickupFeedbackModule from "./pickupFeedback";
+import bagImg from "./bag.png";
```
tiltec: You could keep references to the images in the controller's constructor, then you don't need those. Would be _slightly_ cleaner to test.

```js
// constructor
   images: {
       bagImg, flourImg, // ...
   }

// test
expect($ctrl.isHigherImg($ctrl.flourImg)).to.be.false;
```

PR #582, opened 2017-08-05
WIP: Lots of changes and fixes

tiltec:
Sorry that I didn't split it further...

Two changes worth a screenshot: 

- hide login button when on login page & sidenav button moved from right to left
image

- groupMap: when a store is selected, reduce opacity of the others
image

Have a look at the individual commits for more detail.

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -1,6 +1,6 @@
 machine:
   node:
-    version: 6.1.0
+    version: 4.5.0
 dependencies:
   post:
     - npm run lint
--- a/circle.yml
+++ b/circle.yml
@@ -1,6 +1,6 @@
 machine:
   node:
-    version: 4.5.0
+    version: 8.2.0
 dependencies:
   post:
     - npm run lint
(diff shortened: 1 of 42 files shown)


tiltec:
I can't reproduce the test failures on my machine...

D0nPiano:
Like it a lot! Especially the simplifications on listeners and the groupMap & UI stuff :)

Only thing I noticed:
When I'm on storeDetail, and switch back to groupDetail, the opaque markers are not changing back to solid colors. Also, the user markers are not changing opacity at all (is that intended? Think it makes sense to make all other markers opaque when store is selected)

tiltec:
Agree, I will also reduce opacity of the user markers.

D0nPiano:
groupMap works perfectly now!

login-redirects: wouldn't it make sense to redirect to some kind of error page? If an error occurs, and it redirects to the login-page, I'd assume i just don't have the right to view the specific group/page etc.
If you want to I could create some basic error page - just not sure if it should be 404 (not found) or 500 (internal server error)?

tiltec:
I would be happy about a more complete solution, but I think just redirecting to the login page (instead of being stuck on a blank page) is already quite good. Usually those errors happen if the user got logged out, so showing a 404/500 page is not accurate either.

Internal server error and connection interrupted already show md-toasts, probably at the same time as the redirect to login happens... ah this seems complex, maybe we create an issue?

tiltec:
@D0nPiano can you reproduce those ci failures on your machine? https://circleci.com/gh/yunity/foodsaving-frontend/1581

tiltec:
Could reproduce the failures in docker, fixed them, clicking "merge" :)

PR #584, opened 2017-08-07
Add conversation service

nicksellen:


nicksellen:
Ok, I added tests! Not the most dazzlingly complete tests in the world, but maybe good enough for now?

If I removed the demo UI, would you be happy to merge to master @tiltec? - it would just have an extra service that would not be used anywhere.

I would open a new PR with the demo UI that could be the start of the UI work....

client/app/services/conversation/socket.service.js
```diff
@@ -0,0 +1,53 @@
...
+        this.subscribers.splice(idx, 1);
+      }
+    };
+  }
+
+  /* Connect if not already connected */
+  connect() {
+    if (this.socket) return;
+    this.socket = new ReconnectingWebSocket(
+      "ws://" + window.location.host + "/api/ws",
```
tiltec: Will this also work over HTTPS? I heard something of `wss://`

tiltec:
Looks solid to me, I would merge the service!

You could even leave the `conversation(groupId)` method in the groupService code (but add a little test).

nicksellen:
Oh yeah need to check for htts then use wss.

I would remove it from the group detail entirely as i think depending on ui
it might not end up there.

nicksellen:
> You could even leave the conversation(groupId) method in the groupService code (but add a little test).

Ah, I didn't read it properly first time. I understand now, and did that.

nicksellen:
Assuming the tests pass, then this is good to go!

PR #588, opened 2017-08-08
Improve state redirects

### LINKED ISSUE #484: Fix bugs and logic errors in group URL routing
Opening a page https://foodsaving.world/#!/group/16/store/34 if store `34` does not exist does not show anything. The console reports a 404 error.

It should redirect to the group page.

tiltec:
Remove `hookProvider`, replace with ui-router `resolve` configs. They have the advantage to also run when child states are loaded, in contrast to `hookProvider` and `redirectTo`.

- add resolvable snippets to check auth status and do redirect: `loggedInOrRedirectToLogin`, `loggedOutOrRedirectToHome`
- group has a special resolvable `beMemberOrRedirect`, which replaces the complicated `redirectTo` config and does more
- remove `SessionUser.loaded` promise, it was not well-suited to handle login properly
- enabling caching for `Authentication.update()` so we can call it as often as we want without a server roundtrip. The cache gets emptied on login and logout.

Closes #484


tiltec:
Since there is an interdependency between this PR and #591, is there resistance against merging this?

D0nPiano:
Didn't have a possibility to try out the changes, but just read through the code and it looks good to me. It really makes it clearer. And good that you updated the template, too.

And I think the error status from the api that I used in #591 is enough authentication-update anyways, so even better that you cached it here :)

Issue #589, opened 2017-08-08
Create a 404 NOT FOUND page


tiltec:
@D0nPiano are you motivated to create a _not found_ page for those cases (from #484 )?

- group id does not exist
- store id does not exist (in the group)

We could redirect to them in the `resolve` configs from #588 .

tiltec:
Probably also

- user id not found

D0nPiano:
Sure, will probably have some time to draw / click something together tomorrow :)

PR #591, opened 2017-08-10
404 Page

D0nPiano:
Solves #589

- [x] create page
- [x] redirect when not found from group, store, user and group-info


client/app/components/group/group.js
```diff
@@ -45,18 +45,23 @@ let groupPageModule = angular.module("group", [
...
-            return "group.groupDetail.pickups";
-          } else {
-            $translate("GROUP.NONMEMBER_REDIRECT").then((message) => {
-              $mdToast.showSimple(message);
-            });
-            // re-uses same groupId state parameter
-            return "groupInfo";
-          }
-        });
+        let $state = trans.injector().get("$state");
```
tiltec: This code is being replaced in #588, which does redirects a bit different.

tiltec:
Sorry for all those conflicts!

D0nPiano:
No problem. I changed the beMemberOrRedirect module so that the exceptions are all thrown to the highest level. Now they are resolved there - wherefore I was able to remove the timeout. Other than that it's mostly the same (just improved some tests while merging)

client/app/components/group/group.beMemberOrRedirect.js
```diff
@@ -22,29 +22,42 @@ It also loads group data into the CurrentGroup service, which persists it for th
...
+      })
+      .catch((error) => {
+        if (error.error !== "not_a_member"){
+          return $q.reject({ error: "not_found" });
+        }
+        return $q.reject(error);
+      });
   })
-  .catch(() => {
+  .catch((error) => {
```
tiltec: You could use [destructuring](https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment) here to avoid writing `error.error`:

```
.catch(({ error }) => {
  if (error === "not_found"){
      $state.go("notFound");
```

client/app/components/groupInfo/groupInfo.js
```diff
@@ -27,8 +27,12 @@ let groupInfoModule = angular.module("groupInfo", [
       url: "/group-info/{groupId:int}",
       component: "groupInfo",
       resolve: {
-        groupData: (GroupService, $stateParams) => {
-          return GroupService.get($stateParams.groupId);
+        groupData: (GroupService, $stateParams, $state) => {
+          return GroupService.get($stateParams.groupId)
+            .catch((error) => {
+              if (error.status === 403) $state.go("login");
```
tiltec: This shouldn't be possible as the groupInfo is always accessible.

client/app/components/group/group.beMemberOrRedirect.js
```diff
@@ -22,29 +22,42 @@ It also loads group data into the CurrentGroup service, which persists it for th
...
+    } else if (error.error === "not_a_member"){
+      $state.go("groupInfo", $stateParams);
+      return;
+    }
     $translate("GLOBAL.NOT_LOGGED_IN").then((message) => {
       $mdToast.showSimple(message);
     });
-    $timeout(() => $state.go("groupInfo", $stateParams));
-    return $q.reject("reason");
+    $state.go("login");
```
D0nPiano: Yeah, I wasn't quite sure what the common use case would be - I thought it'd mostly be people who are part of the group (because otherwise they wouldn't have the URL) and just need to login to see it. Therefore I thought I'd redirect it directly to login. Happy either way though.

client/app/components/group/group.beMemberOrRedirect.js
D0nPiano: Cool, didn't know something like that existed

tiltec:
I'm happy to merge this :)

PR #594, opened 2017-08-11
 fix bug (view all groups on group info…

jherfst:
Fix: 
1. Updated .ignore file.
   added extention (.idea)
2. fix bug on group info page. 
    left top icon was not correct.

### DIFF
--- a/client/app/components/groupInfo/groupInfo.html
+++ b/client/app/components/groupInfo/groupInfo.html
@@ -2,7 +2,7 @@
   <md-toolbar class="md-hue-1">
     <div class="md-toolbar-tools">
       <h1 class="md-title" flex>
-        <md-button ui-sref="groupList" class="md-icon-button">
+        <md-button ui-sref="joinGroupList" class="md-icon-button">
           <md-icon md-font-icon="fa fa-group"></md-icon>
           <md-tooltip>
             <span translate="JOINGROUP.ALL_GROUPS"></span>
--- a/.gitignore
+++ b/.gitignore
@@ -10,6 +10,7 @@ coverage
 .env
 nbproject/
 .vscode
+.idea/
 
 # To do: commit again after npm has been fixed
 package-lock.json


client/app/components/groupInfo/groupInfo.html
```diff
@@ -2,7 +2,7 @@
   <md-toolbar class="md-hue-1">
     <div class="md-toolbar-tools">
       <h1 class="md-title" flex>
-        <md-button ui-sref="groupList" class="md-icon-button">
+        <md-button ui-sref="joinGroupList" class="md-icon-button">
```
tiltec: Good find! We should add more end-to-end tests to the foodsaving-e2e repository!

jherfst: Yes I was also thinking about that. Let me now how I can also contribute to the foodsaving-e2e repository.

tiltec: Sure, we can have a look together if you want. The end-to-end tests spin up a completely new environment (frontend, backend, fresh database) and we can run tests using [protractor](http://www.protractortest.org)

Currently there are only two very basic tests: https://github.com/yunity/foodsaving-e2e/blob/master/e2e.spec.js

They can be easily extended, e.g. making sure that clicking on the button you fixed directs the user to the group list.
Would you be interested to try it out?

jherfst: aaaah nice. Yes sure let me give it a try. I will create the end to end test for the fix ( making sure that clicking on the button you fixed directs the user to the group list).

tiltec: The test setup for e2e is still a bit of manual work, I'm afraid. I think you can avoid setting up the backend if you run the e2e-tests against dev.foodsaving.world for now.

tiltec: This seems to be the rough steps to get it running:

1. setup and run the frontend dev server
2. in a new terminal, switch to the `foodsaving-e2e` repo
3. run `npm install`
4. update webdriver
```
$(npm bin)/webdriver-manager update
$(npm bin)/webdriver-manager start
```
5. in yet another terminal, switch to foodsaving-e2e and run `npm test`

this should open chrome and report something like this in the terminal:

```

> foodsaving-e2e@1.0.0 test /home/tic/Development/foodsaving-e2e
> protractor

(node:31929) [DEP0022] DeprecationWarning: os.tmpDir() is deprecated. Use os.tmpdir() instead.
[16:30:26] I/launcher - Running 1 instances of WebDriver
[16:30:26] I/hosted - Using the selenium server at http://localhost:4444/wd/hub

  ✓ e2e suite landing page should have a title: 1346ms
  ✓ e2e suite landing page should have a login button: 1333ms

  2 passing (5s)

[16:30:33] I/launcher - 0 instance(s) of WebDriver still running
[16:30:33] I/launcher - chrome #01 passed
```

jherfst: Ok @tiltec  i will work on it and let you know.

Issue #583, opened 2017-08-06
Add conversations


nicksellen:
The backend conversations is getting close to being useful - https://github.com/yunity/foodsaving-backend/pull/333

I'd like to proceed by working out what will be useful for the frontend service to do, so I can create the correct API endpoints and websocket contract.

The only thing that will have a conversation right now is the groups.

Here is my idea for how it would be used (inside a controller for example):

```js
// message storage in your controller
vm.messages = [];

// the idea here is the service handles all the logic of:
// 1. fetching initial messages (via API)
// 2. receiving new messages (via websocket)
// 3. putting them in the array you pass to it
// it then returns an object you can do further things with
let conversation = conversationService.subscribe(24, vm.messages);

// write a new message into the conversation...
conversation.sendMessage({
  content: 'Yay, a nice message from me!'
});

// go back in time to get older messages, then add them into the array
conversation.fetchPrevious(20);

// it's then important to unsubscribe or they will start piling up
// there is an angular way to call methods when the controller is removed iirc
conversation.unsubscribe();
```

The more annoying bit right now is how to get the conversation id, as I [implemented in the backend](https://github.com/yunity/foodsaving-backend/pull/333) now as a GenericForeignKey on the conversation, so the group doesn't know the conversation id without looking it up.

There is a separate endpoint `/api/groups/{id}/conversation/` for retrieving the conversation info, and no endpoints for listing conversations.

For now I would propose the frontend websocket would receive messages for all the conversations you are in, but would only do anything if something had subscribed to a conversation using the frontend service.

The alternative would be to add/remove conversation subscriptions on the server, but it is more complicated and seems of no benefit right now.

Questions:
1. can this actually work on a technical level? (I think it should...)
2. do you like the API style? (slightly funky managing your array for you...)
3. any input on my assumptions for the frontend UI?

tiltec:
Frontend assumptions seem valid to me. The conversations service would need to call `$apply` when the array changes to trigger a digest cycle.

UI: If we have the group conversation directly in the group page, would it be better to have the most recent message first (similar to "History")? Otherwise I would create a separated chat UI with a more common chat interface (oldest to newest, input at bottom).

> There is a separate endpoint `/api/groups/{id}/conversation/` for retrieving the conversation info, and no endpoints for listing conversations.

Seems fine, although it could be added as a SerializerMethodField. Will look into the backend after this.

nicksellen:
[...]

PR #595, opened 2017-08-11
WIP: Add chat UI

nicksellen:
I added a chat tab on the group with a basic chat UI. Mostly just to show how it works.

Still need to continue discussion about how it should work/look in #583.


tiltec:
I would fix all lint errors before this goes into master, even if this is just a dev UI. Otherwise the status reports become meaningless.

tiltec:
#597 extends on this, will close.

PR #597, opened 2017-08-13
Chat UI

D0nPiano:
A first idea of a chat UI - will probably be needed like this sometime anyways (for private messages) - but yet to be discussed if group-conversations should appear like this too / at first. Was motivated to do it anyways though since I just love the new chat :grin: 

chat

### UPDATE
removed chatFloater for now. Will be another PR.

### TODOs
- [ ] test
- [x] include profile pictures


tiltec:
Lars and Nick, you're both heroes!

random thoughts from me:

- right now, the list of conversations is the same as the group list. don't know when this will change. pretty soon maybe :)
- the chat box on the bottom right looks perfect to me. but stacking them side-by-side (like facebook or foodsharing.de) is not so nice. maybe just allow one open conversation.
- the tab in the group would be more suited for a wall-type (recent first, more formatting capabilities), which seems also what people want (e.g. see Bruno's message in Slack)
- you mention ctrl+enter, probably for line-break support. I'd like to support basic formatting, maybe similar to what [Slack supports](https://get.slack.help/hc/en-us/articles/202288908-Format-your-messages)? It's _very_ stripped markdown.

Does it make sense to have wall and a group chat, like foodsharing.de has? They seem to be for different purposes: wall for announcements, chat for ephemeral stuff.

D0nPiano:
Changed GroupDetail.Chat so that it looks like a wall now. Formating is also supported. Current status:

chat

tiltec:
You mentioned in Slack

> - add the "created_at" dates to messages corretly when getting new messages through the socket
> - adding tests
> - maybe adding the possibility to delete a message (if that's possible backend wise)

I opened a backend issue for the first point. The second one is important to me, though not sure if I'll find time to work on in the next weeks. The last one could be discussed & implemented later.

D0nPiano:
### Current Status
- Got rid of the chatFloater and chat in topbar for now
- added option to add submit button to markdown-input
- redid some design.

screenshot from 2017-09-04 23-55-11

### Current ToDos
- [ ] rename group.chat to group.wall
- [ ] fix conversations service test
- [ ] add tests for markdown-input submit function, chat (/wall) and wallpost

PR #598, opened 2017-08-19
WIP: Add cordova app

nicksellen:
Added initial cordova app. I think I probably added too many files to git though. Would like to try and remove as many as possible, and have a setup install anything you need.


tiltec:
I'm not sure if/how I should review this PR. If we are going to build a progressive web app, the need for having a playstore app might be lower.

nicksellen:
No review needed right now, still got some stuff to work out.

The main purpose of having an installable app is to get push notifications. Doesn't need to go on the play store either, can just be a dirty .apk, like orux maps do (at least for starters).

Issue #465, opened 2017-04-16
Refactor $mdMedia calls into screen size service


tiltec:
We use `$mdMedia` of angular material to determine the screen size. Right now, we usually define a controller function that returns true or false, depending on the screen size.

https://github.com/yunity/foodsaving-frontend/blob/be8f67ef0498198aabc6202c0f68fb2466b154ab/client/app/components/group/store/pickupManage/_pickupManageAction/pickupManageAction.controller.js#L9
https://github.com/yunity/foodsaving-frontend/blob/4215fdd100c7e6cab97f379380df9f3a1da9488b/client/app/components/_topbar/topbar.controller.js#L4

Or it is used in the code directly.
https://github.com/yunity/foodsaving-frontend/blob/4215fdd100c7e6cab97f379380df9f3a1da9488b/client/app/components/group/groupDetail/groupDetail.html#L38

My suggestion is to create a service that provides some wrapper functions for one commonly used breakpoint (e.g. `gt-sm`), which can be accessed from the templates.

cgomezmendez:
You mean wrapper functions like

serviceInstance.IsSmallScreen() ?
or something like that?

tiltec:
Hi @cgomezmendez! Are you motivated to work on this?

PR #599, opened 2017-08-23
refactor: Create service for determining screen size

azzang:
The service provides a function that wraps $mdMedia (an AngularJS Material service) and returns a boolean indicating whether a screen is wider than 960px.

The service can be extended by adding similar functions to test for other breakpoints.

Resolves: #465

### DIFF
--- a/client/app/components/_topbar/topbar.controller.js
+++ b/client/app/components/_topbar/topbar.controller.js
@@ -1,16 +1,11 @@
 class TopbarController {
-  constructor(Authentication, SessionUser, $window, $mdMedia, $scope, $rootScope, $mdSidenav) {
+  constructor(Authentication, SessionUser, $window, ScreenSize, $scope, $rootScope, $mdSidenav) {
     "ngInject";
-    $scope.screenIsSmall = !$mdMedia("gt-sm");
-    $scope.$watch(() => {
-      return $mdMedia("gt-sm");
-    }, (big) => {
-      $scope.screenIsSmall = !big;
-    });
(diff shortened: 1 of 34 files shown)


tiltec:
Looks good to me! Would you be up for replacing calls to `$mdMedia` in our codebase with your brand new service?

azzang:
Great! I'm totally up for replacing the calls. One quick question:

Do you have any preference as far as how the service should be exposed (on
the controller, scope, rootScope, ...)? I'm leaning toward the controller,
with calls from templates that look like:
$ctrl.ScreenSize.isMediumOrLarger(). I think this would be the cleanest
approach, in terms of replacing the calls.

Let me know :)

Jazz

tiltec:
> $ctrl.ScreenSize.isMediumOrLarger()

That seems the best choice to me, too.

Though maybe it makes sense to shorten `isMediumOrLarger()` to something like `gtSm()`?

azzang:
Done.

I added an "is" to your function name (i.e. isGtSm instead of gtSm) since
it returns a boolean. Also, I realized that $mdMedia was being passed
"gt-xs" in a few places, so I added another function (isGtXs) to the
service.

Not sure if I got everything, so any feedback would be great!

tiltec:
Thanks @azzang, great work! (also for the whitespace fixes 👍 )
I will merge it to master now, the changes should be live soon on dev.foodsaving.world

azzang:
Thanks Tilmann. Happy to help!

Issue #427, opened 2017-03-23
Rename "history" to "activity"


tiltec:
We have an audit log called "history" (currently displayed for groups). @nicksellen had the idea to rename it to "activity", and I support this proposal.

Upside: more fitting name
Downside: would need some refactoring, including backend API

What do you think?

A quick solution could be to change only frontend strings, but that could cause confusion.

nicksellen:
I think change is good, could leave it waiting here until somebody wants to do the work...

varu038:
Hello @nicksellen, @tiltec ,

I am new to the process of contributing to open source community although I have been doing coding for quite some time. I would like to start contributing with this bug. I have read through the CONTRIBUTE.md. Is it okay if I start with this bug ..? Can you please explain further ..?

tiltec:
Hi @varu038, great to have you here! First of all, sorry for the late reply - I have patchy internet access due to travels.
I think this is a nice PR to start with, happy if you do so! 

Do you have a tool to help you with the refactoring?
We will need to deploy frontend & backend at the same time, so there should a PR in each repo.

An area where some attention is needed are the translation strings (frontend locale-en.json). When we change the IDs, we need to take care to update all translated files as well and push them to transifex. Otherwise transifex will lose the translations ;)

Oh another thing: did you join our Slack already? slackin.yunity.org
It's our main communication tool, find us in the channel #foodsaving-tool

varu038:
Hello @tiltec ,
What kind of tool are you talking about for refactoring ..?

I have joined the Slack.

varu038:
grep history ./client/app/locales/locale-en.json -in 
52:    "HISTORY": "History",

171:  "HISTORY": {

@tiltec 
Are talking about renaming these to activity ..?

tiltec:
For example a good editor that can help with the renaming. But it can also be done with extensive search/replace.

I use PyCharm for the backend Python code and Atom for the frontend.

tiltec:
Yes, that's the messages (to sync with transifex). I guess you have seen that it's only the tip of the iceberg, most of the changes are in the code. Goal is a complete rename to "activity", so that future contributors won't be confused.

varu038:
Hello @tiltec ,

I am trying to update the translation files but its not allowing me.
How do I do it ..?

tx push -s
Pushing translations for resource foodsaving-tool.frontend:
Pushing source file (client/app/locales/locale-en.json)
Exception: Only maintainers are allowed to update the source file.

[...]

PR #600, opened 2017-08-25
Change instances of history to activity

theresaslee:
Addresses the frontend part of #427.

TODO:
 - [ ] change history api to activity. Did not do this yet since that doesn't exist.


client/app/components/group/_activity/_activityDetail/activityDetail.spec.js
```diff
@@ -1,9 +1,9 @@
-import StoreHistoryModule from "./storeHistory";
+import ActivityDetailModule from "./activityDetail";
```
djahnie: Why did you decide to change `storeHistory` to `activityDetail` and not `storeActivity`? I would have found the second one more intuitive. Did I miss something?

theresaslee: 🙈 I have no idea why I did that. Sorry about that!

theresaslee: Actually, looking at this more closely, I think Github might have mistaken where the changes were coming from when I renamed the files.
I believe [this](https://github.com/yunity/foodsaving-frontend/blob/master/client/app/components/group/_history/_historyDetail/historyDetail.spec.js) is the actual file I converted.

nicksellen: How confusing! I did a diff of the file trees [1] and all looks good there, and all the file contents look good too. Silly computers :)

[1] https://gist.github.com/nicksellen/1bef6d29edeaf2cb6aabbcffbe86efa0

djahnie: Weird... good to know that stuff like this can happen... 😅

tiltec:
Looks good, though I didn't look in detail! 

We need to take care again with the translation files - they are managed on transifex. If we just update them in the repo, they will get overwritten on the next translation update.

The only suitable process to change translation keys seems to me:

1. pull up-to-date translation files from transifex
2. rename the key in all translation files
3. push the changes back to transifex
4. hope that no translator is active between (1) and (3), otherwise their changes got lost
5. make a commit and push to github

Also: the backend part seems harder to me and there is currently no one working on it. How should we proceed?

We can merge this PR without the API change (as you prepared it), but if no one is following up on the backend side, we will get inconsistent naming. That's why, I'd merge only when the backend is ready, too.

tiltec:
Closing this because we merged the vuejs/quasar-based rewrite to master, which has completely different structure.

PR #615, opened 2017-09-10
thorough rename to karrot

tiltec:


.gitignore
```diff
@@ -12,5 +12,5 @@ nbproject/
 .vscode
 .idea/
 
-# To do: commit again after npm has been fixed
 package-lock.json
+cordova
```
nicksellen: What has cordova got to do with the rename?

client/index.html
```diff
@@ -7,7 +7,7 @@
     <meta name="mobile-web-app-capable" content="yes">
     <meta name="apple-mobile-web-app-capable" content="yes">
     <meta name="apple-mobile-web-app-status-bar-style" content="black">
-    <meta name="description" content="Foodsaving Tool">
+    <meta name="description" content="karrot">
```
nicksellen: I wonder if the Proper name is Karrot, or karrot. I would go for a capitalized one. I always feel foodsharing is a bit awkward when it goes at the beginning of a sentence, but people still want to keep it lowercased...

tiltec: I'm fine with lower case, though I don't care very much. But `description` should be set by the deployment anyways, so it should be `Foodsaving Worldwide` or similar.

PR #629, opened 2017-10-23
added verification warning

alangecker:
- [X] basic implementation
- [ ] i18n
- [ ] tests

### DIFF
--- a/src/components/Settings/VerificationWarning.vue
+++ b/src/components/Settings/VerificationWarning.vue
@@ -1,0 +1,37 @@
+<template>
+  <q-alert
+    v-if="!user.mailVerified"
+    color="warning"
+    enter="bounceInLeft"
+    leave="bounceOutRight"
+    appear
+    dismissible
+  >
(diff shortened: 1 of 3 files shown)


tiltec:
Maybe you can use MainAlerts (using vuex `alerts/create` to show the warning on the main page?

alangecker:
@tiltec but, as far as I understand, there is no option for having buttons and state in the MainAlerts like it is used in the VerificationWarning right now?

tiltec:
Maybe MainAlert with a link to settings to resend the verify email? so it's in addition to the work you did here

alangecker:
@tiltec but when you are on the settings it shouldn't be shown twice, so the MainAlert would need to be hidden in that case.
I would just merge it like that and maybe adding global warning later

tiltec:
Ok, then merge ahead :) It was just me thinking out loud.
Do you need approval before merging?

alangecker:
with the current procedure i guess it doesn't really matter :D

PR #648, opened 2017-10-27
Add cordova app

nicksellen:
Doesn't fully work yet, but quite a good start.

A few of the aspects that differ from normal website:
- have to set base url for api requests in webpack build (there is no implicit api backend)
- referer header is not set and django csrf protection complains, so added token auth (see https://github.com/yunity/karrot-backend/pull/415)
- cannot access cookies normally (needed for csrf) (using a modified version of cookie emporer, not currently added to repo)
  - have to check how csrf/token auth interacts...


src/index.ejs
```diff
@@ -14,6 +14,9 @@
     <title>(karrot loading...)</title>
     <link rel="shortcut icon" href="statics/carrot_logo.ico">
     <link rel="icon" type="image/png" href="statics/carrot_logo.png" sizes="32x32">
+    <% if (CORDOVA) { %>
```
tiltec: Can this also get loaded via webpack import?

src/cordova/csrf.js
```diff
@@ -0,0 +1,69 @@
...
+    request.headers[XSRF_HEADER_NAME] = csrftoken
+  }
+  return request
+})
+
+axios.interceptors.response.use((response) => {
+  return updateCsrfToken().then(() => response)
+}, err => {
+  return updateCsrfToken().then(() => Promise.reject(err))
+})
```
tiltec: When using token auth, we shouldn't need CSRF protection anymore. I'm not sure if Django supports this split behavior out of the box (enforcing CSRF only for SessionAuthentication), but it would be nice.

nicksellen:
Things todo are:
- [ ] resolve token/csrf interaction
- [ ] work out which files inside `cordova/` should be inside the repo or not (currently have a custom plugin on my local filesystem)
- [ ] add some stuff to readme/docs (can be a bit fiddly)
- [ ] verify the implemented features basically work (at the moment still get a bunch of console errors to resolve)

src/index.ejs
nicksellen: It's a buildtime configuration option, not runtime, the built files do not contain this. It's defined using the `WebpackDefine` plugin.

src/cordova/csrf.js
nicksellen: Needs investigating

tiltec:
Token auth merged in backend, let's go!

src/cordova/csrf.js
nicksellen: From docs, it seems not. Will try it out in reality

> If you're using SessionAuthentication you'll need to include valid CSRF tokens for any POST, PUT, PATCH or DELETE operations.

http://www.django-rest-framework.org/topics/ajax-csrf-cors/

nicksellen:
Almost ready! Only I discoverted #766 :/

nicksellen:
Needs to wait for https://github.com/yunity/karrot-backend/pull/440 to be merged

tiltec:
This needs a websocket token auth client #766, right?

nicksellen:
Yes, but #766 is addressed in this PR too.

So, it kinda works! (on my actual phone), some issues I'd like to resolve though:
- [ ] websocket connections seem really weird, it connects ok, but only occasionally receives a message on it, I had to fix up a few things on the server to get them to work, but don't think any of that is related
- [ ] cannot join/leave pickups on the group pickup overview page (other pages with join/leave pickups on work fine)
- [ ] find/implement a solution for https://github.com/yunity/karrot-frontend/issues/777
- [ ] publish it somewhere if possible (maybe build in circleci?)

Other issues I don't want to address in this PR:
- offline logic
- reloading data (feels like you should be able to pull down to refresh, quasar has a [nice component](http://quasar-framework.org/components/pull-to-refresh.html) for that)
- maybe a more immediate loading screen (loading cordova apps is somewhat slow...)
- make notification icons nicer (gmail etc have nice coloured ones (on the pull down top menu), our one is just white)

nicksellen:
Need to merge/deploy https://github.com/yunity/karrot-backend/pull/444 first to address the websocket issue, we may run into the issue again if we have lots of active connections though.

docs/android-sdk.md
```diff
@@ -0,0 +1,49 @@
+# Android Setup
+
+Android dev stuff is a bit annoying, you need to get an android sdk from somewhere. The easiest approach is to install [Android Atudio](https://developer.android.com/studio/index.html).
```
tiltec: Typo: Atudio

Issue #642, opened 2017-10-27
Make e2e tests work with real or mock backend


nicksellen:

tiltec:
Had a discussion with @nicksellen yesterday. There are two difficulties regarding the e2e setup:

1. how to serve the frontend to the test client
2. how to handle API requests

The simplest solution for (1) seems to be to run a Express server with the built files (output of `yarn build`).

For (2), an approach with low initial overhead is to respond to API requests in the Express server. Later on, the API requests should get proxied to a real backend, running inside docker-compose or so.

PR #656, opened 2017-10-29
WIP: add mock backend to e2e test

tiltec:
Related to #642 

Implements a partial mock backend in Express

- [ ] mock backend consists of a lot of copy-pasted data
- [ ] it does not join the group after signing up in the test (but the mock server supports it)
- [ ] add e2e test to CI
- [ ] fix e2e command in yarn

Mock backend is complicated, so I'm thinking about removing it and using the real backend instead. Could also solve the test issue mentioned above.

### DIFF
--- a/build/script.serve.js
+++ b/build/script.serve.js
@@ -1,0 +1,808 @@
+const express = require('express')
+const bodyParser = require('body-parser')
+const { join } = require('path')
+
+/**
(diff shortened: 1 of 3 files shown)


tiltec:
Superseded by #718 
Adding a mock backend is not worth it for me

Issue #324, opened 2017-02-18
Accept agreement before joining the group


tiltec:
**UPDATE** lastest proposal in https://github.com/yunity/karrot-frontend/issues/324#issuecomment-405231132

Additional backend field and checkbox required.

tiltec:
Random scribblings about this:

it's allowed to have groups without agreement
group member sets/changes agreement text
new members have to agree before they can join the group

questions:
- if agreement changes, does everybody have to agree again?
- who can change the agreement?
- what happens to members who joined before the agreement was online?
- do we need an overview page "who agreed to the agreement"?
- can members also withdraw their agreement later?

idea: make legal agreement immutable

D0nPiano:
I like the idea and your random scribblings

1. I think it's be best if everyone had to agree again.  It's simple and the most "natural" way of doing it.
2. Mhm, since we don't have roles... anyone then? That's a very tough one though
3. related to 1 - they have to agree again
4. I don't think we need one. I think it would make sense if people who did not agree could not perform any action in that group though (not beeing able to join pickups etc.)
5. If you couldn't perform any action when you didn't agree anyways, you could also just leave the group to disagree

Making the agreement immutable solves 1, 2 and 3. But I think only 2 is really problematic. Therefor, I am for beeing able to change it, I'm just unsure whether everyone should be able to, or only the group-creator. I think best would probably be some other admin way (...reminds me of my thoughts of a [Trust System](https://yunity.atlassian.net/wiki/display/FSINT/Simple+Trust+System) I never really finished)

stefan-fstaiwan:
There should be basic guidelines people have to agree on before they get an account in fstool... basically telling them that we are dealing with food, which means that there might be the risk of harming people and by signing in people basically agree in taking the risk on their own. Also making clear that when they want to pick up food from stores they must agree on a legal contract.
For people who want to share food p2p or through public fridges it is difficult to control any legal statement, but for people who do the pick ups it will actually be necessary. Most food retailers demand it! 
As the legal situations on food vary from country to country I guess we have to go by groups... maybe every group should have an imprint with information on the legal status. Some retailers might ask for stricter requirements then others... nevertheless I think it might be dangerous to have groups without any legal statements.

1. yes
2. No roles yet? Longterm there could be someone like "Betriebsverantwortlicher"
3. I think at some point there should be some kind of a restart of the website, so that people have to agree
4. + 5. agree with  D0nPiano 

[...]

PR #660, opened 2017-10-29
Initial group agreement management UI

nicksellen:
Implements one part of group agreements - showing an alert when the user has not agreed and letting them agree :)

You can try it out at https://group-agreements--karrot.netlify.com/#/group/1/wall (only group 1 has an agreement set).

You cannot edit the agrerements in the UI yet. In progress...

(note: TERMS OF SERVICE comes from the user specified values in the agreement, it would be whatever the group agreement manager set it to...)

agreements


nicksellen:
... and with the dialog opened...

agreements-open

nicksellen:
Ok, agreement management added!
- create initial agreement
- minor edit to agreement (users don't need to re-agree)
- major change to agreement (users need to re-agree)
- remove agreement

Currently nothing is actually restricted on the frontend or the backend, so that is next.

Although this is useful/mergable as it is, so would propose to continue work in future PRs.

nicksellen:
Oh, todo:
- [ ] hide "Minor edit?" checkbox when creating initial agreement

nicksellen:
@stefan-fstaiwan / @Sliverriver /  @brnsolikyl would appreciate your feedback too! Would be nice to know if this will help progress your foodsaving movements :)

You can try it out at https://group-agreements--karrot.netlify.com/ using the same credentials as you might have for any of the dev sites (https://dev.foodsaving.world/ / https://alpha.foodsaving.world/ / https://app.foodsaving.world/), and create a new account if you don't have one, or use foo@foo.com / foofoo.

Although, what is not implemented yet is: _ 3.2 If users join a group, joining and accepting should be done in one step_ (from https://github.com/yunity/karrot-frontend/issues/324#issuecomment-312685256).

nicksellen:
... although I just realised that as there is no UI for role management yet, you can't manage the agreement anyway :) ... but if you want to try it out, let me know, and I can manually give you the correct role...

src/components/Group/GroupEdit.vue
```diff
@@ -183,6 +183,10 @@ export default {
         this.$emit('save', { ...objectDiff(this.group, this.groupEdit), id: this.group.id }, event)
       }
     },
+    timezoneFilter (terms, { field, list }) {
```
tiltec: seems unrelated to the PR, does it enable more relaxed filtering of timezones? I think it's the same behavior as in the angular code, so it's good.

src/store/modules/agreements.js
```diff
@@ -0,0 +1,58 @@
...
+    commit(types.REQUEST_ENTRY)
+    commit(types.RECEIVE_ENTRY, { agreement: await agreements.get(id) })
+  },
+  async agree ({ commit }, id) {
+    commit(types.REQUEST_AGREE)
+    commit(types.RECEIVE_ENTRY, { agreement: await agreements.agree(id) })
+  },
+  async create ({ commit }, agreement) {
+    commit(types.REQUEST_CREATE)
+    agreement = await agreements.create(agreement)
```
tiltec: should handle server validation errors, maybe track the to do tasks in an issue?
by the way, I would rename the `onlyHandleAPIError` helper to `handleServerValidationError`, what do you think about it?

src/components/Group/GroupEdit.vue
nicksellen: Yes, I tried typing "berlin" and it didn't find `Europe/Berlin`. And yes it's unrelated, just seemed too small for the extra PR bureaucracy.

src/components/Group/AgreementForm.vue
```diff
@@ -0,0 +1,158 @@
...
+          :autofocus="true"
+          @blur="$v.agreementEdit.title.$touch"
+          autocomplete="off"
+        />
+      </q-field>
+
+      <q-field
+        icon="fa-file-text-o"
+        :label="$t('AGREEMENT.CONTENT')"
+        :helper="$t('AGREEMENT.CONTENT_HELPER')">
```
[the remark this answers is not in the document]
nicksellen: I think we are undoing each others changes then :)

I notice quasar does the on-a-new-line style, personally I don't like it though. I wonder if there vuetemplatelint we can use...

src/components/Group/AgreementForm.vue
```diff
@@ -0,0 +1,158 @@
+<template>
+  <div class="edit" :class="{ changed: hasChanged }">
+    <form @submit="save">
+      <q-field
+        icon="fa-star"
+        :label="$t('AGREEMENT.TITLE')"
+        :helper="$t('AGREEMENT.TITLE_HELPER')">
```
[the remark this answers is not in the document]
nicksellen: No, I'm very naughty...

stefan-fstaiwan:
I would like to try it out, but I'm confused now, how to do it...

nicksellen:
@stefan-fstaiwan I wrote to you in slack, can chat you through it (I need to do a manual step for you...)

PR #664, opened 2017-10-29
Fix GroupMap if store or group have no location

### LINKED ISSUE #602: Map shows  Darmstadt when no markers
In group 20 there is no store location specified. The map thus doesn't zoom into any location, even though a group location is added.

It would be better if the map zoomed into the group location, if no store location data was provided.
No marker should be displayed at the group location.

tiltec:
- if on group page and no markers are available and group location is not set, show button "set group location"
nogrouplocation
- if only group location is available, show low zoom on group location
grouplocation
- if in store mode and store location is not available, show button "set store location"
nostorelocation

Closes #602

### DIFF
--- a/src/components/Map/StandardMap.vue
+++ b/src/components/Map/StandardMap.vue
@@ -1,5 +1,5 @@
 <template>
-  <v-map ref="map" :bounds="bounds" :maxZoom="15">
+  <v-map ref="map" :bounds="bounds" :center="center" :zoom="zoom">
     <v-tile-layer :url="url" :attribution="attribution"></v-tile-layer>
     <v-marker v-for="marker in markers" :key="marker.id" v-bind="marker" @l-dragend="$emit('markerMoved', $event.target._latlng, marker)" :opacity="opacityFor(marker)">
       <v-popup v-if="marker.popupcontent" :content="marker.popupcontent"></v-popup>
@@ -47,6 +47,7 @@ export default {
     showAttribution: {
       default: true,
     },
+    defaultCenter: {},
   },
   data () {
     return {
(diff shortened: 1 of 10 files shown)


src/pages/Map.vue
```diff
@@ -15,8 +15,9 @@ export default {
   components: { GroupMap },
   computed: {
     ...mapGetters({
-      stores: 'stores/byActiveGroup',
-      users: 'users/all',
+      stores: 'stores/withLocation',
```
nicksellen: Is this really meaning `byActiveGroupWithLocation`?

from reading just the name it looks like it could overfetch the stores, and show ones for other groups.

src/store/modules/stores.js
```diff
@@ -39,7 +39,7 @@ export const getters = {
   all: state => state.idList.map(i => state.entries[i]).sort(sortByName),
   byActiveGroup: (state, getters, rootState, rootGetters) => getters.all.filter(e => e.group === rootGetters['groups/activeGroupId']),
   get: state => (id) => state.entries[id],
-  withLocation: (state, getters) => getters.all.filter(e => e.longitude && e.latitude),
+  withLocation: (state, getters) => getters.byActiveGroup.filter(e => e.longitude && e.latitude),
```
nicksellen: Hmm, as comment above, I find this unclear that it filters by active group without saying so in the name.

I would be happy without the `withLocation` getters, and rely on the map to filter those out (which it has to do anyway...)

src/locales/locale-en.json
```diff
@@ -298,5 +298,7 @@
   "already taken": "Already taken",
   "Enter a valid timezone": "Enter a valid timezone",
   "write a message": "Write a message...",
-  "write the first message": "Write the first message..."
+  "write the first message": "Write the first message...",
```
nicksellen: I think we need to rationalize this locale approach, seems to be append-only lately...

src/components/Map/StandardMap.vue
```diff
@@ -86,8 +87,25 @@ export default {
...
-      else {
-        return L.latLngBounds([{ lat: '49.8990022441358', lng: '8.66415739059448' }])
+    },
+    center () {
+      if (!this.bounds) {
+        if (this.defaultCenter) {
+          return [this.defaultCenter.latitude, this.defaultCenter.longitude]
+        }
+        else {
+          return ['49.8990022441358', '8.66415739059448']
```
nicksellen: Should be most amusing place possible...

src/store/modules/stores.js
tiltec: Yep, will remove the withLocation getters and handle it in the groupMap component

src/locales/locale-en.json
tiltec: For sure, we also need to add a lot of untranslated stuff into locale-en...

src/components/Map/StandardMap.vue
tiltec: it's still Darmstadt 🍂

PR #681, opened 2017-10-30
store status

### LINKED ISSUE #628: add store status
- [green] in cooperation
- [yellow] in negotiation
- [red] don't want to cooperate
- [none/grey?] archived

## Steps
- [x] implement in the backend
- [x] add dropdown to form
- [x] colorize markers in the map
- [x] add colored points for each store to the store list

alangecker:
almost done. still missing:
- [ ] tests
- [ ] i18n
- [ ] backend: deleting pickups when status is set to `archived`

closes #628


nicksellen:
Love the new store status UI!

The co-operating stores are the same colour as users though now, so they are not so clearly separated. Maybe it's ok though.

mappointercolours

tiltec:
If users are active (all users right now, there is no non-active status), they can have the same color as cooperating stores?

src/store/modules/stores.js
```diff
@@ -3,6 +3,44 @@ import stores from '@/services/api/stores'
 import { indexById, onlyHandleAPIError } from '@/store/helpers'
 import router from '@/router'
 
+const statusList = {
```
nicksellen: This does not belong in state (it never changes...), it's a UI consideration.

**Solution**: if it's used in multiple places, just pop it in a `.js` file somewhere with `export default { ...` and load it where it is used

src/pages/Store/Layout.vue
```diff
@@ -16,24 +16,38 @@
       </q-tabs>
     <router-view></router-view>
   </div>
+  <div v-else>
+    <q-card>
+      <k-alert color="warning" icon="fa-trash" :actions="[{label: 'Undelete', handler: undelete}]">
+        Store is not available anymore because it got deleted.
```
nicksellen: Should be put through _i18n_

src/locales/locale-en.json
```diff
@@ -328,5 +329,12 @@
   "write a message": "Write a message...",
   "write the first message": "Write the first message...",
   "set store location": "Set location of this store",
-  "set group location": "Set location of this group"
+  "set group location": "Set location of this group",
+  "STORESTATUS":{
+    "ACTIVE":"in cooperation",
```
nicksellen: In general everything starts with an uppercase letter, is there a reason to do it differently here?

nicksellen:
Would be great to get this merged soon, looks so close!

src/store/modules/stores.js
tiltec: Yeah, probably it's better to have this in a file.. but where? Just besides a component could be fine?

I'd also be fine if we import this file in the store and make it available via getter. I did something similar with the locales, it imports `i18n` and makes the locales available via getter. Makes it possible to mix in more properties.

nicksellen: Ah yeah, I didn't really like the locales approach either :p (as it returns in the format that the autocomplete component wants them in which is UI-specific).

My general view would be that data that does not change does not belong in state. Having to connect to vuex for static data is extra unneeded complication.

tiltec: Would be happy about a PR to improve the i18n static data handling!

djahnie:
Yep it's very nice! But can you maybe find another color so that active users and active stores don't look the same after all..? 😋 What about some kind of orange for users?

tiltec:
What if users can also be inactive in future?

djahnie:
Make them red but using a different shade. Or is there a limit to the amount of colors in use?

tiltec:
My feeling is that people are able to figure out the meaning 3-5 colors without much explanation. If we add more, we should add a legend, but there's not so much space on the minimap.
Also it would be nice to check at some point if people with [color blindness](https://en.wikipedia.org/wiki/Color_blindness) are able to see a difference between the markers (and other highlight colors).

tiltec:
Ooops, there are some conflicts now.. seems a bit annoying to resolve, might take some more thinking..

nicksellen:
I'm gunna have a look into this and try and get it working

nicksellen:
Ok, I think should be ready, I did:
- fixed conflicts from master
- fixed broken tests
- renamed delete/undelete to archive/restore (as that is what it does)
- made archive/restore only patch the status field
- added everything into i18n I could find
- moved store ui options to `src/services/storeStatus.js`
- capitalized and slightly reworded status text

nicksellen:
I did not:
- change the colours

tiltec:
screenshot_20171112-195324

Mobile store list is missing the status icons

Otherwise looks great, would be happy to merge soon.

PR #689, opened 2017-10-31
Small design improvements

D0nPiano:
Smaller design improvements, see commits...


src/components/Layout/KFooter.vue
```diff
@@ -1,12 +1,14 @@
 <template>
   <div class="footer bg-neutral font-primary">
-    <div class="footer-max-width row generic-padding justify-around">
-      <div class="row generic-padding justify-around">
+    <div class="footer-max-width row generic-padding justify-between">
+      <div/>
```
nicksellen: Do we need these self-closing divs? Maybe there is a quasar layout/flexbox/column type thing that works?

src/components/General/KBreadcrumb.vue
```diff
@@ -49,6 +49,6 @@ export default {
     padding 0 16px
     font-weight 500
     text-transform uppercase
-  .lastElement
+body.desktop .lastElement
```
nicksellen: How does this work when it is scoped? the `body` element is not in this component?

src/components/Wall/JoinedPickups.vue
```diff
@@ -33,7 +33,7 @@ export default {
   },
   data () {
     return {
-      showPickups: true,
+      showPickups: false,
```
nicksellen: Me and @tiltec think showing the joined pickups by default is nice, maybe makes it more likely that people will do them :)

src/components/Layout/KFooter.vue
D0nPiano: Yeah, I somehow couldn't get it done otherwise. The "justify-around" option was doing the trick on desktop, but would also add margins on mobile (which was why it was indented so weirdly). I will have a look at the content options. Maybe those help...

...I actually thought of redoing the footer completely somewhen anyways though since I'm still not too happy with it.

src/components/General/KBreadcrumb.vue
D0nPiano: Works just the same, it's only that the element with the class "lastElement" has to be in the same template

src/components/Layout/KFooter.vue
nicksellen: Ok, lets leave it like that for now then if it will be redone in not so long :)

src/components/General/KBreadcrumb.vue
nicksellen: Ok, seems fine, just confused me a bit.

src/components/Wall/JoinedPickups.vue
nicksellen: Great :)

D0nPiano:
Did some more improvements on the footer. -> Ready for review / merge.

screenshot from 2017-11-03 11-12-20
screenshot from 2017-11-03 11-12-41

src/components/Layout/KTopbarUI.vue
```diff
@@ -3,7 +3,7 @@
     <q-btn slot="left" flat class="mobile-only" @click="$emit('toggleSidenav')">
       <i class="fa fa-bars"></i>
     </q-btn>
-    <router-link :to="{name: 'group'}" class="logo">
+    <router-link :to="{name: 'group'}" class="logo desktop-only">
```
tiltec: I tend to use `v-if="!$q.platform.is.mobile"` instead because it prevents rendering, while the `desktop-only` class only hides it via css. But as we don't have performance problems yet, it's fine either way.

D0nPiano: Yeah, makes sense (especially for a picture that doesn't have to be downloaded then...)

PR #690, opened 2017-10-31
Fix for HistoryList updating 

### LINKED ISSUE #678: History component is not reactive
As all state is loaded via a `mounted` lifecycle hook, when the props change it is not updated. A simple way to recreate the issue:
- open https://dev.foodsaving.world/#/user/22
- switch the _22_ in the url to _222_
- notice the history entries are still for user 22

Possible solutions:
- remove use of vuex-connect and swap it for a simple wrapper component that can use watchers to refetch history data on props change
- revert to previous approach (stateful component that does not take props)
- checkout vue-router/vuex-connect docs to look at router update hooks to trigger data loading
- come up with another approach

alangecker:
fixes #678
not beautiful, maybe a different approach would be better, but it works ;)

### DIFF
--- a/src/components/History/HistoryList.vue
+++ b/src/components/History/HistoryList.vue
@@ -1,23 +1,41 @@
+<template>
+  <HistoryListUI
+    :history="history"
+    :status="status"
+    :canLoadMore="canLoadMore"
+    :fetchMore="fetchMore"
+    />
+</template>
 <script>
-import { connect } from 'vuex-connect'
 import HistoryListUI from '@/components/History/HistoryListUI'
-
-export default connect({
-  gettersToProps: {
+import { mapActions, mapGetters } from 'vuex'
+export default {
+  props: {
+    user: { required: false },
+    store: { required: false },
+    group: { required: false },
+  },
+  computed: mapGetters({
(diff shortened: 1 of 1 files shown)


src/components/History/HistoryList.vue
```diff
@@ -1,23 +1,41 @@
...
-    mounted ({ dispatch }) {
-      if (this.user) dispatch('history/fetchForUser', this.user)
-      else if (this.group) dispatch('history/fetchForGroup', this.group)
-      else if (this.store) dispatch('history/fetchForStore', this.store)
-    },
-    destroyed: ({ dispatch }) => dispatch('history/clear'),
+  watch: {
+    user () { this.fetchForUser(this.user) },
+    group () { this.fetchForGroup(this.group) },
+    store () { this.fetchForGroup(this.store) },
```
nicksellen: You could remove a bit of the duplication by creating an `refresh` method that can be called from `mounted()` and each of the watchers. It would do the same as the current mounted function.

tiltec: Agree, but can also be done in a future refactor. Will merge now.

PR #692, opened 2017-11-02
add backend introduction to docs

tiltec:
from https://github.com/mddemarie/Foodsaving-Documentation/blob/master/foodsaving-backend-code-guide.md with my pull request applied

I just went through my open Pull Requests and remembered the backend introduction written by @mddemarie and @id-gue. My proposal would be to add it to the new karrot docs. What do you think?


tiltec:
will merge now, can be removed later again if it doesn't quite fit

Issue #691, opened 2017-11-01
Rationalize request/validation meta in vuex modules


nicksellen:
### Problem

We have lots of repeated action defintions like this:

```js
async fetchListByGroupId ({ commit }, groupId) {
    commit(types.REQUEST_LIST)
    try {
      commit(types.RECEIVE_LIST, { pickups: await pickups.listByGroupId(groupId), groupId })
    }
    catch (error) {
      commit(types.RECEIVE_LIST_ERROR, { error })
    }
  },
```

This is bad because:
- we might implement it slightly differently each time
- hard to improve across all files as lots to change
- makes it feel very heavy to add actions
- low signal/noise ratio (most of the code is boilerplate)

### Solution

There are two useful types of data here:
- meta data about the request, and validation
- longer lived _actual_ data

We want to separate these two concerns, so longer lived data would still be handled as it is, but we seperate all the boring request/validation meta data out into seperate nested vuex modules.

### Implementation

The parent store (e.g. `groups`) needs to have access to nested module getters to get the meta state, and can add it to the data object in an enrich method.

In a vuex module you would create a meta module, which does a few things:
- defines state and mutations for modifying meta about requests and validation (keyed by `metaId`)
- define a getter to get nice stuff by `metaId` (perhaps including a nice error mapper getter thing?)
- includes a function to define an action to be included in the main parent store thing (e.g. the `groups` module itself), which will call the appropriate mutations in the meta module

```js
const meta = createMetaModule()
export const modules = { meta } // I *think* that is how nested modules work?

export const getters = {
   enrich (state, getters) => pickup => {
    return { ...pickup, getters['meta/get'](pickup.id) }
  }
}

// .. blahblhablah

export const actions = {
  // "createRequestAction" returns a regular action definition
  leave: meta.createRequestAction({
    metaId: pickupId => pickupId,
    async request ({ dispatch }, pickupId) {
      await pickups.leave(pickupId)
      commit(types.RECEIVE_LEAVE, { pickupId, userId: rootGetters['auth/userId'] })
    },
  }),
}
```

`createRequestAction` takes a configuration object that wants a few things:
- `metaId` this is the key used to store the meta against, and what you have to use in the getter, can also be a fixed value too (e.g. in the case of meta about a `list` request)
- `request` a function that behaves exactly like a normal action (internally it is wrapped into something that handles error better)

The error handling needs to distinguish between request errors (400/500, etc... anything from axios), and programming errors in the bit of code _after_ the request.
[...]

PR #693, opened 2017-11-02
Rationalize vuex requests

tiltec:
First two iterations of #691

### DIFF
--- a/src/store/helpers.js
+++ b/src/store/helpers.js
@@ -29,7 +29,27 @@ export function onlyHandleAPIError (error, handleFn) {
   }
 }
 
-export function defineRequestModule ({ namespace = 'meta' } = {}) {
+/**
+ * Defines a vues module that can be used to handle async request metadata.
+ *
+ * Instantiation:
+
+ const meta = defineRequestModule()
+ export const modules = { meta }
+
+ * Usage:
+
+async save ({ commit, dispatch }, group) {
+  dispatch('meta/request', {
+    id: `save-${group.id}`,
+    async run () {
+      const updatedGroup = await groups.save(group)
+      commit(types.RECEIVE_GROUP, { group: updatedGroup }) // other commits when successful
+    },
+  })
+},
+ */
+export function defineRequestModule () {
   const types = {
     REQUEST: 'Request',
     RECEIVE_SUCCESS: 'Receive Success',
@@ -51,10 +71,30 @@ export function defineRequestModule ({ namespace = 'meta' } = {}) {
       const error = getters.get(id).error
       return error && error[field] && error[field][0]
     },
+    isWaiting: (state, getters) => id => getters.get(id).isWaiting,
+    success: (state, getters) => id => getters.get(id).success,
   }
 
   const actions = {
-    clear: ({ commit }) => commit('clear'),
+    async request ({ commit }, { id, run }) {
+      commit(types.REQUEST, { id })
+      try {
+        await run()
+        commit(types.RECEIVE_SUCCESS, { id })
+      }
+      catch (error) {
+        const { response: { status = -1, data } = {} } = error
+        if (status >= 400 && status < 500) {
+          commit(types.RECEIVE_ERROR, { id, error: data })
+        }
+        else {
+          throw error
+        }
+      }
+    },
+    clear ({ commit }) {
+      commit('clear')
+    },
   }
 
   const mutations = {
@@ -88,42 +128,11 @@ export function defineRequestModule ({ namespace = 'meta' } = {}) {
     },
   }
 
-  function namespaced (val) {
-    return namespace + '/' + val
-  }
-
-  function createRequestAction ({ metaId, request }) {
-    // we need to use the namespaced commits here, as the commit happens in parent
-    // I didn't find a good workaround, maybe it would be better to dispatch an action from parent to child?
-    return async function (store, arg) {
-      const { commit } = store
-      const id = metaId(arg)
-      commit(namespaced(types.REQUEST), { id })
-      try {
-        await request(store, arg)
-        commit(namespaced(types.RECEIVE_SUCCESS), { id })
-      }
-      catch (error) {
-        console.log('receive error with id', id, error)
-        const { response: { status = -1, data } = {} } = error
-        if (status >= 400 && status < 500) {
-          commit(namespaced(types.RECEIVE_ERROR), { id, error: data })
-        }
-        else {
-          throw error
-        }
-      }
-    }
-  }
-
   return {
-    [namespace]: {
-      namespaced: true,
-      state,
-      getters,
-      actions,
-      mutations,
-      createRequestAction,
-    },
+    namespaced: true,
+    state,
+    getters,
+    actions,
+    mutations,
   }
 }
--- a/src/store/helpers.js
+++ b/src/store/helpers.js
@@ -3,6 +3,8 @@
  * If a helper depends on the vuex store, put it in storeHelpers.js
  */
 
+import Vue from 'vue'
+
 /**
  * Returns an object that maps entries of iterables by their `id` field
  *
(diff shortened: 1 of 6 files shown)


src/store/helpers.js
```diff
@@ -26,3 +28,111 @@ export function onlyHandleAPIError (error, handleFn) {
...
+  const actions = {
+    async request ({ commit }, { id, run }) {
+      commit(types.REQUEST, { id })
+      try {
+        await run()
+        commit(types.RECEIVE_SUCCESS, { id })
+      }
+      catch (error) {
+        const { response: { status = -1, data } = {} } = error
+        if (status >= 400 && status < 500) {
```
nicksellen: This would still need to split the errors that the user can handle (validation errors) from those that they can't (permission errors).

Would be nice to extract that logic out to a helper, so we can test it well. A few standalone functions should suffice, something like:
- `isValidationError (error) => boolean` (need to show validation on page)
- `isRequestError (error) => boolean` (if not also a validation error, means it was handled by axios interceptor and a toast was displayed)

.. if neither of those, then it's some other error.

src/store/helpers.js
```diff
@@ -26,3 +28,111 @@ export function onlyHandleAPIError (error, handleFn) {
...
+    RECEIVE_ERROR: 'Receive Error',
+    CLEAR: 'Clear',
+  }
+
+  const state = {
+    entries: {},
+  }
+
+  const getters = {
+    get: state => id => state.entries[id] || ({
```
nicksellen: I am not a fan of the default value being an object, I think we should move away from the convention as I think it is counter intuitive as you cannot check the result for truthiness, but have to look inside for a particular property

src/store/helpers.js
```diff
@@ -26,3 +28,111 @@ export function onlyHandleAPIError (error, handleFn) {
...
+export function defineRequestModule () {
+  const types = {
+    REQUEST: 'Request',
+    RECEIVE_SUCCESS: 'Receive Success',
+    RECEIVE_ERROR: 'Receive Error',
+    CLEAR: 'Clear',
+  }
+
+  const state = {
+    entries: {},
```
nicksellen: Starting state should be the same as after the CLEAR mutation (an `initialState ()` function would solve that).

src/store/helpers.js
```diff
@@ -26,3 +28,111 @@ export function onlyHandleAPIError (error, handleFn) {
...
+  }
+
+  const getters = {
+    get: state => id => state.entries[id] || ({
+      isWaiting: false,
+      error: null,
+      success: false,
+    }),
+    error: (state, getters) => id => field => {
+      const error = getters.get(id).error
```
nicksellen: Nitpick: destructuring is nicer here :)

src/store/modules/groups.js
```diff
@@ -77,6 +76,11 @@ export const getters = {
     return state.idsList.map(getters.get)
   },
   status: state => ({ isWaiting: state.isWaiting, error: state.error }),
+  saveError: (state, getters) => {
+    // access meta error getter with metaID
+    // a bit cumbersome!
+    return getters['meta/error'](`save-${getters.activeGroupId}`)
```
nicksellen: Maybe we should revisit the idea of multiple meta modules if there are multiple data types, so each meta is just the object id, or a string.

In this case though is there any reason not to just make the meta id the group id?

src/store/helpers.js
```diff
@@ -26,3 +28,111 @@ export function onlyHandleAPIError (error, handleFn) {
...
+ * Defines a vues module that can be used to handle async request metadata.
+ *
+ * Instantiation:
+
+ const meta = defineRequestModule()
+ export const modules = { meta }
+
+ * Usage:
+
+async save ({ commit, dispatch }, group) {
```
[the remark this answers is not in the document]
nicksellen: Another approach is to have a standalone `createRequestAction` function, that _assumes_ the meta module name is meta (but could override it as a configuration value if needed).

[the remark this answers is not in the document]
tiltec: I needed to prefix the commit with `meta`, that's right. Have a look at my first commit in this PR for a working example.

(will respond to the other comments later)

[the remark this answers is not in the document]
tiltec: `createRequestAction` doesn't make much sense on its own, as it assumes the existence of the request/success/error mutations.

What do you think of my first approach? https://github.com/yunity/karrot-frontend/pull/693/commits/433749045b2bf1cc9d9eae37c4a8835978528718#diff-c7f78e508c58f98cb69756719dac6056R95

src/store/helpers.js
```diff
@@ -26,3 +28,111 @@ export function onlyHandleAPIError (error, handleFn) {
...
+  dispatch('meta/request', {
+    id: `save-${group.id}`,
+    async run () {
```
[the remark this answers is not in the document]
tiltec: I'm probably biased, but I like this style ;) It seems to tell me a story: to save the group, I need to start a request that runs and saves the group via service call.

src/store/helpers.js
tiltec: It needs to be handled on some level, I'd prefer not having to handle it inside the components (means they should be able to expect `isWaiting`, `error` and `success` properties to exist)

tiltec: It could be handled in the parent `groups` module though, via additional getter.

src/store/helpers.js
tiltec: Ah, good idea! Didn't spend much time thinking about this code, it just worked already ;)

src/store/helpers.js
tiltec: Extracting the checks seems like a good idea. Though we would need still need the additional logic to get the response data, so the code won't get shorter, only more explicit.

src/store/modules/groups.js
tiltec: To keep the existing behaviour, I would use IDs like `save`, `join`, `leave`, `create`... there seems currently no need to handle the status of multiple groups at the same time.
If we would need the latter, than having multiple meta modules makes sense (namespaced by action?)

tiltec:
some more code! I really like this approach already 🕺 🕺

nicksellen:
I would like to explore a few more options here, I would try some stuff out myself (on top of your work) if you are not motivated to change it further... again, I think it's worth getting right, rather than a quick merge

tiltec:
Yes, please make the changes that are important to you. It's quite hard for me to implement what you have in mind.
The solution with a separate `createRequestAction` module might be a solution without passing a callback into action. `createRequestAction` then gets a namespace as argument where it will commit the mutations.

```js
function createRequestAction ({ namespace, action }) {
  // commit REQUEST, RECEIVE_SUCCESS, RECEIVE_ERROR in given namespace
}
```

src/store/helpers.js
```diff
@@ -26,3 +28,140 @@ export function onlyHandleAPIError (error, handleFn) {
...
+        error,
+        success: false,
+      })
+    },
+    [types.CLEAR] (state, { id }) {
+      Vue.delete(state.entries, id)
+    },
+    [types.CLEAR_ALL] (state) {
+      Object.entries(initialState())
+        .forEach(([prop, value]) => Vue.set(state, prop, value))
```
nicksellen: I think `Object.assign(state, initialState())` should work fine here, as the props in the initial state, are by definition already defined... and object.assign will just assign them again.

See https://vuejs.org/v2/guide/reactivity.html#Change-Detection-Caveats (implicitly says it works fine)

nicksellen:
Rebased to master

nicksellen:
So, goals I would like to satisfy are:
- make the `run()` function have the same signature as the original request handler (so they are drop-in replacements)
- remove duplication of metaId functions
- have the meta module name only defined once
- minimal impact of the meta stuff when writing actions

Which might involve changes:
- multiple meta modules
- multiple modules defined in one function
- ... other stuff I play with on the way :)

Issue #680, opened 2017-10-30
Add linting for vue templates


nicksellen:
https://github.com/vuejs/eslint-plugin-vue looks nice, but seems quite confusing/broken. Me and @tiltec failed to get it to work.

They have a version 4 roadmap/release thing, so maybe if we wait until that is released it'll JustWork.

nicksellen:
Gunna try https://github.com/BenoitZugmeyer/eslint-plugin-html

nicksellen:
> This ESLint plugin allows linting and fixing inline scripts contained in HTML files.
Oh, I didn't read it very well, it's only look at the scripts, not the template... we also already had the plugin installed.

PR #695, opened 2017-11-02
Add mobile options menu

### LINKED ISSUE #645: Add group/store context menu on mobile
There is a menu on desktop:

desktop

But these things are not available on the mobile view right now, there were two ideas to add this:
- 3 dots in the header which opens a menu
- extra tabs (they will scroll), alongside Wall, Description, History, ...

mobile

tiltec:
Closes #645 

Added vertical dots at the end of group and store tab-bar.

### DIFF
--- a/src/components/Group/Navbar.vue
+++ b/src/components/Group/Navbar.vue
@@ -1,25 +1,22 @@
 <template>
   <div slot="navigation" class="shadow-14 inset-shadow row">
-    <q-tabs color="tertiary" hide="icon" class="col">
+    <q-tabs color="tertiary" hide="icon" class="col" ref="groupnav">
       <q-route-tab
         :to="{name: 'group'}"
         exact
-        slot="title"
-        name="wall">
+        slot="title">
         {{ $t('GROUP.WALL') }}
       </q-route-tab>
       <q-route-tab
         :to="{name: 'groupDescription'}"
         exact
-        slot="title"
-        name="description">
+        slot="title">
         {{ $t('GROUP.DESCRIPTION') }}
       </q-route-tab>
       <q-route-tab
         :to="{name: 'groupHistory'}"
         exact
-        slot="title"
-        name="history">
+        slot="title">
         {{ $t('GROUP.HISTORY') }}
       </q-route-tab>
     </q-tabs>
@@ -36,6 +33,13 @@ import GroupOptions from '@/components/Sidenav/GroupOptions'
 
 export default {
   components: { QTabs, QRouteTab, QBtn, QIcon, GroupOptions },
+  watch: {
+    $route (route) {
+      if (route.name !== 'group' && route.name !== 'groupDescription' && route.name !== 'groupHistory') {
+        this.$refs.groupnav.selectTab(null)
+      }
+    },
+  },
 }
 </script>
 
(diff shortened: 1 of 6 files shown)


src/pages/MainLayout.vue
```diff
@@ -15,17 +15,17 @@
         </template>
         <MainAlerts />
         <div class="mainContent row justify-between no-wrap">
-          <div class="whiteSpace gt-sm desktop-only"/>
-            <router-view class="desktop-only sidenav-desktop" name="sidenav"></router-view>
+          <div v-if="!$q.platform.is.mobile" class="whiteSpace gt-sm"/>
+            <router-view class="sidenav-desktop" name="sidenav"></router-view>
```
D0nPiano: Just checked out the branch - the router view (whyever) also needs the `v-if="!$q.platform.is.mobile"`, otherwise it will still display the sidenav on mobile for me.

src/components/Group/Navbar.vue
```diff
@@ -1,38 +1,41 @@
...
+      <q-route-tab
+        :to="{name: 'groupHistory'}"
+        exact
+        slot="title"
+        name="history">
+        {{ $t('GROUP.HISTORY') }}
+      </q-route-tab>
     </q-tabs>
+    <q-btn flat class="bg-tertiary text-white">
+      <q-icon name="fa-ellipsis-v" />
```
D0nPiano: I was always using the `<i class="fa fa-ellipsis-v"/>` style for icons. Is there an advantage of using the quasar ones?

src/components/Group/Navbar.vue
```diff
@@ -1,38 +1,41 @@
...
 
 <script>
-import { QTabs, QRouteTab } from 'quasar'
+import { QTabs, QRouteTab, QBtn, QIcon } from 'quasar'
+import GroupOptions from '@/components/Sidenav/GroupOptions'
 
 export default {
-  components: { QTabs, QRouteTab },
-  computed: {
-    shouldBeDisplayed () {
```
D0nPiano: Now that 'shouldBeDisplayed' is removed, the tabs are still around on the member page - and the last active one (history, description...) is still highlighted. I think we should either remove the tabs, or remove the highlighting

D0nPiano: Or probably we should just add the "members" option to the tabs. Don't think it's available somewhere else....

src/pages/MainLayout.vue
tiltec: ah interesting! Seems to be different than a regular component or slot then..

src/components/Group/Navbar.vue
tiltec: I guess both is fine, `q-icon` seems to be a bit more explicit than the `i` tag.
It provides the same interface for using fontawesome and materialicons, if we need the latter at all: http://quasar-framework.org/components/icons.html#Basic-Usage

src/components/Group/Navbar.vue
tiltec: I think it would be best if the last active one wouldn't be highlighted. Seems like a bug in quasar to me...
The check happens one component above, it checks if the route name starts with `group`. This is just a very basic workaround to have group navigation available on most group pages that are not in the top nav (members, invitations...)

src/pages/MainLayout.vue
tiltec: Ok now I get it.. I totally missed that the div above was self-closing, because router-view has child indentation. I really hope we get #680 soon...

tiltec:
Should be ready to merge!

src/pages/MainLayout.vue
nicksellen: Divs for spacing also makes me feel quite uneasy

Issue #543, opened 2017-07-02
Add package-lock.json file again after npm bugs have been ironed out


tiltec:
We didn't get consistent and working builds and there has been a lot of bugfixing in npm in the last days. After npm works reliable, we should commit package-lock.json again and merge #541

tiltec:
Had a bad experience with npm 5.3.0 on my brothers laptop, we tried to `git pull && npm install` an old clone of the foodsaving-frontend repo. npm failed reliably 👎  by not installing deps-of-deps or even removing them. Not even `rm -r node_modules` helped..

`yarn` did the job perfectly and quickly, so I would make it as default in the README and the circle.yml. Greenkeeper-lockfile supports yarn.lock, too.

PR #697, opened 2017-11-05
Display wall/history dates as distances in words

ReneCode:
As Nick suggested I used the data-fns package to get the "distanceInWordsToNow".
For the spec test I was not sure how to mock the 'i18n/locale' getter - so I used the default behavior.


src/components/General/DateAsWords.vue
```diff
@@ -0,0 +1,39 @@
+<template>
+  <div>{{dateInWords(date)}}</div>
+</template>
+
+<script>
+import { mapGetters } from 'vuex'
+import distanceInWordsToNow from 'date-fns/distance_in_words_to_now'
+
+const locales = {
+  en: require('date-fns/locale/en'),
```
nicksellen: can you try loading the locales on-demand? we try and keep the bundle sizes as small as possible.

we do it for the app transations here ---> https://github.com/yunity/karrot-frontend/blob/master/src/store/plugins/i18n.js#L10

src/components/General/DateAsWords.vue
```diff
@@ -0,0 +1,39 @@
+<template>
+  <div>{{dateInWords(date)}}</div>
```
nicksellen: A tooltip showing the full date would be great too

nicksellen: (one of these works nicely http://quasar-framework.org/components/tooltip.html)

nicksellen:
Looks great! Just a couple of comments.

package.json
```diff
@@ -37,6 +37,7 @@
     "babel-plugin-dynamic-import-node": "^1.1.0",
     "babel-runtime": "^6.25.0",
     "clone-deep": "^1.0.0",
+    "date-fns": "^1.29.0",
```
tiltec: Remember to add `yarn.lock` too before merging

src/components/General/DateAsWords.spec.js
```diff
@@ -0,0 +1,24 @@
...
+})
+
+import DateAsWords from './DateAsWords'
+import { mountWithDefaults } from '>/helpers'
+import store from '@/store'
+
+describe('DateAsWords', () => {
+  it('renders 1 day ago', () => {
+    const now = new Date('2017-08-12T12:00:00Z')
+    Date.now = jest.genMockFunction().mockReturnValue(now)
```
nicksellen: Is this different to using `jest.fn()`?

tiltec:
Looks really good, looking forward to merge this!

nicksellen:
Related issue is https://github.com/yunity/karrot-frontend/issues/311

package.json
ReneCode: oh, seems that I have to install yarn (normaly I'm working with npm)

nicksellen: Yup, we ran into a lot of issues with npm 5 (see https://github.com/yunity/karrot-frontend/issues/543)

src/components/General/DateAsWords.spec.js
ReneCode: I've seen an example with "getMockFunction()" - but you're right: jest.fn() does the same and looks simpler.

PR #699, opened 2017-11-05
i18n cleanup, first round

tiltec:


src/locales/locale-en.json
```diff
@@ -50,7 +36,11 @@
   },
   "FORM": {
```
nicksellen: I would name this `VALIDATION` personally

tiltec: makes sense, will do

scripts/find_unused_locales.js
```diff
@@ -0,0 +1,41 @@
+/**
```
nicksellen: We assume grep below, so might as well assume unix env, and start the file with:
```
#!/usr/bin/env node
```
then remove the .js extension and `chmod +x` it, then can execute it directly...

PR #705, opened 2017-11-05
Convert and update locales

tiltec:
- finally converted locale-en from angular format to vue-i18n format (using single `{`) and removed the hotfix code
- pushed locale-en to transifex to get translators active again
- pulled translations from transifex to avoid vue-i18n warnings about missing translations
- add new Czech locale, even if it only has a few strings translated


tiltec:
There's actually not so much to review here, will merge now.

Issue #668, opened 2017-10-30
Rationalize validation and error handling


tiltec:
groupEdit and storeEdit are already in quite a good shape, need to bring the other forms there too.

Add client-side validation (vuelidate)
- [ ] pickupEdit
- [ ] pickupSeriesEdit
- [x] agreementEdit
- [ ] login
- [ ] passwordReset (check if email?)
- [ ] signup
- [ ] ProfileEdit
- [ ] Settings.vue (require display name...)
- [ ] changeEmail

Add server validation error handling
- [x] invitationsUI (does only nonFieldErrors right now)
- [ ] agreementEdit (quite complex, won't do it)
- [x] groupInfoCard (only shows "password wrong", without checking)
- [x] login (needs rework, see #658)
- [x] ProfileEdit
- [x] changePassword
- [x] changeEmail
- [x] WallInput
- [x] VerificationWarning
- [x] verifyEmail

Things to look out for:
- disable submit if request is already in progress
- add visual feedback if request is in progress (loader, spinner)

Other related tasks:
- [x] storeedit, groupedit need a spinner
- [x] pickup (series) edit need a `canSave` function refactored

tiltec:
Also need to handle server permission errors, e.g. the server returns 403 with 

```json
{ "detail": "You are already a member." }
```

tiltec:
Talked a bit about it with @nicksellen 

    two kinds of errors: validation (HTTP400-499) and other errors
    Validation errors are caught and handled inside the store as data
    Other errors are handled in interceptors
    Goals: other errors end up in Sentry, no validation errors end up in sentry
    
    DRF: field errors, nonFieldErrors (of validation errors), detail (of permission errors)
    
    Now we have three kinds:
        - validation errors - the user can handle them
        - permission errors - the user can maybe handle them, maybe not. Should probably not make the request in the first place (out-of-date information)
        - other errors
        
    Origin of errors:
        - our of date auth information
        - followed wrong link
        - programming error
        
    DRF HTTP errors:
        - 400: returns field or nonFieldError
        - 401, 403: returns detail

tiltec:
#691 is related, most request handling happens in vuex stores

tiltec:
We currently don't use the `detail` field of `401` and `403` responses at all.

tiltec:
This is partially dependent on #626, it will introduce a nice and uniform way to access validation errors across all vuex modules.

tiltec:
[...]

PR #712, opened 2017-11-06
add changeEmail component

### LINKED ISSUE #698: Missing change-email frontend
The settings page currently only says "Mail" at the bottom, should add the functionality to let users change their e-mail address.

tiltec:
I moved the verificationWarning closer to changeEmail in settings

image

Closes #698

### DIFF
--- a/src/components/Settings/ChangeEmail.vue
+++ b/src/components/Settings/ChangeEmail.vue
@@ -1,0 +1,59 @@
+<template>
+  <div class="edit">
+    <VerificationWarning v-if="!user.mailVerified" />
+
+    <q-field
+      icon="fa-star"
+      :label="$t('USERDATA.EMAIL')">
+      <q-input type="email" v-model="newEmail"/>
+    </q-field>
+
+    <q-btn color="primary" @click="save">{{ $t('BUTTON.CHANGE_EMAIL') }}</q-btn>
+
+  </div>
+</template>
+
+<script>
+import { QField, QInput, QBtn } from 'quasar'
+
+import VerificationWarning from '@/components/Settings/VerificationWarning'
+
+export default {
+  components: { QField, QInput, QBtn, VerificationWarning },
+  props: {
+    user: { required: true },
+  },
+  data () {
+    return {
+      newEmail: '',
+    }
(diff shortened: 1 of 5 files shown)


nicksellen:
I know this is just about the frontend component but I got a bit confused with the change email functionality:
- changing it to the email address of another user gives a 400 error, but no error is shown on the page, additionally the field keeps the changed email address so it looks like it worked
- I changed it to something else, and it seems like it was immediately updated as it only showed the changed one (although I still have to login with the old email address which is not shown)
- I couldn't click on the verification email link as they are dummy email accounts, but also could not change the email address back to what it was before 

Although as it has `email` and `unverifiedEmail` address fields it seems like it should support this.

I think this is one of the UI/UX things we would benefit from just copying from <some popular site> (airbnb, etc...).

For now though, maybe just showing that there is a pending email address change separately from the current email address?

tiltec:
You're right, the change-email implementation in some ways.

> changing it to the email address of another user gives a 400 error, but no error is shown on the page, additionally the field keeps the changed email address so it looks like it worked

True, I was kind of waiting for #668 before adding it. We could leave this PR around until #668 is done and the add the necessary bits, or implement it in the "old" boilerplate-y way (I'm not so motivated right now), or just leave it out for now.

> I changed it to something else, and it seems like it was immediately updated as it only showed the changed one (although I still have to login with the old email address which is not shown)

Ah right, this is one of this bits I forgot to copy from the old angular page. I think we should show the old verified email in the email field. If there is no verified email, show the unverified one.

> could not change the email address back to what it was before

I discovered this too yesterday, it seems like a backend issue. Will open an issue there.

> I think this is one of the UI/UX things we would benefit from just copying from (airbnb, etc...).

That would be great. I think this was one of the features where I went into the trap thinking of an "uninteresting but necessary and easy-to-implement" thing and getting surprised by the huge number of states that need to be handled. Maybe also "django-allauth" or similar could help us there.

tiltec:
Backend issue: https://github.com/yunity/karrot-backend/issues/430

tiltec:
I added the change-email thing to the list in #668

PR #714, opened 2017-11-06
Disable Sidenav on tablets if logged out

### LINKED ISSUE #708: Disable Sidenav on tablets if logged out
Currently, the sidenav is shown on mobile screens => 1024px also when logged out. Should be disabled.

tiltec:
Closes #708


nicksellen:
Why not hide the sidenav with the `v-if` for the left slot? Setting the breakpoint seems hacky...

tiltec:
I thought that it introduced a visual bug, but the bug seems to be there anyway:

image

PR #718, opened 2017-11-09
Add end-to-end tests with real backend

tiltec:
I enjoyed writing the end-to-end tests with TestCafe, they have quite nice selectors. The tests depend on each other because I didn't implement any database instrumentation. The tests should also work if there is content in the database, makes it easier to prototype tests locally without resetting the database after every run.
You can use either the dev frontend from `yarn dev`  or the production files from `yarn build && yarn serve`.

I didn't manage to get the CI integration working yet. I tried running everything in one container, but the circleci prebuilt images don't have yarn and python together and I couldn't install it via apt-get. I tried running docker-compose, but couldn't access the backend via `localhost:8000`.

I have two options now:
- use another image that has yarn + python or at least a usable package manager
- build a custom docker image
- figure out networking with remote docker and docker-compose

I'll try the first one, as it seems easier.


tiltec:
Now I get

```
Downloading workspace layers
Error downloading workspace layer for job 3d696238-6468-4a13-9332-077b5e947828: RequestError: send request failed
caused by: Get https://circle-production-customer-artifacts.s3.amazonaws.com/?list-type=2&prefix=picard%2F57968bbac9e77c000197c573%2Fworkflows%2Fworkspaces%2Fa6f23dde-d19f-472f-81d6-6d6c64c92b74%2F0%2F3d696238-6468-4a13-9332-077b5e947828: x509: failed to load system roots and no roots provided
```

in circleci

tiltec:
Progressing with the first option, now it's stuck at connecting to the database.

I think it would make sense using the docker-compose method though, it gives us the opportunity to easily use the maildev API to verify sent e-mails.

tiltec:
Yay, seems to work! Running the browser on saucelabs :)

tiltec:
Seems ready for review!
I just run in one browser on saucelabs (Chrome 61 on Windows 10) for now, adding more browsers should be as easy as uncommenting them in `testcafe.js`.
Test duration is quite long now unfortunately, most because of backend setup and saucelabs connect.

To dos for the future:

- [ ] use docker-compose to make backend setup faster and have maildev available
- [ ] run chromium and firefox tests directly on circleci to speed up test execution
- [ ] enable safari test on saucelabs

tiltec:
Oh come on Node Security, make it green!

.circleci/config.yml
```diff
@@ -31,6 +43,62 @@ jobs:
       - run: yarn test --ci --runInBand
       - run: bash <(curl -s https://codecov.io/bash)
 
+  install-e2e-dependencies:
+    <<: *e2e-defaults
```
nicksellen: Doesn't seem to need to startup redis/postgres for the prepare job, which seems to take up >1 minute https://circleci.com/gh/yunity/karrot-frontend/4232

.circleci/config.yml
```diff
@@ -103,6 +171,13 @@ workflows:
       - test:
           requires:
             - install-dependencies
+      - install-e2e-dependencies:
+          requires:
+            - install-dependencies
+      - e2e-test:
```
nicksellen: Is this reasonable to run on every build? Or should we put it behind a hold step?

e2e/testcafe.js
```diff
@@ -1,33 +1,41 @@
...
-  await t.takeScreenshot()
-  await t.click(loginbutton) // pressing enter does not submit the form on "Chrome@61.0:OS X 10.9"
-  await t.takeScreenshot()
-  const location = await t.eval(() => window.location)
-  await t.expect(location.hash).eql('#/group/1/wall')
-  const title = await t.eval(() => document.title)
-  await t.expect(title).eql('05_testgroup | Karrot')
-})
+createTestCafe('localhost')
+  .then(tc => {
```
nicksellen: No async/await available here? I know top-level is not normally possible, but can do it with an IIFE

src/components/Group/GroupEdit.vue
```diff
@@ -4,14 +4,15 @@
     <h3 v-else><i class="fa fa-edit"></i> {{ $t('GROUP.EDIT') }}</h3>
     <q-card>
       <div class="edit" :class="{ changed: hasChanged }">
-        <form @submit="save">
+        <form @submit.prevent="save">
```
nicksellen: Ah good, I noticed this was doing a full page submit recently :)

.circleci/config.yml
```diff
@@ -31,6 +43,62 @@ jobs:
...
+            apt-get install -y python3 python3-dev virtualenv
+            cd karrot-backend
+            env/bin/python manage.py migrate
+            env/bin/python manage.py runserver
+          background: true
+      - run:
+          name: run frontend
+          command: yarn serve
+          background: true
+      - run: wget --retry-connrefused -T 30 http://localhost:8080
```
nicksellen: Maybe document what this is for, I know, but could look odd for people that don't....

.circleci/config.yml
```diff
@@ -31,6 +43,62 @@ jobs:
...
+      - run: wget --retry-connrefused -T 30 http://localhost:8000/api/groups-info/
+      - run:
+          name: list all available browsers
```
nicksellen: What is this for?

build/script.serve.js
```diff
@@ -0,0 +1,47 @@
...
+Object.keys(proxyTable).forEach(function (context) {
+  var options = proxyTable[context]
+  if (typeof options === 'string') {
+    options = { target: options }
+  }
+  app.use(proxyMiddleware(context, options))
+})
+
+// handle fallback for HTML5 history API
+app.use(require('connect-history-api-fallback')())
```
nicksellen: Why this? We don't use html5 history stuff...

e2e/tests.js
```diff
@@ -0,0 +1,66 @@
+import { Selector, Role } from 'testcafe'
+const testTime = Math.floor(new Date() / 1000)
+console.log(testTime)
```
nicksellen: Seems not needed?

.circleci/config.yml
tiltec: True. Although I think displayed times are not accurate because it also counts in the time until the required steps are ready and also if it's limited by the number of VMs (max 4 in parallel).

.circleci/config.yml
tiltec: Just for informational purposes. For some reason, I can't access the list of available browsers on saucelabs on my machine, but it works in the CI. Might also be helpful if someone wants to change the tested browsers without setting up a saucelabs account.

.circleci/config.yml
tiltec: I think it's reasonable to run it before merging a pull request, so probably every time when a PR is open. I know it currently takes much time to run, so putting it behind a button could be a intermediate measure before speeding it up.

tiltec: Maybe I'll put it behind the deploy-to-production approval just for simplicity?

build/script.serve.js
tiltec: I copied it from `script.dev.js`, does it make sense to remove from both, to have the same setup for serving dev and built files?

e2e/testcafe.js
tiltec: oops, just copied it from some testcafe example ;)

src/components/Group/GroupEdit.vue
tiltec: I think it's needed in more places, will add an issue :)

.circleci/config.yml
nicksellen: Would seem reasonable for now.

build/script.serve.js
nicksellen: Yes, as long as we remember if we switch to fancy urls in the future (which I would like... just not high priority)

nicksellen:
I think the last remaining thing is just making it not run all the time, until it is a bit faster.

Lots of room for speed improvements though :)

build/script.serve.js
tiltec: Hm, then I'd rather leave it in. It's more of a "minimal change" to me if "yarn dev" and "yarn serve" behave the same.

PR #719, opened 2017-11-10
add slightly improved history detail view

### LINKED ISSUE #715: Add detail view to history entries
The history payload contains a wealth of information that are both helpful for users and for giving support to users (figure out what happened etc).

It is currently not accessible by the user anymore, probably due to the redesign.

@alangecker would you be up for fixing this?

tiltec:
Closes #715 
I spent some time making it look half-decent. I still remember Nick saying that dev-only UIs should look as basic as possible to make it clear that it's intentional and not a design mistake. My previous history detail view was shitty enough I hope, still it was removed by Chandi without replacement. 

I think having _some_ history detail view is helpful for providing user support, so I spent some hours today making it look okayish. What do you think?

image

Still to do, probably for later:
- nice handling of the `payload` information
- hide raw data if all important content is shown otherwise
- provide tooltips for icons
- add links to related pages (group, store, pickup collectors...)


src/store/modules/history.js
```diff
@@ -28,6 +31,9 @@ export const state = initialState()
 
 export const getters = {
   get: (state, getters, rootState, rootGetters) => id => {
+    if (!state.entries[id]) {
+      return undefined
```
nicksellen: can just be `return` (is implicitly `undefined`) and seems reasonable to make into a oneliner...

src/store/modules/history.js
```diff
@@ -77,6 +92,9 @@ export const actions = {
     }
     commit(types.RECEIVE, { entries: data.results, cursor: data.next })
   },
+  async fetchById ({ commit, state }, id) {
+    commit(types.RECEIVE, { entries: [await historyAPI.get(id)], cursor: state.cursor })
```
nicksellen: This seems confusing :/ It will add the id to the idList no?

src/pages/HistoryDetail.vue
```diff
@@ -0,0 +1,13 @@
+<script>
+import { connect } from 'vuex-connect'
+import HistoryDetail from '@/components/History/HistoryDetail'
+
+export default connect({
+  gettersToProps: {
+    entry: 'history/active',
+  },
+  lifecycle: {
+    mounted: ({ dispatch, getters }) => dispatch('history/setActive', getters['route/params'].historyId),
```
nicksellen: I guess this has the same non-reactive problem as the history list did before? If the route params change, this will not update... how does it avoid it?

tiltec: It does not work. I added a new commit to work on it, but it doesn't work and I don't know how to solve it. Maybe I'd have to get rid of vuex-connect.

src/store/modules/history.js
tiltec: Yes, I'll get rid of it

src/pages/HistoryDetail.vue
nicksellen: Yup, that's what had to happen with the history stuff. I think getting rid of vuex-connect might be the only way :/ (because then you can add watchers using a normal component, which vuex-connect could let you do in theory... but does not seem officially supported so would be unwise to hack in imo).

tiltec: I fixed it by putting the loading into the router vuex plugin, which seems our currently best solution for handling this.
(I also noticed that the history list view shows sometimes wrong entries when using back navigation, need to investigate more)

PR #722, opened 2017-11-11
Rationalize vuex requests

nicksellen:
A did it in a seperate branch/PR so we can compare the approaches better (https://github.com/yunity/karrot-frontend/pull/693 is @tiltec 's initial approach, this builds on that)


nicksellen:
Yes, I worried about the magic a bit (.. and would also have to change some of the current actions, as some of them have `{ groupId }` argument.

I'll have a look at using the getters in more places... it's also possible to add the state into the `byAction` part even when it has an id, either adding the details per id inside that (an earlier iteration had that), or just include the last one (which needs a bit of handling for overlapping action calls).

Actually, having some handling like this means we can do some stuff like disallow individual actions to be run concurrently (or, if we had a cancellable promises implementation, let the most recent win, etc...).

I'll come up with something for now...

nicksellen:
Phew, I ended up touching quite a lot of things. I wrote some docs about it too in `docs/forms.md` which might be worth reading (they don't deploy for branches so would have to look locally).

I focused on pickups and groups (forms for create/save, and join/leave loading things), would do pickup series before I consider it ready to merge. I'm quite happy with the approach, I manage to remove quite a few lines in the vuex modules and end up with more features.

I didn't really address the `id` magic, but did add a way to prefix the `id` if the auto-detected id isnot for the main type, e.g.:

```
...withPrefixedIdMeta('agreements/', {
```
The other more general thing I did was to use the `__unenriched` field in the form for groups too (it's essential actually to get good validation and reset behaviour, docs explain a bit more).

What I most liked was I could just move the existing actions into a `withMeta` block and remove any mutation commits related to request stuff, and it doesn't really matter if you add almost everything in there either.

A few todos off the top of my head:
- [ ] make sure the axios interceptors are handling exactly the errors that the meta is not
- [ ] implement new style for pickup series

Things I would not address in this PR, but would like to in another:
- forms that are not pickup/group related
- rationalize approach for forms (pickup/group/pickupseries have enough duplication to abstract I think)

docs/forms.md
```diff
@@ -0,0 +1,138 @@
...
+
+Our approach for forms is to clone an _editing_ version of the data for use in the form. This allows us to:
+- know if the value of the data has changed (`diff(source, edit)`)
+- reset the form (`edit = source`)
+- emit the diff of changes in a `@save` event (`emit(edit)`) (works nicely with http `PATCH`)
+- handle underlying changes in the source value as we wish
+
+In many places we enrich the data with additional fields,
+and we need to use the original data (available as `__unenriched`) to correctly only detect changes in the
+underlying value, not the derrived data.
```
tiltec: typo: derrived

src/store/helpers.js
```diff
@@ -26,3 +28,173 @@ export function onlyHandleAPIError (error, handleFn) {
...
+}
+
+export function metaStatusesWithId (getters, actions, id) {
+  const result = {}
+  for (let action of actions) {
+    result[action + 'Status'] = getters['meta/status'](action, id)
+  }
+  return result
+}
+
```
tiltec: Building identifiers via string replacement makes "grepping" around in the codebase harder. I think it could be worthwhile in this case, although I would be cautious to continue metaprogramming like this.

src/store/modules/groups.js
```diff
@@ -1,54 +1,30 @@
 import groups from '@/services/api/groups'
 import groupsInfo from '@/services/api/groupsInfo'
 import router from '@/router'
-import { indexById, onlyHandleAPIError } from '@/store/helpers'
+import { indexById, withMeta, createMetaModule, withPrefixedIdMeta, metaStatusesWithId, metaStatuses } from '@/store/helpers'
 import { Toast } from 'quasar'
 import i18n from '@/i18n'
 
-export const types = {
+// const meta = defineRequestModule()
```
tiltec: remove comment

src/store/modules/groups.js
```diff
@@ -110,241 +86,156 @@ export const getters = {
...
-    }
-    catch (error) {
-      router.replace('/notfound')
-      return
-    }
+      commit(types.SET_ACTIVE, { groupId })
 
-    dispatch('pickups/clear', {}, { root: true })
+      try {
+        await dispatch('fetchGroup', groupId)
```
tiltec: merge conflict ahead! I worked on this bit today.

test/helpers.js
```diff
@@ -63,10 +63,56 @@ export function polyfillRequestAnimationFrame () {
 
 export function mountWithDefaults (Component, options = {}) {
   const localVue = createLocalVue()
+  // localVue.productionTip = false
```
tiltec: not needed?

src/__mocks__/vue.js
```diff
@@ -0,0 +1,3 @@
+const vue = require.requireActual('vue')
```
tiltec: Wohoo, wizardry!

src/store/helpers.js
nicksellen: I agree. Hopefully this metaprogramming for metadata is perfect for this usecase, so it hides away the crap, but the main actual data can just still use normal actions/mutations.

src/store/modules/groups.js
nicksellen: Fixed

docs/forms.md
nicksellen: Fixed (plus found a bunch more errors in there)

test/helpers.js
nicksellen: Indeed, fixed

src/store/helpers.js
```diff
@@ -26,3 +28,173 @@ export function onlyHandleAPIError (error, handleFn) {
...
+    return true
+  }
+  if (error.request) {
+    return true
+  }
+  return false
+}
+
+/**
+ * Defines a vuex module that can be used to handle async request metadata.
```
nicksellen: Should remove this set of docs, and just link to the gitbook docs

src/store/helpers.spec.js
```diff
@@ -0,0 +1,72 @@
...
+      expect(run).toBeCalled()
+      expect(store.getters['meta/status']('run', id).pending).toEqual(false)
+    })
+
+    it('handles validation errors', async () => {
+      const validationErrors = { foo: 'bar' }
+      run.mockImplementationOnce(() => { throw createValidationError(validationErrors) })
+      await store.dispatch('run', { id })
+      expect(run).toBeCalled()
+      expect(store.getters['meta/status']('run', id).pending).toEqual(false)
```
nicksellen: Need to check the `validationErrors` are set here

src/store/modules/auth.js
```diff
@@ -3,6 +3,12 @@ import authUser from '@/services/api/authUser'
 import router from '@/router'
 import { onlyHandleAPIError } from '@/store/helpers'
 
+/*
```
nicksellen: Remove commented out section

test/helpers.js
```diff
@@ -70,3 +70,48 @@ export function mountWithDefaults (Component, options = {}) {
...
+  })
+}
+
+export function mockActionOnce (store, actionName) {
+  const originalValue = store._actions[actionName]
+  const restore = () => { store._actions[actionName] = originalValue }
+  const mockFn = jest.fn()
+  store._actions[actionName] = [async () => {
+    try {
+      await mockFn()
```
nicksellen: Should return the value so it still works like the real action

tiltec:
Looking forward to merge this soon 🐣

nicksellen:
Ok, should be good now!

Things I just worked on:
- using the new meta action status for pickup series edit
- created formMixin to contain duplicated logic (documented in docs/forms.md too), could be rolled out to more forms, but I think 2 is enough for right now...

src/store/helpers.spec.js
tiltec: Is this done?

nicksellen: Now it is :)

PR #723, opened 2017-11-11
Improve wall ui

### LINKED ISSUE #635: Improve wall UI
(aka _make the wall look sexy_)

See https://github.com/yunity/karrot-frontend/pull/630 and https://github.com/yunity/karrot-frontend/pull/634.

tiltec:
image

Closes #635 
Fixes most of the points:

- [x] message is most prominent
- [x] wall input is now part of the list
- [x] ctrl+enter to submit works

### DIFF
--- a/src/components/Wall/Wall.vue
+++ b/src/components/Wall/Wall.vue
@@ -20,11 +20,12 @@
       {{ messageReceiveStatus.error }}
     </q-alert>
     <template v-if="messageReceiveStatus.success">
-      <WallInput :status="sendStatus" @send="$emit('send', arguments[0])" :placeholder="messagePrompt" />
       <q-infinite-scroll
         :handler="loadMore"
         ref="infiniteScroll">
-        <q-list :highlight="false" inset-separator class="bg-white desktop-margin" v-if="messages.length > 0">
+        <q-list :highlight="false" inset-separator class="bg-white desktop-margin">
+          <WallInput :status="sendStatus" @send="$emit('send', arguments[0])" :placeholder="messagePrompt" />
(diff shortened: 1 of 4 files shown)


nicksellen:
It's awkward that the line seperating the messages does not fully extend across the width.

Also, I find the spacing in the top right a bit off. Too close to the avatar, and the avatar itself isn't lined up with anything vertically. Perhaps including the date in the bottom right?

Maybe it doesn't need the message icon in the top left too? For short messages the text is shoved into the bottom left corner, and loads of whitespace elsewhere.

tiltec:
Agree, will work on those bits.

I tried to move the date to the bottom, didn't succeed. I couldn't make `q-item-side` the same height as `q-item-main`.

tiltec:
image

tiltec:
I don't like the horizontal lines too much, but striped also looked odd. Maybe we just merge it for now until we have better ideas?

@nicksellen 
I could move the timestamp on top of the message, would that be better?

nicksellen:
I'm not actually sure it's an improvement though :p

Here's a bit of facebook ui, which is maybe a good inspiration:
- name of poster inline with the message
- timestamp on bottom (left aligned with author/message)
- image in left column
- full width line

tiltec:
Current status

image

nicksellen:
Oooh, beautiful! Everything is nicely aligned and spaced :)

nicksellen:
I think you didn't push the code yet?

tiltec:
image

I wanted to put the profile picture on the input field as well :)

Issue #638, opened 2017-10-27
Ensure redirects are handled correctly


nicksellen:
A few cases that may not be handled correctly right now:
- user visits a group detail page they are not a member of
- user visits a page that does not exist at all (no route)
- item does not exist (e.g. group id that does not exist)
- other things like this (but for the store)

In general some of this stuff should/is be handled, but is maybe not completely correct right now. Should also have some tests unit and/or e2e.

tiltec:
As e2e tests seem to be a bit too complicated to add in the next week, I would try to go through obvious cases and test it manually.

tiltec:
## Currently handled quite ok

> invalid group info id

https://dev.foodsaving.world/#/groupInfo/99999
redirects to /notfound

> user visits a page that does not exist at all (no route)

Shows 404

## handled, but might be confusing

> user visits a group detail page they are not a member of

https://dev.foodsaving.world/#/group/9999/wall
Redirects to  currentGroup. I was confused to see the wall of a different group, although there was a very short toast. Does not change the URL in the browser.

## Currently not handled at all (shows empty or half-empty page)

- [ ] invalid user id: https://dev.foodsaving.world/#/user/999
- [ ] invalid store id: https://dev.foodsaving.world/#/group/6/store/9999/pickups

tiltec:
I think the method of redirecting to `/notfound` is a ugly workaround. It should rather keep the URL in the browser, but show the most detailed error message possible ('not a member of the group', 'route invalid', 'store id invalid')

tiltec:
Now that we have the `routeError` vuex module, we should use it for all kinds of "id not found/accessible" errors

PR #727, opened 2017-11-12
add route error page

tiltec:
Instead of redirecting to `/notfound`, keep the same route and switch the router view with the error page.
Can show custom messages (and perhaps more complex components via slot), controlled by vuex.

Part of #638


src/store/modules/groups.js
```diff
@@ -164,8 +162,9 @@ export const actions = {
 
     const userId = rootGetters['auth/userId']
     if (!group.members.includes(userId)) {
-      Toast.create.warning(i18n.t('GROUP.NONMEMBER_REDIRECT'))
-      throw new Error('user is not a member')
+      const message = i18n.t('GROUP.NONMEMBER_REDIRECT')
+      dispatch('routeError/set', message, { root: true })
```
nicksellen: Better to set the translation key, not the actual translation text? (otherwise will not update if locale changes)

tiltec:
Ready!

PR #733, opened 2017-11-12
Make date as words reactive

### LINKED ISSUE #729: Time in distance is not always reactive
Many times it seems to work ok, but sometimes it does not update:
- visit https://dev.karrot.world/#/group/13/wall (or any other group)
- notice the _time in distance_ content on the wall
- switch the language
- notice many things change, but not the time in distance

nicksellen:
Fixes #729 

I made it reactive in two ways:
1. when the locale changes
2. when _now_ changes (as it is relative to now)

It updates now every 10 seconds, and if there were _load_ of components on the page doing this it could get inefficient (don't really need to be updating "7 months ago" every 10 seconds to see if it's now "8 months ago"), but it is fine for now.

I used a nice technique I didn't know about before, you can create vue instances (`new Vue({})`) just to use the reactivity features of them.


src/components/General/DateAsWords.vue
```diff
@@ -1,23 +1,25 @@
 <template>
   <div>
-    {{dateInWords(date)}}
-    <q-tooltip>{{ $d(new Date(date), 'long') }}</q-tooltip>
+    {{ dateInWords }}<q-tooltip>{{ tooltipContent }}</q-tooltip>
```
tiltec: what happened to the newline?

src/services/dateFnsHelper.js
```diff
@@ -0,0 +1,25 @@
...
+    locale: 'en',
+    locales: {},
+    now: new Date(),
+  },
+  created () {
+    setInterval(() => { this.now = new Date() }, 10 * 1000)
+  },
+  watch: {
+    async locale (locale) {
+      if (locale === 'zh') locale = 'zh_cn' // https://date-fns.org/v1.29.0/docs/I18n
```
tiltec: Is zh the only local that needs this or can we map the locales more generally?

src/components/General/DateAsWords.vue
tiltec: The CI failure seems related

nicksellen: It is not related, they are "contains" matching. The test failure is because the way of faking the now date no longer works (actually it was a bit broken before as it never reset itself). I'm working on fixing up the tests, and adding some more now :)

src/services/dateFnsHelper.js
nicksellen: Yes, is the only one. I switched it to traditional chinese now actually, that seems to be the preferred version in Taiwan according to the internet.

nicksellen:
I wanted to write a test that advanced the timers some minutes and saw that the rendered text changes, but I failed :/

src/services/dateFnsHelper.js
tiltec: Isn't there something like cn_TW?

nicksellen: It's using `zh_tw'` now,  which is called "Chinese Traditional" by date-fns.

> The Simplified is the standard writing form employed in the mainland of China and the Traditional form is mainly used in Taiwan and Hong Kong. Although the majority of the population uses Simplified Chinese there is a trend towards returning to Traditional Chinese. 
says http://www.iwtservices.com/resources/chinese_sort.htm

It probably means we should account for these locales with regions at some point...

PR #734, opened 2017-11-12
Add metadata to index.html

tiltec:
Didn't test them yet.

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -1,22 +1,22 @@
 <!DOCTYPE html>
 <html>
   <head>
-    <meta charset="utf-8">
-    <meta name="format-detection" content="telephone=no">
-    <meta name="msapplication-tap-highlight" content="no">
-    <meta name="viewport" content="user-scalable=no, initial-scale=1, maximum-scale=1, minimum-scale=1, width=device-width">
-    <meta name="mobile-web-app-capable" content="yes">
-    <meta name="apple-mobile-web-app-capable" content="yes">
-    <meta name="apple-mobile-web-app-status-bar-style" content="black">
+    <meta charset="utf-8" />
+    <meta name="format-detection" content="telephone=no" />
+    <meta name="msapplication-tap-highlight" content="no" />
+    <meta name="viewport" content="user-scalable=no, initial-scale=1, maximum-scale=1, minimum-scale=1, width=device-width" />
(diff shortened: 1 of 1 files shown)


src/index.html
```diff
@@ -1,22 +1,25 @@
...
 
     <title>(karrot loading...)</title>
-    <link rel="shortcut icon" href="statics/carrot_logo.ico">
-    <link rel="icon" type="image/png" href="statics/carrot_logo.png" sizes="32x32">
+    <link rel="shortcut icon" href="statics/carrot_logo.ico" />
+    <link rel="icon" type="image/png" href="statics/carrot_logo.png" sizes="32x32" />
   </head>
   <body>
-    <div id="q-app"></div>
+    <div id="q-app" />
```
nicksellen: This is _real_ html here, and self-closing `<div>` is invalid. We can use it in the vue templates as it's loaded by a more relaxed parser of our choosing, this file is parsed by the parser of the unknown users unknown browsers choosing.

[html5 spec says](https://www.w3.org/TR/html5/syntax.html#start-tags) (about opening tags):

> Then, if the element is one of the void elements, or if the element is a foreign element, then there may be a single "/" (U+002F) character. This character has no effect on void elements, but on foreign elements it marks the start tag as self-closing.

- void elements _area, base, br, col, embed, hr, img, input, link, meta, param, source, track, wbr_
- foreign elements are _Elements from the MathML namespace and the SVG namespace._

src/index.html
```diff
@@ -1,22 +1,25 @@
...
   <head>
-    <meta charset="utf-8">
-    <meta name="format-detection" content="telephone=no">
-    <meta name="msapplication-tap-highlight" content="no">
-    <meta name="viewport" content="user-scalable=no, initial-scale=1, maximum-scale=1, minimum-scale=1, width=device-width">
-    <meta name="mobile-web-app-capable" content="yes">
-    <meta name="apple-mobile-web-app-capable" content="yes">
-    <meta name="apple-mobile-web-app-status-bar-style" content="black">
-    <meta name="description" content="karrot">
+    <meta charset="utf-8" />
```
nicksellen: Why self close all these `<meta>` elements, they are void elements, and it has no effect, see below.

src/index.html
tiltec: Ah good! Didn't know that, will fix/undo these changes.

PR #747, opened 2017-11-13
Multiline pickup description

### LINKED ISSUE #728: Pick-up comment layout
I wanted to add more information to a pick-up series and found the field to do so pretty small:
image
 Also, I was disappointed that I couldn't add any formatting at all: No paragraphs, no boldness, no italics.

Use case is that the Wednesday pick-up is normally done by Villa Klug and the Friday pick-up by us. I wanted to add that to the pick-up info, while pointing out that this doesn't mean that the pick-ups are reserved for only us and no other people are allowed to join...

It's not very evident like this:
image

tiltec:
Closes #728 

I implemented multiline support for pickup description with a frontend-only limit of 500 characters.

I found out that the pickupEdit and pickupSeriesEdit textarea didn't autogrow properly. The most likely reason is unfortunate: q-collapsible renders the child element and hides it, and the textarea has a q-resize-observable that doesn't work when hidden.
I worked around this issue by introducing a `visible` state into the pickupManage component. Not the best solution, maybe someone can come up with a better one?

### DIFF
--- a/src/components/Pickups/PickupEdit.vue
+++ b/src/components/Pickups/PickupEdit.vue
@@ -41,7 +41,7 @@
         :error="hasError('description')"
         :error-label="firstError('description')"
         >
-        <q-input v-model="edit.description" type="textarea" :min-rows="1" :max-height="100" />
+        <q-input v-model="edit.description" type="textarea" />
       </q-field>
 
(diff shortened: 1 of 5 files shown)


tiltec:
I'm tempted to file an issue in quasar about the autogrow issue.

Issue #726, opened 2017-11-12
Rationalize loading of route-dependent data


tiltec:
Currently it's a bit scattered around, some are loaded via `src/store/plugins/router.js`, [some via `mounted` lifecycle](https://github.com/yunity/karrot-frontend/blob/bdd7326c8edcc4dbbd017b994278a0223fe9d0a6/src/pages/Store/PickupsManage.vue#L167) or alike, and [some use `watch`](https://github.com/yunity/karrot-frontend/blob/bdd7326c8edcc4dbbd017b994278a0223fe9d0a6/src/components/History/HistoryList.vue#L35).

A common problem is that loading via `mounted` lifecycle is non-reactive to route changes when the component stays rendered. `watch` or `beforeRouteEnter` solves this problem, but it's not available when using vuex-connect, forcing us to write boilerplate. Loading with router plugin is a good solution, but it's detached from the page itself (coupling only via param id).

Ideally, we would load all data with `beforeEnter`/`beforeUpdate` hooks in the route definitions. I think the problem is that we can't easily use vuex there. Maybe we get inspired by how the breadcrumbs work, introducing a custom `resolve` object to the route definitions that runs in a router plugin.

nicksellen:
Yes, I think introducing another custom meta property that can do some vuex stuff would be a nice idea. I guess it needs to be a function so it can make use of route params, although perhaps being able to specify by default as a simple string (for action triggering) would be nice.

nicksellen:
Maybe a good approach is to remove all the complicated logic in `selectGroup` (and similar places) and put it into the routing logic and/or plugins, and maybe some changes to minimize cross-dependence between vuex modules.

I would like to get rid of things that look like procedural logic (fetch this, fetch this, fetch this, etc). It is more accurately expressed as declarative logic (when this changes, fetch this, when that changes, fetch that).

We want an approach that inherently works by design, not one that works by careful thinking about the ordering of statements.

tiltec:
Declarative approaches sound like a good idea, usually they need a good understanding of the requirements to be powerful enough without allowing unwanted side effects though.

Maybe a list of actions that need to be triggered while loading could be enough? Ideally, they shouldn't depend on each other, so redirects and detection of route errors should happen before. It's a bit tricky since this can depend on resolved data.

At some point, we have to trigger the fetching actions. I guess it's best do to this in a router afterEach hook (or beforeEach if redirects are processed first).

nicksellen:
[...]

PR #751, opened 2017-11-14
Trigger route param actions before route

### LINKED ISSUE #750: Pickup series displays from all stores if loaded directly
Similar to #746, would be happy to wait for #726 

Can see it on this page --> https://dev.karrot.world/#/group/4/store/70/pickups/manage

nicksellen:
Fixes #750 because it relies on a `mounted()` lifecycle hook.

(Still need to address this more generally in #726)

### DIFF
--- a/src/store/plugins/router.js
+++ b/src/store/plugins/router.js
@@ -47,9 +47,7 @@ export default store => {
     }
   })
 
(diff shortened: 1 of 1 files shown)


tiltec:
It shouldn't break anything, let's try it out!

PR #752, opened 2017-11-14
Polyfill Intl via service if needed

### LINKED ISSUE #748: Internationalization API is not available on some mobile browsers
We received some errors today from Chrome 51 on Android 7. It seems that it doesn't support Intl, although it should:
https://caniuse.com/#feat=internationalization

When digging further, it seems to be a common problem on Samsung phones, which can be worked around by providing a polyfill:
https://github.com/Financial-Times/polyfill-service/issues/509

I'm not a huge fan of bloating our codebase with polyfills, but maybe a polyfill service could help?
https://polyfill.io/

https://sentry.io/foodsaving-worldwide/karrot/issues/403966409/

```
ReferenceError: Intl is not defined
  at _localizeDateTime (./node_modules/vue-i18n/dist/vue-i18n.esm.js:1333:1)
  at _d (./node_modules/vue-i18n/dist/vue-i18n.esm.js:1351:1)
  at apply (./node_modules/vue-i18n/dist/vue-i18n.esm.js:1391:1)
  at $d (./node_modules/vue-i18n/dist/vue-i18n.esm.js:194:1)
  at call (js/0.463cada66b1a8d3b51c0.js:16005:19)
...
```

nicksellen:
Docs at docs/build.md

Fixes #748

### DIFF
--- a/build/plugins/IntlPolyfillServicePlugin.js
+++ b/build/plugins/IntlPolyfillServicePlugin.js
@@ -1,0 +1,39 @@
+const { readFileSync } = require('fs')
+const { ConcatSource } = require('webpack-sources')
+
+const checkIntlPolyfillSource = readFileSync(require.resolve('./checkIntlPolyfill'), { encoding: 'utf8' }).trimRight()
+
+/**
+ *   A simpler version of https://github.com/SebastianS90/webpack-polyfill-injector that only cares about Intl
+ */
+class IntlPolyfillServicePlugin {
+  apply (compiler) {
+    compiler.plugin('compilation', compilation => {
+      compilation.plugin('optimize-chunk-assets', (chunks, callback) => {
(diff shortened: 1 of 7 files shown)


nicksellen:
Ah, was using node version 6, and didn't like the trailing comma (which our eslint enforces...). I just updated netlify to use node 8. Fixed now :)

nicksellen:
> not that I understand every bit of it.
could I document some more to improve that?

tiltec:
Your code seems fine, I'm only missing knowledge about webpack (does it attach the code to every chunk? is a chunk the same as a .js file?)

nicksellen:
I wondered that too, it just got added to the `app.<hash>.js` file.

I think it's because it rejects chunks that don't have an entry module:

```js
if (!chunk.hasEntryModule()) return 
```

PR #753, opened 2017-11-14
Detect if keyboard is open and hide mobile nav

### LINKED ISSUE #731: Typing text on mobile often hides the typed text below navbar
Some ideas how to solve it:
- always scroll to input on focus
- scroll to cursor when entering a newline
- hide bottom navbar when typing (= focus on input)

screenshot_20171112-195045

nicksellen:
Fixes #731

Logic is (only if `Platform.is.mobile` is already true):
- collect the height+width of the window on load (using quasar height/width dom utils on the window)
- listen for resize changes on the window
- on resize look if we have a different height+width, if so we conclude it must be because the keyboard is open!

(calculating width+height is to handle the case of rotation, those that value will stay the same)

There might be a better way to use the information about client/window dimensions (see http://ryanve.com/lab/dimensions/) but this approach WorksForMe right now :)

### DIFF
--- a/src/main.js
+++ b/src/main.js
@@ -20,11 +20,11 @@ import './socket'
 import i18n from './i18n'
 import log from '@/services/log'
 import './raven'
-import { KeyboardPlugin } from '@/services/keyboard'
+import { DetectMobileKeyboardPlugin } from '@/services/detectMobileKeyboard'
 
 Vue.config.productionTip = false
 Vue.use(Quasar)
-Vue.use(KeyboardPlugin)
+Vue.use(DetectMobileKeyboardPlugin)
 
 if (process.env.NODE_ENV !== 'production') {
   log.setLevel('debug')
--- a/src/main.js
+++ b/src/main.js
@@ -20,9 +20,11 @@ import './socket'
 import i18n from './i18n'
 import log from '@/services/log'
 import './raven'
+import { KeyboardPlugin } from '@/services/keyboard'
 
 Vue.config.productionTip = false
 Vue.use(Quasar)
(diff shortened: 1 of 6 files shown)


nicksellen:
Hmm, maybe needs a bit more work, doesn't seem to work on https://hide-mobile-nav-if-keyboard-active--karrot.netlify.com/

nicksellen:
Oh, sort of works, but I think it can pickup the incorrect original screen size... so maybe need a fancier method

nicksellen:
Ok, switching to `document.documentElement.client*` values seems to work better on mobile chrome (changes when keyboard appears/hides, does not change when address bar appears/hides).

I made http://nicksellen.co.uk/keyboard.html to try this stuff out.

However, not so on firefox mobile. In both cases though I think the amount of change is clear enough. My keyboards are 227px high, and the address bar 48px. Maybe >150px is a reasonable split.

nicksellen:
Ok, implemented that and it works in chrome/firefox mobile browsers, so I think it's approximately ok now.

src/main.js
```diff
@@ -20,9 +20,11 @@ import './socket'
 import i18n from './i18n'
 import log from '@/services/log'
 import './raven'
+import { KeyboardPlugin } from '@/services/keyboard'
 
 Vue.config.productionTip = false
 Vue.use(Quasar)
+Vue.use(KeyboardPlugin)
```
tiltec: I propose DetectTouchKeyboardPlugin to make it clearer

nicksellen:
Ok, I renamed and added tests :)

PR #755, opened 2017-11-15
Refactor auth (login, changeEmail, changePassword) to use meta module

### LINKED ISSUE #658: Rework Login.vue error handling (TypeError: this.status.error[i(...)(...)[0]] is undefined)
https://sentry.io/foodsaving-worldwide/karrot/issues/394634594/

```
TypeError: this.status.error[i(...)(...)[0]] is undefined
  at errorMessage (/js/30.6745a1ac1fea881e0967.js:1:3973)
  at xr.prototype.get (/js/vendor.4906e353dde3502bedf6.js:6:40297)
  at xr.prototype.evaluate (/js/vendor.4906e353dde3502bedf6.js:6:41296)
  at Kt/< (/js/vendor.4906e353dde3502bedf6.js:6:14305)
  at showShake (/js/30.6745a1ac1fea881e0967.js:1:4032)
...
```

tiltec:
Transition to the new code went smoothly.

Closes #658 

I figured that in case of simple forms (just one or two fields), it makes sense to handle all errors (field-related, nonField, permission) in one place, called `anyError`. I wonder if it makes sense to have a mixin for this, on the other hand it's not super repetitive.

I added some more spinner buttons when submitting the form (unfortunately it changes the button size, might need a layout tweak in the future).

### DIFF
--- a/src/store/modules/auth.js
+++ b/src/store/modules/auth.js
@@ -41,6 +41,8 @@ export const getters = {
 }
 
 export const actions = {
+  clearLoginStatus: ({ dispatch }) => dispatch('meta/clear', ['login']),
+
   ...withMeta({
 
     async check ({ commit, dispatch }) {
--- a/src/store/modules/auth.js
+++ b/src/store/modules/auth.js
@@ -1,10 +1,14 @@
 import auth from '@/services/api/auth'
 import authUser from '@/services/api/authUser'
 import router from '@/router'
-import { onlyHandleAPIError } from '@/store/helpers'
 import { types as userTypes } from '@/store/modules/users'
+import { createMetaModule, withMeta, metaStatuses } from '@/store/helpers'
+
+export const modules = { meta: createMetaModule() }
 
 export const types = {
+  SET_USER: 'Set user',
+  CLEAR_USER: 'Clear user',
 
   SET_REDIRECT_TO: 'Set RedirectTo',
   CLEAR_REDIRECT_TO: 'Clear RedirectTo',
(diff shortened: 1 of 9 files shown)


src/store/modules/auth.spec.js
```diff
@@ -56,7 +56,7 @@ describe('auth', () => {
     mockLogin.mockImplementationOnce(throws({ response: { data: { foo: 'some error info' }, status: 403 } }))
```
nicksellen: You can use `createValidationError()` now.

src/store/modules/auth.spec.js
```diff
@@ -39,14 +39,14 @@ describe('auth', () => {
   it('can login', async () => {
     mockLogin.mockReturnValueOnce(user())
     await store.dispatch('auth/login')
-    expect(store.getters['auth/status'].error).toBeNull()
+    expect(store.getters['auth/loginStatus'].error).toBeUndefined()
```
nicksellen: Error is not a status field at all now.

src/store/modules/auth.js
```diff
@@ -53,147 +37,117 @@ export const getters = {
   userId: state => state.user && state.user.id,
   status: state => state.status,
   redirectTo: state => state.redirectTo,
+  ...metaStatuses(['login', 'update', 'changePassword', 'changeEmail']),
 }
 
 export const actions = {
+  clearLoginStatus: ({ dispatch }) => dispatch('meta/clear', ['login']),
```
nicksellen: Maybe a helper for these would be useful, I was triggering root actions `groups/meta/clear` but I think having the actions here would be better.

tiltec: Let's see how much we actually need these `clear` actions, I wouldn't add a helper just for the sake of it. When we have a wrapper like this, it should be easy to refactor the modules.

tiltec:
Can be merged!

PR #758, opened 2017-11-15
refactor profileedit

tiltec:
Use editMixin and allow enter-to-submit.

editMixin expects the `__unenriched` property on the value, so I added that for `auth/user`.

### DIFF
--- a/src/mixins/editMixin.js
+++ b/src/mixins/editMixin.js
@@ -12,7 +12,7 @@ export default {
     }
   },
   watch: {
-    'value' (current, previous) {
+    value (current, previous) {
       // we want to make sure it's _really_ changed or we risk undoing the users changes
       if (current !== previous || !deepEqual(current, previous)) {
         this.reset()
--- a/src/mixins/editMixin.js
+++ b/src/mixins/editMixin.js
@@ -7,27 +7,24 @@ export default {
     value: { required: true },
   },
   data () {
-    const source = this.value.id ? this.value.__unenriched : this.value
     return {
-      source,
-      edit: cloneDeep(source),
+      edit: cloneDeep(this.value),
     }
   },
   watch: {
-    'value.__unenriched' (current, previous) {
+    'value' (current, previous) {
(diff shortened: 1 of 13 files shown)


tiltec:
Oh no, the test looks really broken... should try again tomorrow...

tiltec:
I ended up removing the `__unenriched` field from the mixin, it's unneeded because the object diff does it's job well.

src/mixins/editMixin.js
```diff
@@ -7,27 +7,24 @@ export default {
...
-    const source = this.value.id ? this.value.__unenriched : this.value
     return {
-      source,
-      edit: cloneDeep(source),
+      edit: cloneDeep(this.value),
     }
   },
   watch: {
-    'value.__unenriched' (current, previous) {
+    'value' (current, previous) {
```
nicksellen: Doesn't need to be a string here any more.

src/mixins/editMixin.js
```diff
@@ -7,27 +7,24 @@ export default {
...
       // we want to make sure it's _really_ changed or we risk undoing the users changes
       if (current !== previous || !deepEqual(current, previous)) {
```
nicksellen: We have to be careful here as enriched fields might change based on other data, and if it changed then it would cause the form to be reset and overwrite the users changes.

The reason to have the watching in the first place is to get the updated data after saving to the server (other fields could have changed (or an id added in the case of creation)).

One solution would be to store a changes object, so we can reset from `value`, then reapply any changes (where there should be none in the case of updating due to save, and some in the case of update due to enriched field change).

Although I would be ok if we don't address this now, but to make a note of it somewhere (docs, or an issue).

tiltec: You're right, this should be addressed. As I see it, this would mean a change to `editMixin` only, so I'd do it in a later PR.

What would be a case where this behavior would get visible to the user? I can't think of any right now, it should only trigger during page load and after submit.

tiltec: See #771

Issue #741, opened 2017-11-13
Add a central location for open pick-up slots


djahnie:
Sylvia from Groß Gerau complained about the fact, that the new release doesn't show all open slots in pick-ups in a central place anymore. Right now we only show _completely_ empty pick-ups (orange button) and those the currentl logged-in user has already signed up for (green button).

![image](https://user-images.githubusercontent.com/17573771/32725505-a21a70ba-c875-11e7-9665-1440c8822487.png)

How about adding a third (maybe blue..?) button for pick-ups which are not empty but still have open slots?

Original mail from Sylvia:

> Hallo Janina,
> heute ist das neue karrot Release 3 online gegangen.
> Wir von Essen für Alle (EfA) vermissen die Übersicht, bei der alle Betriebe im Kalender zu sehen waren.
> Wo könne wir das finden?
> Wir haben derzeit 11 registrierte Betriebe und 17 Abholer.
> Bei den neuen Release sehe ich aber zunächst nur die Betriebe in denen ich aktuell zur Abholung registriert bin.
> Jeder kann bei EfA in jedem Betrieb abholen, oft kommen einzelne Tage als Ersatzabholung für die Tafel hinzu.
> Es ist also wichtig, dass jeder alle offenen Slots in den Betrieben sieht, damit er schnell reagieren kann.
> Es ist sehr mühsam ggf. mehrfach am Tag alle Betriebe anzuklicken, um zu sehen wie viele Slots besetzt sind und durch wen.
> Liebe Grüße
> Sylvia

tiltec:
I understand that they are asking for the same feature that we had in the angular version, a list of all pickups of all stores in the group.
I think it could make sense to create such a page, but separate from the wall. It could be called "All pickups" and just be a list by date.

I'm unsure where to put it in the mobile navigation, the top navbar is already quite full. Maybe in the "more options" menu of the group? But that's quite hidden...

djahnie:
I think it's important to have it very prominent. And I actually liked to find the pick-up info directly on the main group page, even though I agree that wall and pick-up info are not the same. So maybe it makes sense to seperate them.

How about making one 'pick-ups' page, which displays all the categories of interesting pick-ups and the navbar on mobile shows (from left to right): wall, pick-ups, description - and history moves under the three dots..?

tiltec:
[...]

PR #759, opened 2017-11-16
Add group pickuplist

djahnie:
As requested in #741 

@tiltec and me created a new page, which displayes an overview over _all_ upcoming pickups.

image

For mobile there is a new button in the bottom bar (because we couldn't think of anything better for now...)

screen shot 2017-11-16 at 19 20 38


djahnie:
Oops, forgot the newly created - and most important - file in the initial commits... 😅

PR #767, opened 2017-11-17
Add locales to intl polyfill

### LINKED ISSUE #748: Internationalization API is not available on some mobile browsers

nicksellen:
I ditched the polyfill service and went with webpack dynamic imports.

I also did a somewhat unrelated change to statically name all the date-fns locale imports (it reduces the number of webpack js files from ~190 to ~60, because if you use a dynamic import name it just grabs everything from the directory). Using dynamic import name with intl results in >900 output files :)

We can also stick any other custom polyfills in that file if the need arises. There seem to be some outstanding issues around that in sentry.

Fixes #748


src/polyfill.js
```diff
@@ -0,0 +1,35 @@
...
+  data () {
+    return {
+      locale: 'en',
+    }
+  },
+  watch: {
+    locale (locale) {
+      if (!window.IntlPolyfill) return
+      switch (locale) {
+        case 'de': return import('intl/locale-data/jsonp/de')
```
tiltec: I would prefer to have it so it accepts all languages. Keeps the cognitive overhead low when adding another locale. 

`import ('intl/locale-data/jsonp/${locale}')`

Anecdote: I needed to update 3 locations in the old angular codebase when adding another language and I routinely kept forgetting at least one of them. Made it quite annoying. As this happens quite rarely, it's hard to get a routine there.

nicksellen: There are 719 locales available for intl, and including them all (plus all the date-fns) ones means we have close to 1000 built files in webpack. This slows things down (although I did not measure it).

Another idea I had (because I agree it's annoying and easy to forget something if you have to edit 3 things) is to have a file per locale which would load everything needed for it (karrot, date-fns, intl (if needed)).

Would having to add one file with 3 things in it lower the cognitive overhead sufficiently?

nicksellen: >  I routinely kept forgetting at least one of them
> As this happens quite rarely

We can also just add a test to make sure everything can be defined.

tiltec: > Would having to add one file with 3 things in it lower the cognitive overhead sufficiently?

> We can also just add a test to make sure everything can be defined.

Both of them seem good solutions, will make an issue out of it.

PR #773, opened 2017-11-17
Add redirects and tests

### LINKED ISSUE #638: Ensure redirects are handled correctly

### LINKED ISSUE #762: User page is empty if user hasn't joined any group
(more info needed?)

tiltec:
Closes #638 
Closes #762

I added some unit tests. I also addressed #762 because it was related: to work around race conditions when loading the page, I just try to load the entry directly from the server if it's not in vuex yet.

To do:
- [x] fix invalid group id redirect

### DIFF
--- a/src/store/modules/groups.js
+++ b/src/store/modules/groups.js
@@ -96,7 +96,7 @@ export const actions = {
       commit(types.SET_ACTIVE, { groupId })
 
       await dispatch('fetchGroup', groupId)
-      const hasError = Object.keys(getters['meta/status']('fetchGroup', groupId).validationErrors).length > 0
+      const hasError = getters['meta/status']('fetchGroup', groupId).hasValidationErrors
       if (hasError) {
         const groupExists = !!getters.get(groupId)
         const error = { translation: groupExists ? 'GROUP.NONMEMBER_REDIRECT' : 'NOT_FOUND.EXPLANATION' }
--- a/src/store/modules/groups.js
+++ b/src/store/modules/groups.js
@@ -95,10 +95,12 @@ export const actions = {
 
       commit(types.SET_ACTIVE, { groupId })
 
-      try {
-        await dispatch('fetchGroup', groupId)
-      }
-      catch (error) {
+      await dispatch('fetchGroup', groupId)
+      const hasError = Object.keys(getters['meta/status']('fetchGroup', groupId).validationErrors).length > 0
+      if (hasError) {
+        const groupExists = !!getters.get(groupId)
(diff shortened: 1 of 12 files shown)


src/store/modules/groups.spec.js
```diff
@@ -116,7 +122,7 @@ describe('groups', () => {
     })
 
     it('can not join a group', async () => {
-      mockJoin.mockImplementationOnce(() => { throw new Error('some error') })
+      mockJoin.mockImplementationOnce(throws(Error('some error')))
```
nicksellen: Ah, I didn't now `Error()` is the same as `new Error()`, looks neater, would be nice to move towards consistency. No change requested.

src/store/modules/groups.js
```diff
@@ -95,10 +95,12 @@ export const actions = {
 
       commit(types.SET_ACTIVE, { groupId })
 
-      try {
-        await dispatch('fetchGroup', groupId)
-      }
-      catch (error) {
+      await dispatch('fetchGroup', groupId)
+      const hasError = Object.keys(getters['meta/status']('fetchGroup', groupId).validationErrors).length > 0
```
nicksellen: I have seen this elsewhere, perhaps we need an extra property return from the getter function, I would called it `hasValidationErrors` (to distinguish it from other errors).

src/store/modules/users.js
```diff
@@ -248,5 +259,8 @@ export const mutations = {
   },
   [types.RECEIVE_USER] (state, { user }) {
     Vue.set(state.entries, user.id, user)
+    if (!state.idList.includes(user.id)) {
```
nicksellen: Do we actually need/use the `idList` for users? I left it off when I added the agreements module as we only ever need to get by id.

tiltec: Maybe for the users list? But they might need to get sorted on-the-fly anyways. I'd leave it in for now.. we could open an issue in the future.

src/store/modules/groups.spec.js
tiltec: It does work, but maybe it's not the best style? I remember reading that one should always use `new Error`.

src/store/modules/groups.js
tiltec: Though about it, will add it!

Issue #764, opened 2017-11-16
Don't allow past dates in pickupEdit


tiltec:
The Date input shouldn't allow anything that is before today.

![image](https://user-images.githubusercontent.com/4410802/32914729-e178b854-cb16-11e7-97c3-5e733be9079c.png)

PR #769, opened 2017-11-17
Small improvements no. 3

D0nPiano:
Fixes #765 #764 #730

PR #774, opened 2017-11-17
Issue #764: Limit min date for one-time pickups to today

timnovis:
Fixes issue #764

### DIFF
--- a/src/components/Pickups/PickupEdit.vue
+++ b/src/components/Pickups/PickupEdit.vue
@@ -21,7 +21,7 @@
         :error="hasError('date')"
         :error-label="firstError('date')"
         >
-        <q-datetime type="date" v-model="edit.date" :min="this.today" :display-value="$d(edit.date, 'dateShort')"/>
+        <q-datetime type="date" v-model="edit.date" :min="today" :display-value="$d(edit.date, 'dateShort')"/>
       </q-field>
 
       <q-field
--- a/src/components/Pickups/PickupEdit.vue
+++ b/src/components/Pickups/PickupEdit.vue
@@ -21,7 +21,7 @@
         :error="hasError('date')"
         :error-label="firstError('date')"
         >
-        <q-datetime type="date" v-model="edit.date" :display-value="$d(edit.date, 'dateShort')"/>
+        <q-datetime type="date" v-model="edit.date" :min="this.today" :display-value="$d(edit.date, 'dateShort')"/>
       </q-field>
 
       <q-field
--- a/src/components/Pickups/PickupEdit.vue
+++ b/src/components/Pickups/PickupEdit.vue
@@ -21,7 +21,7 @@
         :error="hasError('date')"
         :error-label="firstError('date')"
         >
-        <q-datetime type="date" v-model="edit.date" :min="today" :display-value="$d(edit.date, 'dateShort')"/>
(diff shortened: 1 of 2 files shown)


src/components/Pickups/PickupEdit.vue
```diff
@@ -21,7 +21,7 @@
         :error="hasError('date')"
         :error-label="firstError('date')"
         >
-        <q-datetime type="date" v-model="edit.date" :display-value="$d(edit.date, 'dateShort')"/>
+        <q-datetime type="date" v-model="edit.date" :min="this.today" :display-value="$d(edit.date, 'dateShort')"/>
```
nicksellen: You can leave out the `this.` in templates

src/mixins/editMixin.js
```diff
@@ -26,14 +26,22 @@ export default {
     hasChanged () {
       return !this.isNew && !deepEqual(this.value, this.edit)
     },
+    today () {
```
nicksellen: Builtin classes (e.g. `Date`) are not reactive, so this will never update.

You could either just initialize it in the `data()` function (it doesn't really _need_ to be reactive).

Or, if you want a reactive value, you can get it using:

```js
import dateFnsHelper from '@/services/dateFnsHelper'

// ... stuff

computed: {
  now () {
    return dateFnsHelper.now // <--- this value is reactive!
  }
}
```

src/mixins/editMixin.js
```diff
@@ -26,14 +26,22 @@ export default {
...
       if (this.isNew) {
         this.$emit('save', this.edit, event)
       }
       else {
-        this.$emit('save', { ...objectDiff(this.value, this.edit), id: this.value.id }, event)
+        this.$emit(
+          'save',
+          { ...objectDiff(this.value, this.edit), id: this.value.id },
+          event,
+        )
```
nicksellen: Could you revert this unrelated formatting change?

src/components/Pickups/PickupEdit.vue
timnovis: Oops! I knew that. Sorry.

src/mixins/editMixin.js
timnovis: Fix pushed

src/mixins/editMixin.js
timnovis: Fix pushed

src/components/Pickups/PickupEdit.vue
timnovis: Fix pushed

src/mixins/editMixin.js
nicksellen: Ah, I just noticed this is in the `editMixin`. It would be more appropriate in the `PickupEdit` component as the past date only relates to that, not editing in general.

Also, when you move it, maybe rename the computed property to `now` so it's consistent with the name inside the helper. Thanks!

tiltec:
Thanks for your initiative! However, did you notice that @D0nPiano already assigned himself to the issue and proposed a fix as part of #769?

nicksellen:
Yeah, but this one is :sparkles: _responsive_ :sparkles:

timnovis:
Updated!

tiltec:
True, I'd always favor a reactive solution :)

PR #775, opened 2017-11-18
Apply outside changes to edited values instead of replacing them

### LINKED ISSUE #771: editMixin: keep user's modifications when enriched field changes
Nick commented:

>We have to be careful here as enriched fields might change based on other data, and if it changed then it would cause the form to be reset and overwrite the users changes.
>
>The reason to have the watching in the first place is to get the updated data after saving to the server (other fields could have changed (or an id added in the case of creation)).
>
>One solution would be to store a changes object, so we can reset from value, then reapply any changes (where there should be none in the case of updating due to save, and some in the case of update due to enriched field change).

Need to investigate how to trigger the problem and how to solve it in a nice way.

tiltec:
Closes #771
Prevents flickering of values while sending. Somehow it also fixes the spinner on q-button after sending.

### DIFF
--- a/src/mixins/editMixin.js
+++ b/src/mixins/editMixin.js
@@ -13,10 +13,8 @@ export default {
   },
   watch: {
     value (current, previous) {
-      // we want to make sure it's _really_ changed or we risk undoing the users changes
-      if (current !== previous || !deepEqual(current, previous)) {
-        this.reset()
-      }
+      const changes = objectDiff(previous, current)
+      this.edit = cloneDeep({ ...this.edit, ...changes })
     },
   },
   computed: {


src/mixins/editMixin.js
```diff
@@ -13,10 +13,8 @@ export default {
   },
   watch: {
     value (current, previous) {
-      // we want to make sure it's _really_ changed or we risk undoing the users changes
-      if (current !== previous || !deepEqual(current, previous)) {
-        this.reset()
-      }
+      const changes = objectDiff(previous, current)
+      this.edit = cloneDeep({ ...this.edit, ...changes })
```
nicksellen: Should be able to leave out the `cloneDeep` here? As the `{}` creates a new object, and I don't think we have any interesting nested data? (maybe it's not needed in the other places either...).

I guess we should get this under test coverage so can answer these questions better...

Change if you want, or keep it like this :)

tiltec: I'd leave it for now. By the way, I'm quite sure that the objectDiff doesn't work if a value gets added that is not on the top level ob the `edit` object, e.g. hypothetical `edit.location.latitude`

PR #780, opened 2017-11-19
Add android build step

nicksellen:
Almost there! But, some odd errors that don't happen when I connect via ssh, or run locally (in same docker image). Hmmm.


nicksellen:
Ok, it seemed to fix itself, so all good!

Doesn't build production versions yet (only versions that point to dev.karrot.world), but that can come next.

PR #781, opened 2017-11-19
Improve code style of GroupEdit, StoreEdit and AgreementForm (editMixin, statusMixin)

tiltec:
Related issue #668 

I ran into problems while handling default values in forms, previously they either got passed in from a parent component (that's what editMixin was made for) or they got defined in the `default` field of a prop. I think the former is nicer, however the latter works together with vuex-connect nicely.

I tried to use `methodsToProps` and pass the method as `value` prop, but that breaks the change-detection `editMixin` and I didn't manage to fix it yet.


tiltec:
I see three options here:

- get rid of vuex-connect -> makes wrapper components more verbose
- get rid of editMixin -> makes edit forms more verbose
- magically make it work -> could make editMixin more complicated

I'd favor the third options, but I need some more insights.

tiltec:
I decided to override the definition of the `value` prop in the edit form to store the defaults, seems most straightforward to me.
If we would want to keep the defaults out of the component, we could use the `stateToProps` hook of vuex-connect to pass in the defaults object.

src/components/Group/AgreementForm.vue
```diff
@@ -1,14 +1,14 @@
 <template>
   <div class="edit" :class="{ changed: hasChanged }">
-    <form @submit.prevent="save">
+    <form @submit.prevent="checkSave">
```
tiltec: could get a better name instead of `checkSave`...

tiltec:
Ready for review!

src/components/Group/AgreementForm.vue
nicksellen: `maybeSave`?

PR #782, opened 2017-11-20
switch to new API

tiltec:
Comes with https://github.com/yunity/karrot-backend/pull/446

### DIFF
--- a/src/services/api/authUser.js
+++ b/src/services/api/authUser.js
@@ -16,4 +16,12 @@ export default {
   delete (id) {
     return axios.delete(`/api/auth/user/`)
   },
+
+  verifyMail (key) {
+    return axios.post('/api/auth/user/verify_mail/', { key })
+  },
+
+  resendVerificationRequest () {
+    return axios.post('/api/auth/user/resend_verification/')
+  },
 }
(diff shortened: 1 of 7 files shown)


src/services/api/authUser.js
```diff
@@ -16,4 +16,12 @@ export default {
...
     return axios.delete(`/api/auth/user/`)
   },
+
+  verifyMail (key) {
+    return axios.post('/api/auth/user/verify_mail/', { key })
+  },
+
+  resendVerificationRequest () {
+    return axios.post('/api/auth/user/resend_verification/')
+  },
```
nicksellen: I would make these use async/await, without returning anything, otherwise some API methods return the data, yet these ones the full response object.

The use of them does not use the return data, so it should not provide it.

src/store/modules/verifymail.js
```diff
@@ -28,7 +28,7 @@ export const actions = {
   async verify ({ commit }, key) {
     commit(types.REQUEST)
```
nicksellen: Could migrate to use meta status?

tiltec: Definitely, but I'd make it a separate PR 🚫

src/services/api/authUser.js
tiltec: I agree, but I didn't want to make any unrelated changes except for moving to the new API. I'll write it onto my list of things to do.

tiltec:
The endpoint refactoring means we should release a new dev app version too, otherwise reset password is broken.

PR #783, opened 2017-11-21
Add hooks to trigger actions on route enter and leave

tiltec:
My first try at #726, quite basic for now.

This is the interface:

```javascript
meta: {
      beforeEnter: 'groups/selectGroupInfo', // triggers vuex action, passing route params as argument
      afterLeave: 'groups/clearGroupInfo', // triggers vuex action
}
```

I think it fulfills most of our requirements, I think could get away without functions in beforeEnter etc for quite a while. Eventually I think they might become necessary though.
I like the implicit way of passing in the route params into the action, although it assumes that all params are integers, which might change in the mid-term future. Wouldn't worry about that now.

---

- [x] remove console.log
- [ ] see if historyList can also get refactored
- [x] add tests
- [x] add docs
- [x] fix failing tests
- [x] test all changed bits by clicking around

### DIFF
--- a/src/routes/main.js
+++ b/src/routes/main.js
@@ -48,8 +48,7 @@ export default [
         { type: 'activeGroupInfo' },
       ],
       beforeEnter: 'groups/selectGroupInfo',
-      onError: 'group id not found',
-      onLeave: 'groups/clearGroupInfo',
+      afterLeave: 'groups/clearGroupInfo',
     },
     components: {
       default: GroupInfo,
--- a/src/routes/main.js
+++ b/src/routes/main.js
@@ -89,6 +89,7 @@ export default [
       breadcrumbs: [
         { type: 'activeGroup' },
       ],
+      beforeEnter: 'groups/selectGroup',
     },
     components: {
       default: GroupLayout,
--- a/src/routes/main.js
+++ b/src/routes/main.js
@@ -47,6 +47,9 @@ export default [
         { translation: 'JOINGROUP.ALL_GROUPS', route: { name: 'groupsGallery' } },
         { type: 'activeGroupInfo' },
       ],
+      beforeEnter: 'groups/selectGroupInfo',
+      onError: 'group id not found',
+      onLeave: 'groups/clearGroupInfo',
     },
     components: {
       default: GroupInfo,
--- a/src/routes/main.js
+++ b/src/routes/main.js
@@ -63,6 +63,7 @@ export default [
         { translation: 'JOINGROUP.ALL_GROUPS', route: { name: 'groupsGallery' } },
         { translation: 'GROUP.CREATE_TITLE', route: { name: 'groupCreate' } },
       ],
+      beforeEnter: 'groups/fetchTimezones',
     },
     components: {
       default: GroupCreate,
@@ -165,6 +166,7 @@ export default [
           breadcrumbs: [
             { translation: 'GROUP.INVITE_TITLE', route: { name: 'groupInvitations' } },
           ],
+          beforeEnter: 'invitations/fetch',
         },
         components: {
           default: GroupInvitations,
@@ -178,6 +180,7 @@ export default [
           breadcrumbs: [
             { translation: 'GROUP.EDIT', route: { name: 'groupEdit' } },
           ],
+          beforeEnter: 'groups/fetchTimezones',
         },
         components: {
           default: GroupEdit,
--- a/src/routes/main.js
+++ b/src/routes/main.js
@@ -330,6 +330,8 @@ export default [
       breadcrumbs: [
         { type: 'activeUser' },
       ],
+      beforeEnter: 'users/selectUser',
+      afterLeave: 'users/clearSelectedUser',
     },
     components: {
       default: User,
--- a/src/routes/main.js
+++ b/src/routes/main.js
@@ -262,6 +262,8 @@ export default [
               breadcrumbs: [
                 { translation: 'PICKUPMANAGE.TITLE', route: { name: 'storePickupsManage' } },
               ],
+              beforeEnter: 'pickupSeries/fetchListForActiveStore',
+              afterLeave: 'pickupSeries/clearList',
             },
           },
           {
--- a/src/routes/main.js
+++ b/src/routes/main.js
@@ -228,6 +228,8 @@ export default [
           breadcrumbs: [
             { type: 'activeStore' },
           ],
+          beforeEnter: 'stores/selectStore',
+          afterLeave: 'stores/clearSelectedStore',
         },
         props: {
           sidenav: true,
--- a/src/routes/main.js
+++ b/src/routes/main.js
@@ -79,6 +79,8 @@ export default [
       breadcrumbs: [
         { translation: 'HISTORY.DETAILS', route: { name: 'historyDetail' } },
       ],
+      beforeEnter: 'history/setActive',
+      afterLeave: 'history/clearActive',
     },
   },
   {
--- a/src/store/modules/groups.js
+++ b/src/store/modules/groups.js
@@ -90,7 +90,7 @@ export const actions = {
 
   ...withMeta({
 
-    async selectGroup ({ commit, state, dispatch, getters, rootState }, groupId) {
+    async selectGroup ({ commit, state, dispatch, getters, rootState }, { groupId }) {
       if (state.activeGroupId === groupId) return
 
       commit(types.SET_ACTIVE, { groupId })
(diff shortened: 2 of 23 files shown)


tiltec:
Before I forget it: it's fully reactive, the `onLeave` action gets also triggered when you replace something in the URL and hit "enter". No more state watching...

src/routes/main.js
```diff
@@ -47,6 +47,9 @@ export default [
         { translation: 'JOINGROUP.ALL_GROUPS', route: { name: 'groupsGallery' } },
         { type: 'activeGroupInfo' },
       ],
+      beforeEnter: 'groups/selectGroupInfo',
+      onError: 'group id not found',
```
nicksellen: There could be other reasons for an error... do we not get a reasonable message from the error itself?

src/routes/main.js
```diff
@@ -47,6 +47,9 @@ export default [
...
+      onLeave: 'groups/clearGroupInfo',
```
nicksellen: `beforeLeave` (or maybe it has to be `afterLeave`) would make it:
a) consistent with vue router lifecycle names
b) clearer what it does * (it can never do it exactly ON route, only just before, or just after)

\* if combined with my suggestion below :)

src/store/plugins/router.js
```diff
@@ -50,15 +50,32 @@ export default store => {
...
+      if (m.meta.beforeEnter) {
+        try {
+          console.log('beforeEnter', to.params.groupInfoId)
+          // TODO parse all params, assuming that all are integer
+          const params = { groupInfoId: parseInt(to.params.groupInfoId, 10) }
+          store.dispatch(m.meta.beforeEnter, params)
+        }
+        catch (error) {
+          console.log('onError', m.meta.onError)
+          store.dispatch('routeError/set', { translation: m.meta.onError })
```
nicksellen: Maybe find a way here to catch route errors specifically, and rethrow others. You can make custom error classes :) Or just add certain data to a normal error. Then get the translation key from the thrown error.

src/store/plugins/router.js
```diff
@@ -50,15 +50,32 @@ export default store => {
   })
 
   router.beforeEach(async (to, from, next) => {
+    for (let m of from.matched) {
+      if (m.meta.onLeave) {
+        console.log('onLeave', m.meta.onLeave)
+        store.dispatch(m.meta.onLeave)
```
nicksellen: Maybe it makes a difference running it an `afterEach` handler? Would make the sequence of things clearer to me. No specific reason, but my senses tell me the ordering could confuse things, or maybe a specific reason is clearing out the old data just that bit earlier.

Also, I wonder if they should await the result of the action.

tiltec:
Thanks for your input! How about if we could get rid of the `onError` hook for most cases and instead `throw new RouteError({ translation: 'asdfa' })` in an action to pass data into the `routeError/set` action?

nicksellen:
> Thanks for your input! How about if we could get rid of the onError hook for most cases and instead throw new RouteError({ translation: 'asdfa' }) in an action to pass data into the routeError/set action?

Yes, sounds good to me!

src/routes/main.js
tiltec: `beforeLeave` seems good, the question is if the action is allowed to abort the transition? I mostly intended it to be a clear-up task, so `afterLeave` should be fine as well.

src/store/plugins/router.js
tiltec: `afterEach` runs after the transition already started (is confirmed), I think. So it actually runs after `beforeEach` and not when you leave the route.

tiltec:
That should also allow us to use the meta module thing together with the beforeEnter hook: as the meta handler will catch any `404` responses, we need to throw a fresh error anyways.

tiltec:
- maybe support multiple actions - seems not needed right now though
- what about optional stuff, like fetchAbout? It shouldn't trigger an routeError when it fails to resolve
- if beforeEnter needs additional data from a parent beforeEnter, it can go via store. then the route params remain unused

- [ ] remove console.log
- [ ] see if historyList can also get refactored
- [ ] add tests

nicksellen:
> if beforeEnter needs additional data from a parent beforeEnter, it can go via store. then the route params remain unused

I didn't understand this. What can go via store?

nicksellen:
> what about optional stuff, like fetchAbout? It shouldn't trigger an routeError when it fails to resolve

Maybe it's _just_ fetch about. This would seem reasonable to never error, just to return an object that says the about status locally. I think we went through that as an option before.

tiltec:
>> if beforeEnter needs additional data from a parent beforeEnter, it can go via store. then the route params remain unused
>
> I didn't understand this. What can go via store?

If you have two routes `group` and `group/pickups` (pickups is a child of group), then `pickups` has a `beforeEnter: 'getPickupsForGroup'`, which relies on the `currentGroup` being set by the parent `beforeEnter`. That means that `getPickupsForGroup` won't use the router `params`, but the data inside vuex to determine which group it should load the pickups for.

Is this good design (it's basically how it currently is) or should `beforeEnter` actions mostly act on the route `params`?

tiltec:
I added some tests - took me a while to figure out that testing the whole router plugin is too hard, so I test the function in isolation. Maybe that's more of a unit :)
Now only need to fix the other failing tests...

tiltec:
Ready for final review :)

src/store/modules/stores.spec.js
```diff
@@ -39,10 +32,15 @@ describe('stores module', () => {
...
+    expect.assertions(2)
     mockGet.mockImplementationOnce(throws(createValidationError({ detail: 'Not found' })))
-    store.dispatch('stores/selectStore', 9999)
-    expect(routeError.actions.set).toBeCalled()
+    try {
+      await store.dispatch('stores/selectStore', { storeId: 9999 })
+    }
+    catch (e) {
+      expect(e.type).toEqual('RouteError')
+    }
```
nicksellen: This does not error on the case where the error is not thrown, you can do it with this:

```js
await expect(store.dispatch('stores/selectStore', { storeId: 9999 })).rejects.toHaveProperty('type, 'RouteError')
```

src/store/modules/users.spec.js
```diff
@@ -62,9 +55,23 @@ describe('users', () => {
...
+    expect.assertions(1)
     mockGet.mockImplementationOnce(throws(createValidationError({ detail: 'Not found' })))
-    store.dispatch('users/selectUser', 9999)
-    expect(routeError.actions.set).toBeCalled()
+    try {
+      await store.dispatch('users/selectUser', { userId: 9999 })
+    }
+    catch (e) {
+      expect(e.type).toEqual('RouteError')
+    }
```
nicksellen: Same here...

src/store/modules/users.spec.js
```diff
@@ -62,9 +55,23 @@ describe('users', () => {
     expect(store.getters['users/byActiveGroup'].map(e => e.id)).toEqual([user1.id, user2.id])
   })
 
-  it('sets routeError if user is not accessible', () => {
+  it('can select user', async () => {
+    expect.assertions(1)
+    mockGet.mockReturnValueOnce(user1)
+    await store.dispatch('users/selectUser', { userId: user1.id })
+    expect(store.getters['users/activeUser'].id).toEqual(user1.id)
+    mockGet.mockReset()
```
nicksellen: Why reset them?

If it's needed would be better to reset them all in a before/after hook using [jest.resetAllMocks() ](http://facebook.github.io/jest/docs/en/jest-object.html#jestresetallmocks)

src/store/modules/groups.js
```diff
@@ -90,7 +90,7 @@ export const actions = {
 
   ...withMeta({
 
-    async selectGroup ({ commit, state, dispatch, getters, rootState }, groupId) {
+    async selectGroup ({ commit, state, dispatch, getters, rootState }, { groupId }) {
```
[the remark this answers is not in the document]
tiltec: Oh, I didn't think of that! But thinking about it further, does it make sense if a beforeEnter action is inside withMeta anyways? It needs its own error handling anyways: if the error doesn't matter, ignore it right away, and if it _does_ matter, throw `RouteError`.

src/store/modules/users.spec.js
tiltec: It didn't work without resetting them, even if `mockReturnValueOnce` should just mock it a single time. Maybe I'll apply the resetAllMocks thing in future...

tiltec:
Ok, another round of editing.. and them I'm done 🗡

src/store/modules/groups.spec.js
```diff
@@ -226,39 +220,35 @@ describe('groups', () => {
...
       store = createStore({
         groups: require('./groups'),
-        routeError,
       })
     })
 
-    it('sets routeError if group does not exist', async () => {
-      await store.dispatch('groups/selectGroupInfo', 9999)
-      expect(routeError.actions.set).toBeCalled()
+    it.only('throws routeError if group does not exist', async () => {
```
nicksellen: You left in a `.only` :)

tiltec:
I feel bad! drive-by fixed in #790

PR #790, opened 2017-11-24
Refactor conversations

tiltec:
- rewrite conversations module in fcm-style and withMeta
- split Conversations components from Wall
- rename WallInput to ConversationCompose
- rename WallMessage to ConversationMessage

Should work, although I didn't do any testing of the error handling.


tiltec:
This `CMessage` looks weird to me now, maybe I rename to `ConversationMessage`?

nicksellen:
> This CMessage looks weird to me now, maybe I rename to ConversationMessage?

Yes, and fits the style guide point that you should name components that are only relevant within another one as a prefix e.g. `Conversation` and `ConversationMessage`

src/store/modules/conversations.js
```diff
@@ -1,243 +1,119 @@
...
+      return {
+        messages: getters.activeMessages,
+        canLoadMore,
+        ...metaStatusesWithId(getters, ['send', 'fetch', 'fetchMore'], id),
+      }
+    },
   },
+  actions: {
+    ...withMeta({
+      async send ({ commit, state, dispatch }, conversationId, messageData) {
```
nicksellen: Oh, I didn't know you pass two args? Is it supported?

src/store/modules/conversations.js
```diff
@@ -1,243 +1,119 @@
...
-    }
-  },
-  [types.RECEIVE_MESSAGES_ERROR] (state, { conversationId, error }) {
-    console.log(error)
-    if (state.activeConversationId === conversationId) {
-      state.receiveStatus = {
-        isWaiting: false,
-        error,
-        success: false,
+      if (!getters.activeMessages.find(e => e.id === message.id)) {
```
nicksellen: A bit inefficient to go through all the messages each time, ok for now I'm sure but could be much improved.

src/store/modules/conversations.js
```diff
@@ -1,243 +1,119 @@
...
+    setMessages (state, { conversationId, messages, cursor }) {
+      Vue.set(state.messages, conversationId, messages)
       Vue.set(state.cursors, conversationId, cursor)
-    }
-    if (state.activeConversationId === conversationId) {
-      state.receiveMoreStatus = {
-        isWaiting: false,
-        error: null,
+    },
+    appendMessages (state, { conversationId, messages, cursor }) {
```
nicksellen: Not sure I quite understand why we would prepend and append messages.

src/components/Conversation/CMessage.vue
```diff
@@ -25,7 +25,7 @@ import { QItem, QItemSide, QItemMain, QItemTile } from 'quasar'
 import DateAsWords from '@/components/General/DateAsWords'
 
 export default {
-  name: 'WallMessage',
+  name: 'CMessage',
```
nicksellen: `ConversationMessage` fits the styleguide rule better.

src/components/Conversation/Compose.vue
```diff
@@ -22,12 +22,13 @@
 <script>
 import ProfilePicture from '@/components/ProfilePictures/ProfilePicture'
 import { QItem, QItemMain, QInput, QBtn, QItemSide, QItemTile } from 'quasar'
+import statusMixin from '@/mixins/statusMixin'
 
 export default {
-  name: 'WallInput',
+  name: 'Compose',
```
nicksellen: `ConversationCompose` might fit better, or could this also be a more standalone component?

src/components/Conversation/CMessage.vue
tiltec: done

src/components/Conversation/Compose.vue
tiltec: done

src/store/modules/conversations.js
tiltec: I thought it works, then I tried again, it fails. Will pass an object instead.

src/store/modules/conversations.js
tiltec: I agree, we could map the messages by ID, store in an object and then do a simple key lookup. But I'm sure that this is not the reason why the page is sluggish sometimes ;)

src/store/modules/conversations.js
tiltec: prepend: when the user writes a message and it should go to top
append: when loading more message history from the server

will add comments :)

src/store/modules/conversations.js
nicksellen: I was thinking about an  "always sorted list" approach (would also fit with de-duplication)

Issue #760, opened 2017-11-16
Follow vuejs styleguide more closely


nicksellen:
https://vuejs.org/v2/style-guide/ has a lot of detailed stuff, also looks like the estlint vue plugin will be able to check for these things too when it's ready (#680).

tiltec:
Thanks! I'm reading it and will follow it more. Anything that seems especially useful for now (before the linter works?)

nicksellen:
these seem most useful/implementable:

from essential:
* Prop definitions should be as detailed as possible.
* Always use key with v-for.
* Component names should always be multi-word, except for root App components. (this seems a bit overly worried, but still if it's in _essential_ would be nice to follow it)

from strongly recommended:
* Component name casing in templates 
* Multi-attribute elements (although maybe not as strictly as they say)

from potentially dangerous patterns:
* v-if/v-if-else/v-else without key might need understanding more, I didn't on first read

tiltec:
Regarding our discussion, I also noticed https://vuejs.org/v2/style-guide/#Multi-attribute-elements-strongly-recommended

```html
<img
  src="https://vuejs.org/images/logo.png"
  alt="Vue Logo"
>
```

The closing `>` on the newline is not explicitly mentioned in the text, so it might be a personal preference of the style guide author, or a no-brainer (given the similarity with the selfclosing tag).

tiltec:
Only one left: the `vue/require-prop-types` rule ✏️

PR #793, opened 2017-11-26
Add eslint vue plugin

### LINKED ISSUE #680: Add linting for vue templates

nicksellen:
Fixes #680, partially addresses #760

Turns out I had to disable the `html` plugin to get it working. The `essential` preset was easy to add. The `strongly-recommended` needs a lot of manual work that can't be `--fix`'d so I would keep to `essential` in this PR.


nicksellen:
Actually, if I switch to `strongly-recommended` then disable 3 of the rules, then we get ~300 more autofixes done.

PR #795, opened 2017-11-26
Refactor groups module

### LINKED ISSUE #794: Split up groups module in groupsInfo and currentGroup modules

tiltec:
Closes #794 

I wanted to do this for quite a while already, so I went through it quite thoroughly.

- Split up groups module into three new ones, make them _fcm-style_
  - `groups` now has everything related to public information, plus join, leave, edit, create
  - `currentGroup` has all the bits relevant to the currently selected group
  - `timezones` handles the timezones needed for group editing
- Rename all occurences of _active_ group to _current_
- Clarify GroupsGallery and GroupsInfo components
  - all components that belong to the gallery are now prefixed with `GroupsGallery`
  - the group info page is now called `GroupsPreview`
- Reduce number of attributes per line in some templates
- Add some types for props
- Return `undefined` from getter when group or currentGroup does not exist

I made some unrelated changes to the search code as well, to improve code style and reduce the delay.


src/components/GroupJoin/GroupsGallery.story.js
```diff
@@ -1,7 +1,7 @@
 import { storiesOf } from '@storybook/vue'
 import { action } from '@storybook/addon-actions'
 
-import GroupPreviewGallery from './GroupPreviewGallery'
+import GroupsGallery from './GroupsGalleryUI'
```
nicksellen: I would call it `GroupGallery` as the plural is sort of built into the `Gallery` bit (wouldn't be a gallery with just one... like `GroupList` instead of `GroupsList`), just slightly jars for me as it is now, doesn't flow when I say it very nicely. Not a strong feeling though.

src/pages/Group/Description.vue
```diff
@@ -18,7 +18,7 @@ export default {
   components: { QCard, Markdown },
   computed: {
     ...mapGetters({
-      group: 'groups/activeGroup',
+      group: 'currentGroup/get',
```
nicksellen: `get` sounds like it would be a function-getter, but it's presumably not? maybe `group` is a better name? (even though it's kind of duplicated...)

src/store/modules/groups.js
```diff
@@ -2,246 +2,133 @@ import Vue from 'vue'
...
+    activeUserGroups: (state, getters, rootState, rootGetters) => {
+      let activeUser = rootGetters['users/activeUser']
+      return activeUser ? getters.all.filter(el => el.members.includes(activeUser.id)) : []
     },
-
-    async create ({ commit, dispatch }, group) {
-      const createdGroup = await groups.create(group)
-      commit(types.RECEIVE_GROUP, { group: createdGroup })
-      router.push({ name: 'group', params: { groupId: createdGroup.id } })
+    my: (state, getters) => getters.all.filter(e => e.isMember).sort(sortByName),
```
nicksellen: `mine` sounds like it would be more correct grammar to me

src/components/GroupJoin/GroupsGallery.story.js
tiltec: Agree, GroupGallery could be enough to figure out the plurality… I just wanted to make extra sure that I don't confuse them with the single-group preview anymore. Will think about it more and then change it later.

src/pages/Group/Description.vue
tiltec: or something like `value`?

PR #796, opened 2017-11-26
Add default props to everything

nicksellen:
Only two more vue eslint plugin rules to comply with to reach `strongly-recommended` level...

I couldn't verify all the pages work fine, but I clicked randomly around the site (which got quite a few prop-related errors).


tiltec:
Looks good!
Sidenote: I wonder if we should keep the `required: false` modifiers because they throw so many errors on loading/unloading. Maybe it's better to have sane defaults instead?

tiltec:
(I meant `required: true`, edited it)

PR #801, opened 2017-12-17
add storyshots

tiltec:
Storyshots are snapshot tests derived from storybook stories: https://github.com/storybooks/storybook/tree/master/addons/storyshots
I adapted the code to our VueJS codebase and it works quite nicely so far!

Some considerations:
- there's only one snapshot file for all stories, it would be nicer to have one file per story
- it would be nicer to have some control over which tests should run, e.g. during debugging
- stories and component tests could share more code, inspiration is https://github.com/mthuret/storybook-addon-specifications
- a Math.random mock was necessary to snapshot AmountPicker reliably, it would be nice to reset it after usage to avoid side effects
- '@/' alias imports for images cause errors in jest, because `moduleNameMapper` does not replace them with the `fileMock` anymore. The `@` alias might be completely unneeded, as webpack already looks in `src/` for imports by default (was disabled in our storybook config before)
- require.context does not work in jest, so I used the `glob` package
- snapshots are not super nice, they contain lots of comments (possibly rendering errors!?)
- [x] fix timezone dependency that breaks down in CircleCI

### DIFF
--- a/.storybook/config.js
+++ b/.storybook/config.js
@@ -6,12 +6,13 @@ import '../src/themes/app.mat.styl'
 import Vue from 'vue'
 import Quasar from 'quasar'
 import Vuex from 'vuex'
-import VueI18n from 'vue-i18n'
+import MockRouterLink from '>/MockRouterLink'
 
 Vue.config.productionTip = false
 Vue.use(Quasar) // Install Quasar Framework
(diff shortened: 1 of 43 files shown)


tiltec:
Uii, increase of `+952` lines covered!

PR #804, opened 2017-12-21
Fix style issues

djahnie:
Related to #760.
I started with the multi-attribute elements.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -27,7 +27,7 @@ module.exports = {
   'rules': {
     // We relax some of the vue rules for now as they require a lot of manual changes
     // TODO: comply with the rules, then remove these lines
-    'vue/max-attributes-per-line': 0,
+    // 'vue/max-attributes-per-line': 0,
     'vue/require-prop-types': 0,
 
     // allow paren-less arrow functions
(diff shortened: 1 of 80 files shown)


tiltec:
My personal feeling is that allowing 3 attributes in the first line is OK. eslint-plugin-vue docs are not clear in this regard, I opened an issue about it: https://github.com/vuejs/eslint-plugin-vue/issues/300

In short: style guide says only 1 attribute allowed, eslint-plugin-vue author feels that 3 should be ok.

djahnie:
Some attributes are quite long though, so that three could already be too much... Anyways, since I saw your comment too late and like this hard 'one-attr-only' rule, that's how I did it. At least there's no errors anymore... ;)
But I'm not done, right? Which rule should/can I tackle next?

.eslintrc.js
```diff
@@ -27,7 +27,7 @@ module.exports = {
   'rules': {
     // We relax some of the vue rules for now as they require a lot of manual changes
     // TODO: comply with the rules, then remove these lines
-    'vue/max-attributes-per-line': 0,
+    // 'vue/max-attributes-per-line': 0,
```
tiltec: This line is not needed anymore.

PR #806, opened 2017-12-26
Improve validation errors

tiltec:
Related to issue #668 

- only show vuelidate errors when the form has been 'touched'
- add checklist to docs
- add server-side validation errors in ConversationCompose, GroupPreviewUI
- improve ProfileEdit, VerifyMail, Signup
- fix InvitationForm, Login, PasswordReset

I guess the issue can be closed then.

### DIFF
--- a/docs/book.json
+++ b/docs/book.json
@@ -2,6 +2,7 @@
   "title": "Karrot Developer Documentation",
   "gitbook": ">3.0.0",
   "plugins": [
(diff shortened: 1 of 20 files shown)


tiltec:
Should be good to go!

tiltec:
This also fixes the time dependency of snapshot tests that has been breaking tests since today or so.

PR #807, opened 2017-12-26
Some design changes

### LINKED ISSUE #802: Intuitive editing
Buttons are not consistently recognizable in the design and by that making it very unintuitive> longer time spend navigating the site or even leaving frustrated for first time visitors. 

Example:
The editing for the group has a pencil icon as button while the "check and manage pickups " does not have that icon but instead flashes green when i hover over it....but is apart from that not visibly distinguishable from other surfaces that are non clickable.

I would recommend to give all buttons and clickable surfaces a clear design that makes them easy to find...like for example here in github the green buttons.

tiltec:
A little round of changes that might enhance UX.

Closes #802 

Comparing old (left) and new (right); Sidebar:

image4525

Store page:
(group description page also got the green pencil)

image4549

Now some new features; if group has no stores yet...

image

If store has no pickups yet...

image

What do you think about it?


D0nPiano:
Cool, looks great! Really like the new edit buttons and the new way of handling the cases of not having pickups/stores. Will look through the code changes now.

D0nPiano:
Currently, we use
- http://fontawesome.io/icon/shopping-basket/
- http://fontawesome.io/icon/calendar-o/
- http://fontawesome.io/icon/clock-o/

for pickups. Would be nice to have some consistency there.
We could also create something stacked, like:
https://jsfiddle.net/seoa7mmu/1/

....I think I am for the simple basket, but I can also the the advantages of the other icons.

tiltec:
Thanks for the review and the additions @D0nPiano. Now I finally get what `fa-fw` is good for 😅 

I noticed that the buttons touch each other in mobile view, did you intend this? Looked better to me with the `right` option. Although I agree that it's nicer on desktop now ;)
image

tiltec:
Done for now, I decided to switch to `fa-shopping-basket` as general pickup icon :)

D0nPiano:
Great!
Just added some margin to the buttons.

And also perfect to have a unified icon for pickups.  Two more ideas:
1) I might change the ones on the "manage pickupdates" page back to a calendar or use a stacked one to differentiate the manage page from the pickups page

2) On the store page, when there are no pickups, the button next to the bed sais "you can create a pickup on the store page" which you are on already. Would change the button somewhen.

I can also make those two changes; probably tomorrow.

tiltec:
Sounds good! I think these could be PR to keep the number of changes lower and to get some feedback from people who might use dev.karrot.world from time to time. I'll merge this one now, but not deploy to production.

Issue #763, opened 2017-11-16
Support more than 10 collectors in pickups


tiltec:
I think it would be nice to offer an UI option to set more than 10 collector slots in a pickup and possibly also allow unlimited collectors (by setting it to `null`, I think... need to look in the backend again)

PR #814, opened 2017-12-28
Small Design Improvements no. 4

### LINKED ISSUE #772: Map and email are misaligned on mobile profile page
screenshot_20171117-184621

D0nPiano:
Fixes #772, #763 

See commits for other changes

### Flipped message & datetime
screenshot from 2017-12-29 12-25-13

### Removed buttons, added message for empty description
screenshot from 2017-12-29 12-26-54


tiltec:
Thanks, quite nice! I was unsure about the change regarding the ConversationMessage, but it seems fine now.

Some other bits:
- If I set max slots to null like this, it shows me no pickup slot at all:
image
image
The problem likely existed before, as max slots isn't required for a pickup on the server. Maybe we can show a "+∞ " there?

- the profile page looks better now, but the email still breaks in the middle:
image
Would be nicer if it actually was on top of the map. _and clickable with mailto_ ;)

D0nPiano:
Regarding Pickups: Is there any specific reason to say its unlimited then? I think if someone sets it to 0, it probably needs further explanation in the comment field (if you need to be invited somewhere else, if its some other event (party, cooking etc), if an unlimited amount of people can pick up etc.), wherefore I like it better the way it is now since it seems to be "more flexible" than having an "unlimited people can join" message. 

Mails: Ahh, now I got what you meant. I like the map in the top right corner though, so I removed the arrow, made the text have ellipsis on overflow, made the map smaller on smaller screens and added mailto. I think it looks better not to have it on top, but I am also willing to change it again if you like the other version better.

tiltec:
> Is there any specific reason to say its unlimited then?

That's how the backend behaves; if you set `maxCollectors: null`, it doesn't enforce a maximum.

> I think if someone sets it to 0, it probably needs further explanation in the comment field (if you need to be invited somewhere else, if its some other event (party, cooking etc), if an unlimited amount of people can pick up etc.)

It's not 0, it's null.

This feature is never been requested or used by anyone, so we could also change the backend to require `maxCollectors` to be non-null. What do you think, should I do this backend change to simplify the frontend in this regard? Or could this be a helpful feature to some groups?

tiltec:
> I like the map in the top right corner though, so I removed the arrow, made the text have ellipsis on overflow, made the map smaller on smaller screens and added mailto.

Nice solution! 🏅

D0nPiano:
Ahhh, sorry I misunderstood your comment before. I don't think it's very obvious that it's an unlimited amount of people allowed if there is nothing standing in there. Maybe some sort of Radio-Button would be good ("No People", "Unlimited no people", "Slider + field")?

But for now, until there is a usecase for unlimited people, I would leave the backend and just add boxes to pickups with `maxSlots = null`. Then it's just a secret that it's unlimited when you leave the input field empty ;)

D0nPiano:
-> Added (almost) unlimited amount of boxes and +∞ sign
-> Added "unlimited placeholder" for input field

screenshot from 2017-12-31 14-57-34

src/components/Pickups/PickupUsers.vue
```diff
@@ -100,7 +101,13 @@ export default {
     ...mapGetters({
       currentUser: 'auth/user',
     }),
+    hasUnlimitedPlaces () {
+      return !this.pickup.maxCollectors && this.pickup.maxCollectors !== 0
```
tiltec: `return this.pickup.maxCollectors === null` would be more precise?

PR #815, opened 2017-12-28
Collapsible SidenavBoxes

D0nPiano:
Enables 'Collapse' functionality for sidenav boxes on desktop.

Makes it possible to display group Sidenav all the time, without it taking to much space.
Group sidenav is folded in on storePages automatically now.

On group pages, the StoreSidenav is hidden.

screenshot from 2017-12-28 22-28-22

### DIFF
--- a/src/store/modules/sidenavBoxes.js
+++ b/src/store/modules/sidenavBoxes.js
@@ -1,5 +1,3 @@
-import { withMeta, createMetaModule } from '@/store/helpers'
-
 function initialState () {
   return {
     mapCollapsed: false,
@@ -11,7 +9,6 @@ function initialState () {
 
 export default {
   namespaced: true,
-  modules: { meta: createMetaModule() },
   state: initialState(),
   getters: {
     mapCollapsed: state => state.mapCollapsed,
@@ -20,16 +17,14 @@ export default {
     groupCollapsed: state => state.groupCollapsed,
   },
   actions: {
-    ...withMeta({
-      hideGroupSidenav ({ commit }) {
-        console.log('test1')
-        commit('hideGroup')
-      },
-      showGroupSidenav ({ commit }) {
-        console.log('test2')
-        commit('showGroup')
-      },
-    }),
+    hideGroupSidenav ({ commit }) {
+      console.log('test1')
+      commit('hideGroup')
+    },
+    showGroupSidenav ({ commit }) {
+      console.log('test2')
+      commit('showGroup')
+    },
   },
   mutations: {
     toggleMapCollapsed (state) { state.mapCollapsed = !state.mapCollapsed },
--- a/src/store/modules/sidenavBoxes.js
+++ b/src/store/modules/sidenavBoxes.js
@@ -1,0 +1,44 @@
+import { withMeta, createMetaModule } from '@/store/helpers'
+
+function initialState () {
+  return {
+    mapCollapsed: false,
+    storeCollapsed: false,
+    storesCollapsed: false,
+    groupCollapsed: false,
+  }
+}
+
+export default {
+  namespaced: true,
+  modules: { meta: createMetaModule() },
+  state: initialState(),
+  getters: {
+    mapCollapsed: state => state.mapCollapsed,
+    storeCollapsed: state => state.storeCollapsed,
+    storesCollapsed: state => state.storesCollapsed,
+    groupCollapsed: state => state.groupCollapsed,
+  },
+  actions: {
+    ...withMeta({
+      hideGroupSidenav ({ commit }) {
+        console.log('test1')
+        commit('hideGroup')
+      },
+      showGroupSidenav ({ commit }) {
+        console.log('test2')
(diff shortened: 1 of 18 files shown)


tiltec:
I like this change - makes the sidebar more consistent across group and store.

Maybe it would be clearer to use the same collapse/uncollapse icon also in the wall notifications (availablePickups, joinendPickups)?

If you could run `yarn test -u` to generate updated snapshots and commit them, that would be great! It makes it possible to see if something unexpected changed.

D0nPiano:
Thanks! Do you think the now sidenavBoxes store endpoint makes sense? And closing / opening the group box in the store endpoint?

And I actually wondered how to update those! Will do that.

src/store/modules/sidenavBoxes.js
```diff
@@ -0,0 +1,44 @@
...
+    mapCollapsed: false,
+    storeCollapsed: false,
+    storesCollapsed: false,
+    groupCollapsed: false,
+  }
+}
+
+export default {
+  namespaced: true,
+  modules: { meta: createMetaModule() },
```
tiltec: You don't need the meta module because you're not doing async actions like server requests here. Meta provides pending status and validation errors.

src/store/modules/sidenavBoxes.js
```diff
@@ -0,0 +1,44 @@
...
+  state: initialState(),
+  getters: {
+    mapCollapsed: state => state.mapCollapsed,
+    storeCollapsed: state => state.storeCollapsed,
+    storesCollapsed: state => state.storesCollapsed,
+    groupCollapsed: state => state.groupCollapsed,
+  },
+  actions: {
+    ...withMeta({
```
tiltec: can be removed

tiltec:
> Do you think the now sidenavBoxes store endpoint makes sense? And closing / opening the group box in the store endpoint?

You mean the vuex module? Yes, I think it's nice to have these in state management, gives us a clear place where to control them.

src/store/modules/stores.js
```diff
@@ -62,11 +62,13 @@ export default {
         }
       }
       dispatch('pickups/setStoreFilter', storeId, {root: true})
+      dispatch('sidenavBoxes/hideGroupSidenav', null, {root: true})
```
D0nPiano: I didn't really know where to put the two lines
`dispatch('sidenavBoxes/hideGroupSidenav', null, {root: true})` and
`dispatch('sidenavBoxes/showGroupSidenav', null, {root: true})`

Opening and hiding the groupSidenav when you enter/leave a store page works perfectly here though as far as I know.

tiltec: I think this is the best location. In `routes/main.js`, you will find those lines for the store route

```
          beforeEnter: 'stores/selectStore',
          afterLeave: 'stores/clearSelectedStore',
```
Means that those actions will be triggered every time you visit/leave a store page or a child page.

src/store/modules/sidenavBoxes.js
```diff
@@ -0,0 +1,39 @@
...
+export default {
+  namespaced: true,
+  state: initialState(),
+  getters: {
+    mapCollapsed: state => state.mapCollapsed,
+    storeCollapsed: state => state.storeCollapsed,
+    storesCollapsed: state => state.storesCollapsed,
+    groupCollapsed: state => state.groupCollapsed,
+  },
+  actions: {
```
tiltec: I'm unsure if we should keep the mix of mutations and actions as interface. It seems unnecessary boilerplate to keep the actions as extra layer (you can also `commit('sidenavBoxes/hideGroup')`), on the other hand we mostly used actions in other modules.

Actions have the advantage that they can easily be extended and also call other actions and mutations. The downside is a bit more boilerplate. Should we make a policy here to only have actions as interface? (This also affects the `map` module). @nicksellen your opinion would be valuable here!

nicksellen: I would be inclined to keep to just exposing getters + actions in the store modules. Boilerplate can be reduced by writing functions that generate it. In this case the getters, actions, and mutations could all be generated.

tiltec: Something like this could be nice:

```
import { toggles, switches } from '@/stores/helpers'
function initialState () {
  return {
    map: true,
    store: false,
    stores: true,
    group: true,
  }
}

export default {
  namespaced: true,
  state: initialState(),
   // actions: toggleMap; getters: showMap etc
  ...toggles(['map', 'store', 'stores']), 

   // actions: hideGroup, showGroup; getters: showGroup
  ...switches(['group']),
}
```

- no need to duplicate the 'sidenav' name inside the 'sidenav' module
- getter names should be positive: 'showMap', 'showGroup' (reduces cognitive burden of double negation: `mapCollapsed: false`)
- defaults get set via initialState

What do you think? Could be fun to code a helper like this :)

tiltec: I created #819 - let's go ahead and merge, later refactor.

PR #818, opened 2017-12-30
(Random) store banners

D0nPiano:
More PRs :)

I thought it'd be nice to be able to differentiate stores better. Therefore, I changed the code for random pictures a bit to create random banners and added those to all store-related pages. They are different for different stores, and therefore you are (hopefully) always able to tell that you are (1) still on a store page and (2) which store you are on.

screenshot from 2017-12-30 11-24-25


src/components/General/RandomBanner.vue
```diff
@@ -0,0 +1,112 @@
+<template>
+  <span/>
+</template>
+
+<script>
```
tiltec: Could you refactor a bit to share more code between RandomPicture and RandomBanner?

src/pages/Store/Pickups.vue
```diff
@@ -27,12 +20,19 @@
...
+              class="hoverScale"
             >
               <q-tooltip v-t="'STOREDETAIL.MANAGE'" />
             </q-btn>
           </router-link>
-        </q-item-side>
-      </q-item>
+        </div>
+        <span>
+          <Markdown
```
tiltec: Could be moved one level up and the <span> removed?

src/pages/Store/Layout.vue
```diff
@@ -37,6 +37,11 @@
         <StoreOptions/>
       </q-btn>
     </div>
+    <div class="grey-border store-banner">
+      <RandomBanner
+        style="display: block; height: 5vw; min-height: 30px; max-height: 48px; overflow: hidden"
```
tiltec: Seems enough to put it in a class!

D0nPiano:
Done :heavy_check_mark: 

Yeah, makes sense to move them to one file. :+1:  Especially because when working on the banners, I noticed that the profile pictures had a `viewbox` attribute, the correct one is `viewBox` though. That was the reason, why the profile pictures were non-resizable before. So now you actually don't really need to pass RandomArt a size anymore for profile pics.

Also, I added something to add a title to the banners. Don't know if it's useful yet. Thought of displaying the store name, but removed it again because it would be repeating the topbar. Still not sure about it though. So if anyone has input or any other use, go aheah :laughing: 

store
(text was **removed** again)

tiltec:
Great, thanks!

> I noticed that the profile pictures had a viewbox attribute, the correct one is viewBox though.

Yeah, I sometimes find similar bugs while refactoring. I find it very useful 😄 

> Still not sure about it though. So if anyone has input or any other use, go aheah

How about changing the topbar color/style, depending on the store/group? Might a bit too much though ;)

src/pages/Store/Pickups.vue
```diff
@@ -31,12 +20,20 @@
...
             </q-btn>
           </router-link>
-        </q-item-side>
-      </q-item>
+        </div>
+        <Markdown
+          v-if="store.description"
+          :source="store.description"
+        />
+        <i v-if="!store.description">
```
tiltec: You can use `v-else` here instead of repeating the negated conditional.

src/components/ProfilePictures/ProfilePicture.vue
```diff
@@ -4,12 +4,16 @@
...
-      <RandomPicture
-        :name="user.displayName"
+      <RandomArt
+        :text="user.displayName"
         :seed="user.id"
         :size="size"
+        :width="size + 'px'"
+        :height="size + 'px'"
+        style="display: block; overflow: hidden"
+        :style="{ width: size + 'px' , height: size + 'px' }"
```
tiltec: I'm a bit confused here, why is it necessary to set width/height two times? `RandomArt` doesn't define the `width` and `height` props, are these native attributes?

Issue #639, opened 2017-10-27
Add pickup feedback UI


nicksellen:
Pickup feedback is implemented in the backend. There are also some frontend components available. But no UI for adding feedback.

Perhaps a good place to put it for now is on the main group page at the top (where it shows upcoming pickups), or perhaps in a notification-style thing in the top right?

See https://github.com/yunity/karrot-frontend/issues/159 for broader discussion

D0nPiano:
Since the quesion of user feedback arose again lately, what is our current status?

1. There is a first draft of a feedback-creation form on https://dev.karrot.world/#/group/1/feedback
2. There is a discussion on Notifications / bells:
  2.1 Website Notifications #577 that was closed due to:
  2.2 General discussion on Notifications #257
3. We had a dev feedback UI in angular #605 

**Plans**
So what exactly are our plans? From a UI point, I would
- [ ] Improve feedback creation form (1.)
  - [ ] make it more responsive
  - [ ] make it more form-like (add create button etc.)
- [ ] Add Website notification for possibility to give feedback (related to 2.1)
  - [ ] either as bells in topbar (with facebook-like list)
  - [ ] or as notification on group page (like empty & your pickups)
- [ ] think of a way to display feedback for user, store and group

I could probably start on those pretty soon, but don't really have time / motivation for the store related stuff. Would one of you guys be willing to help me out there? :grimacing: 

**Quesions**
- Does a bells thing make sense from our current backend implementation? Still didn't read into the store logic too much to tell
- Is there an update on the notification endpoint for the api?
- Is the feedback amound for a user / store / group already available backend wise?

tiltec:
I can help with the vuex things and also the backend stuff :) Backend might take a bit longer, depending on the requirements. Maybe we go for a basic implementation without notifications for now, like the one I did in the angular frontend, to get some opinions from actual users.

I would start with the possibility to see the feedback, and then work back to give the feedback.

Maybe something like this?
- A page that lists all feedback per store and per group (we can submit some feedback via API on dev.karrot.world)
- Extend the feedback creation form so that the user can select a pickup to give feedback to
- Put a button to the feedback creation form somewhere
- Deploy and get opinions from users :)

djahnie:
I'm happy to see activity here! ^_^

The minimal proposal from @tiltec would make it possible that every user can give feedback to every pickup, right? So if everyone was to enter an amount if saved food it, what would happen? I guess calculating a mean or median makes sense here.

[...]

PR #822, opened 2018-01-03
Design Improvements no. 5

D0nPiano:
### Feedbacklist:
- amount of saved food for stores in message above

### Layout
- removed right whitespace again (off-centered login boxes before etc.)
  
### Feedback List - @tiltec 's requests in #639 
- [x] pickupFeedback form: highlight "other pickups" button more
- [x] amountviewer: make number less jumpy when sliding, e.g. by always showing the decimal
- [x] Rename 'you took home' -> 'how much food was saved on this pickup' based on feedback from Bruno in https://github.com/yunity/karrot-frontend/issues/159#issuecomment-328352706
- [x] Show group sidenav when in feedback page
- [x] feedbackList: show hint when there is no feedback yet
- [x] FeedbackItem: linkify user name and store name
- [x] use editMixin for pickupFeedback form to also allow editing feedback (see isEditable attribute of feedback to check if editing is allowed)
- [x] Add infinite scroll to feedback list to have proper pagination support
- [x] Remove feedback tab in pickups, add feedback to sidebar instead; I think we shouldn't introduce yet another navigation concept

### Stuff for some other PR
- [ ] Show FeedbackNotice in group/store feedback lists (maybe needs another component layer)
- [ ] pickupFeedback form: show pending status on create button
- [ ] add toast after you created feedback saying "Thank you <q-btn give more feedback/>
- [ ] update available-feedback after user created feedback (currently, x does not change in 'you can give feedback to x pickups' after you gave feedback)


tiltec:
> use editMixin for pickupFeedback form to also allow editing feedback (see isEditable attribute of feedback to check if editing is allowed)
> Add infinite scroll to feedback list to have proper pagination support

I'll add these in another PR.

> perhaps change default ordering of feedback to sort by pickup date, not by feedback creation date

I would change this directly in the backend if we agree on it.

src/components/Pickups/PickupItem.vue
```diff
@@ -10,7 +10,7 @@
           <slot>Date or Store Slot</slot>
         </div>
         <div
-          class="description multiline"
+          class="description"
```
tiltec: Why did you change this? The pickup comment is supposed to support line breaks, as the input field is a textarea.

src/pages/Store/Feedbacks.vue
```diff
@@ -1,10 +1,65 @@
...
+  components: { FeedbackList, QCard, QTooltip, QBtn },
+  computed: {
+    ...mapGetters({
+      feedback: 'feedback/filtered',
+      fetchStatus: 'feedback/fetchStatus',
+    }),
+    totalAmount () {
+      let amount = 0
+      for (let fb in this.feedback) {
+        console.log(this.feedback[fb])
```
tiltec: Remove console.log ;)

src/components/Statistics/FeedbackItem.vue
```diff
@@ -72,9 +80,16 @@ export default {
...
     },
+    storeId () {
+      const { about: { store: { id } = {} } = {} } = this.feedback
+      return id
+    },
     userName () {
       return this.feedback && this.feedback.givenBy && this.feedback.givenBy.displayName
     },
+    userId () {
+      return this.feedback && this.feedback.givenBy && this.feedback.givenBy.id
```
tiltec: I think we need a nicer way to resolve properties without throwing errors, maybe I'll add a helper soon...

src/components/Pickups/PickupItem.vue
D0nPiano: Multiline adds a linebreak at the very beginning, making it look like this:
multiline-on

Without it, it still supports line breaks, and looks like this:
multiline-off

src/pages/Store/Feedbacks.vue
D0nPiano: Oups :laughing:

D0nPiano:
What about:
feedbackitem

I think its better than before, but I'm still not perfectly sure about the header (Store & date) - definitely open for more ideas :D

D0nPiano:
What about:

feedbackitem2


I think its better than before, but I'm still not perfectly sure about the header (Store & date) - definitely open for more ideas :D

src/components/Pickups/PickupItem.vue
tiltec: I mean intentional linebreaks, when somebody hits 'enter' to create a newline.

tiltec: I entered some linebreaks. It should show like this:

image

With this PR, it shows like this:

image

tiltec:
The new design looks good! I would remove the comma after the date though.. similar to the group pickup view:

image

src/components/Pickups/PickupItem.vue
tiltec: The preceding space is annoying though, I wonder where it comes from...

D0nPiano: Haha, I found it! :laughing: it was in the HTML - see my last commit

src/components/Statistics/FeedbackItem.vue
```diff
@@ -28,8 +28,12 @@
               <DateAsWords :date="createdAt" />
             </span>
           </small>
-          <div class="comment">{{ comment }}</div>
-          <div>
+          <div
+            v-if="comment"
+            class="comment">
```
tiltec: Ah, I just saw that the comment input is a textarea. So it would make sense to make it `multiline` too.

D0nPiano: Yeah, makes sense. Was also just wondering about that. -> Changed it :white_check_mark:

PR #833, opened 2018-01-07
Made some small css improvements and fixed the cancel button

meersschautarnaud:
Removed some unnecessary white spaces after optimizing the CSS.
The cancel button, from the create store form, now works.

### DIFF
--- a/src/components/Store/StoreEdit.vue
+++ b/src/components/Store/StoreEdit.vue
@@ -108,7 +108,7 @@
             </q-btn>
             <q-btn
               type="button"
-              @click="$emit('cancel')"
+              @click="$router.push('/group/' + $route.params.groupId)"
               v-if="isNew"
             >
               {{ $t('BUTTON.CANCEL') }}
--- a/src/components/Store/StoreEdit.vue
+++ b/src/components/Store/StoreEdit.vue
@@ -108,7 +108,7 @@
             </q-btn>
             <q-btn
               type="button"
-              @click="$router.push('/group/' + $route.params.groupId)"
+              @click="cancel()"
               v-if="isNew"
             >
               {{ $t('BUTTON.CANCEL') }}
@@ -215,6 +215,9 @@ export default {
           },
         ],
       })
+    },
+    cancel () {
+      this.$router.go(-1)
     },
   },
   validations: {
--- a/src/__snapshots__/storyshots.spec.js.snap
+++ b/src/__snapshots__/storyshots.spec.js.snap
@@ -116,7 +116,7 @@ exports[`Storyshots Conversation Conversation 1`] = `
                         <div class="q-item-">
                             <div class="no-wrap">
                                 <div to="[object Object]"><span class="text-bold text-secondary uppercase">Mira Bellenbaum</span></div> <span class="message-date"><small class="light-paragraph"><div>
-  4 months ago<span class="q-tooltip animate-scale" style="transform-origin:bottom center 0px;">Aug 11, 2017, 3:43 PM</span></div>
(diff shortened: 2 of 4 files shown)


src/components/Store/StoreEdit.vue
```diff
@@ -108,7 +108,7 @@
             </q-btn>
             <q-btn
               type="button"
-              @click="$emit('cancel')"
+              @click="$router.push('/group/' + $route.params.groupId)"
```
tiltec: This works, but I would say that it does the wrong thing and it does it at the wrong place. 

- I think it should go back to the previous page, not the the group home page.
- I wouldn't trigger the router push here, but in the parent component. That keeps the `StoreEdit` component more isolated.
- We always use named routes instead of string concat

As the parent component is a connect component, you'll need something similar to this: https://github.com/yunity/karrot-frontend/blob/master/src/pages/GroupGallery.vue#L14

tiltec:
Tests fail because you changed the templates; run `yarn test -u` to update the snapshots and commit them. [More info on snapshot testing](https://facebook.github.io/jest/docs/en/snapshot-testing.html#snapshot-testing-with-jest)

tiltec:
Interesting, somehow `jest` didn't pick up UTC as time zone and instead used `Europe/Brussels`, which is different on our CI server... I wonder why this happened...

Can you verify that `yarn test -u` outputs the following? 

```
yarn run v1.3.2
node test/jestUTC.js -u
...
```

src/components/Store/StoreEdit.vue
```diff
@@ -216,6 +216,9 @@ export default {
         ],
       })
     },
+    cancel () {
```
tiltec: What's your reasoning putting it into the component vs putting it in the parent?

meersschautarnaud: There are 2 vue files (Create.vue and Edit.vue) that use the same StoreEdit.vue file. In stead of changing 2 files, I chose to edit just 1.
`yarn run v1.3.2
$ node test/jestUTC.js -u`
This is the output when I run `yarn test -u`

tiltec: > There are 2 vue files (Create.vue and Edit.vue) that use the same StoreEdit.vue file. In stead of changing 2 files, I chose to edit just 1.

That's a valid goal, but not necessarily the best when it comes to software architecture.

We're trying to approach the concept of [presentational components](https://medium.com/@dan_abramov/smart-and-dumb-components-7ca2f9a7c7d0) and we tried to split it so that presentational components go into `src/components/`, while container components go into `src/pages/`.

By that logic, I would say that routing decisions belong to the container component, because it's a state transition.

We're not perfect with the split, but I would say that `StoreEdit` is one of the more _pure_ presentational components in karrot and I would like to keep it that way.

tiltec:
I just noticed that the "cancel" button is only shown on the "create store" page, not the "edit store" page. I wonder if there was any UX decision regarding this or it just happened via copy-paste. How should we go about here?

(Maybe input from @D0nPiano and @nicksellen could be helpful)

D0nPiano:
The design changes make sense (I think the whitespace one is partly fixed already so there mitght be conflicts, but I like this solution even better). Thanks @meersschautarnaud! 

@tiltec I don't think there was a decision to not show them - Adding cancel buttons there would make sense in my opinion.

tiltec:
Store edit will have 4 buttons then: cancel, reset, archive, save .... seems a lot.
My guess is that we can remove the "reset" button?

tiltec:
Reworked in #849

PR #840, opened 2018-01-11
add weight input

tiltec:
Refs #159 request from @nicco11a
input box is still a bit misplaced...

image

I tried to add `class="row"` but the damn thing didn't move...

### DIFF
--- a/src/components/Statistics/AmountPicker.vue
+++ b/src/components/Statistics/AmountPicker.vue
@@ -52,7 +52,7 @@ export default {
         return Math.min(70, this.value)
       },
       set (v) {
-        this.value = v
+        this.$emit('input', v)
       },
     },
     valueToNumber: {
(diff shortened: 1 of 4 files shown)


D0nPiano:
Makes sense to have an input field there (just forgot to put one there :smile: ) Just improved some style & handling of very large numbers

scale

src/components/Statistics/AmountPicker.vue
```diff
@@ -16,28 +16,55 @@
...
+      set (v) {
+        this.value = v
+      },
+    },
+    valueToNumber: {
+      get () {
+        return this.value
+      },
+      set (v) {
+        const value = parseFloat(v, 10)
```
D0nPiano: Just changed parseInt to parseFloat - or was there a specific reason to use Int?

tiltec: by mistake :)

src/components/Statistics/AmountViewer.vue
```diff
@@ -67,6 +70,9 @@ export default {
         }
         amount -= 0.15
       }
+      if (amountImages.length > 6) {
```
tiltec: shouldn't this be `> 9`?

tiltec: I think you can leave the check out, slice does it for you:

> If end is greater than the length of the sequence, slice extracts through the end of the sequence (arr.length).

PR #848, opened 2018-01-15
Conversation improvements

### LINKED ISSUE #770: Update presence status
The backend supports keeping track of prescence (is the user actually at the computer). This is to enabler smart decisions about whether to send push notifications.

It uses the websocket connection to update it, see [the tests](https://github.com/yunity/karrot-backend/blob/master/foodsaving/subscriptions/tests/test_consumers.py#L54-L66) for what to send.

Some of the following libaries/apis might be useful:
- https://developer.mozilla.org/en-US/docs/Web/API/Page_Visibility_API (if this is well supported could be good, although doesn't detect idling I guess)
- https://github.com/serkanyersen/ifvisible.js/
- https://github.com/defunctzombie/away

### LINKED ISSUE #845: Display unread status and mark messages as read
Conversation has these fields

```
    "seen_up_to": null,
    "unread_message_count": 288
```

and can mark messages as seen:

`/api/conversations/{id}/mark/` with `{ "seen_up_to": "{message_id}" }`

Maybe we scroll down to the first unread message (in group wall)? And when the user scrolls back up, mark all as read.

tiltec:
Closes #770 
Closes #845 

Listen for user activities and report presence to backend
Show unread notification status and show button to mark all read

If the user has unread messages and writes a new message, all messages get marked as read. This is caused by our backend datastructure.

Notes:
I had the impression that the frontend got a bit more sluggish on startup after I implemented #770 
UPDATE: It feels alright now.

### DIFF
--- a/src/store/plugins/presenceReporter.js
+++ b/src/store/plugins/presenceReporter.js
@@ -9,14 +9,8 @@ export default store => {
     timer = setTimeout(setAway, 30 * 1000) // 30 seconds
   }
 
-  // listen on events
-  document.onload = resetTimer
-  document.onmousemove = resetTimer
-  document.onmousedown = resetTimer // touchscreen presses
-  document.ontouchstart = resetTimer
-  document.onclick = resetTimer // touchpad clicks
-  document.onscroll = resetTimer // scrolling with arrow keys
-  document.onkeypress = resetTimer
+  ['load', 'mousemove', 'keydown', 'DOMMouseScroll', 'mousewheel', 'mousedown', 'touchstart', 'touchmove', 'click']
+    .forEach(event => document.addEventListener(event, resetTimer))
 
   // listen on tab & window visibility, mostly works on desktop browser
   Events.$on('app:visibility', state => {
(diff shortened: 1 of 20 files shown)


src/store/plugins/presenceReporter.js
```diff
@@ -0,0 +1,25 @@
...
+  }
+
+  // listen on events
+  document.onload = resetTimer
+  document.onmousemove = resetTimer
+  document.onmousedown = resetTimer // touchscreen presses
+  document.ontouchstart = resetTimer
+  document.onclick = resetTimer // touchpad clicks
+  document.onscroll = resetTimer // scrolling with arrow keys
+  document.onkeypress = resetTimer
```
nicksellen: Using `addEventListener` calls is more proper way to add events listeners, and doesn't suffer from conflicts if something else adds events listeners

tiltec: Good catch. I didn't think about it very much after copying it from Stackoverflow...

PR #850, opened 2018-01-17
Small design improvements no. 6

D0nPiano:
I included a workaround for making the search work on every character. Couldn't find any better solution.

See commits for other changes


src/routes/main.js
```diff
@@ -71,21 +70,6 @@ export default [
       default: GroupCreate,
     },
   },
-  {
-    name: 'historyDetail',
-    path: '/history/:historyId',
```
tiltec: By removing the route, you break existing URLs, which I consider bad user experience and generally not OK. It might be ok in this special case (I suppose that history detail URLs haven't been shared among people), but it would be better to have at least some kind of fallback (alias or redirect).

D0nPiano:
Definitely can see your points, and changed the PR accordingly.

> Splash pages are left-aligned instead of centered (like last time), is this on purpose?

Nope, just forgot about it again. should be centered now.

> By removing the route, you break existing URLs

Readded the old history page. Furthermore, when you open a history item on the new page, the URL (`group/x/history/`) changes to the older URL (`/histroy/yyy`).
I am not sure if replacing the current history state is the best way of doing it, but [the according vue-router issue](https://github.com/vuejs/vue-router/issues/703) didn't look too promising for other solutions.

> I wouldn't remove the groupOptions in desktop view

Readded the buttons on desktop

src/themes/app.mat.styl
```diff
@@ -102,4 +102,16 @@ body.desktop .desktop-margin
     clear both
     display block
     visibility hidden
-    height 0px
\ No newline at end of file
+    height 0px
+
+.slide-toggle-enter-active,
```
tiltec: This `app.mat.styl` file seems to become a collection of unscoped styles.

Maybe try to put the styles into more specific places, e.g. put the transition styles into a new file and import it where needed?
Same for the `.edit-box` class, there's no need to have it site-wide.

src/components/Statistics/PickupFeedback.vue
```diff
@@ -121,19 +132,30 @@ export default {
...
-  margin-top 2.5em
+  margin-top 2.5em !important
   .randomBanner
     display: block
     height: 26px
     overflow: hidden
 </style>
+
+<style lang="stylus">
+.q-field-dark.grey-font
```
tiltec: Please use a custom class to avoid matching unrelated components, e.g. `.pickup-feedback .q-field-dark .grey-font`. Unscoped CSS can get annoying...

src/components/History/HistoryEntry.vue
```diff
@@ -48,8 +62,36 @@ export default {
...
-  components: { ProfilePicture, DateAsWords, QItem, QItemSide, QItemMain, QItemTile },
+  data () {
+    return {
+      detailIsShown: false,
+    }
+  },
+  methods: {
+    toggleDetail (event) {
+      if (this.detailIsShown) {
+        window.history.replaceState({}, null, `#${this.$route.path}`)
```
tiltec: What about `this.$route.fullPath`? Shouldn't need the additional `#` then.

src/components/History/HistoryEntry.vue
```diff
@@ -48,8 +62,36 @@ export default {
...
+      }
+      else {
+        window.history.replaceState({}, null, `#/history/${this.entry.id}`)
```
tiltec: The component should remain agnostic of the exact route structure. You can (probably) achieve this with `this.$router.resolve({ name: 'historyDetail', params: { historyId: this.entry.id } }).href`.

D0nPiano: Oh, fancy! Makes sense :slightly_smiling_face: :+1:

src/components/Statistics/PickupFeedback.vue
D0nPiano: I actually didn't find a way to do it using scoped css, since quasar is creating the component there. The only way of changing the input field is using unscoped css somewhere I believe...

But I added a css class for pickup-feedback now and made it a subclass (I think that was also your suggestion)

src/themes/app.mat.styl
D0nPiano: Yeah, makes sense. I just moved slide-toggle & edit-box to their own files. Will leave hoverScale in app.mat for now though.

Issue #670, opened 2017-10-30
Display membership roles in UI


nicksellen:
Idea is to include it on the group members page https://karrot-dev.foodsaving.world/#/group/6/members

tiltec:
Seems to be connected to having minimal admin roles.

nicksellen:
They exist already! Check the API response for getting group info, has "memberships" field or something like that iirc

tiltec:
I raised the priority of this issue to Release 5 because the other "minimal admin roles" issues are in there. I think we need to show the membership roles when group agreement have an effect.

tiltec:
Stepping down from this because it seems to go into the wrong direction. My thinking is that we should allow group members to self-select roles that gives them certain rights and require explicit approval from other group members (one, some or all) if the action has big effects.

nicksellen:
We realised we have some uncertainty about the approach to take with roles and permissions, and this task is a bit too concrete for now.

Issue #550, opened 2017-07-03
Minimal implementation of admin roles


nicksellen:
With @tiltec and @djahnie we discussed the admin role features that would be needed for implementing #324.

#356 discusses the wider topic of additional roles and is still relevant for the near future.

For #324 all that is needed is some set of users within the whole group that are able to edit the legal agreement. We realised we don't need to restrict other actions yet (e.g. ability to add stores).

In the future we would like to implement more non-hierarchical methods of managing admins but it is more realistic to start with a simple system.

We discussed whether having admins in a group should be optional or not, and realised if it was optional then it would be difficult to turn it on again in the future: if nobody is an admin, who has permission to enable admins? (this would be the case for existing groups, or new groups that initially selected not to have admins).

One option would be to have an owner of the group (by default the original creator, but could be reassigned) - but then we have just added an extra level of hierarchy and not solved anything.

Based on thoughts above we came up with the following concept:
1. each group must have at least one admin
2. if a group does not have an admin, then the oldest member of the group is promoted to admin (to comply with point 1)
3. only admins can edit the legal agreement
3. only admins can create other admins
4. only admins can remove other admins (they can remove the original admin of the group as all admins have equal powers)

This should allow us to implement the smallest possible admin system that allows us to continue with #324, whilst also leaving the future open to more admin-specific roles, and to build more mechanisms to create and remove admins.

djahnie:
Tiny clarification in case someone misunderstood:
The _oldest_ member is the one who joined first. Age doesn't make any difference here... 😉

Sliverriver:
i am starting to feel uncomfortable with the word "admin" probably because of many bad hierarchie examples missuse of power from foodsharing and facebook groups.

isn't it more like you have a feature - and you assign people to this feature. and not having a role that has 234234 features. 

or you have feature 1,2,3 and make a virtual role a that has feature 1 and a role b that has feature 1 and 2. and you assign the user to the role. 

that is also something that got a bit to much on foodsharing. like you have the orgateam(role admin) and every "admin" function was put into that role. which made it less flexible to give one function to person without giving them all access. 

How can we assign "powerfull" features (features that give power over other users) and prevent the missuse like on fs.de ? 
is it "if you gain a privilege you lose another?" for example if i can ban a user from the group i am not allowed to do pickups ? just brainstorming to give an idea.

Sliverriver:
[...]

PR #854, opened 2018-01-21
WIP: Very simple user role management

tiltec:
Related to #670 and #550


D0nPiano:
What is the status of this PR? Is it waiting for something specific, or does it just need approval?

tiltec:
It looks a bit shitty and I don't know what it's actually good for, so I just see it as a proof of concept for now. I wouldn't merge it as is.

D0nPiano:
Ah, ok, makes sense. I like that it enables you to pick who is able to change the group agreement - maybe I'll think of some other UI for it these days then...

tiltec:
I don't think the group agreement is useful how it's currently implemented, basically it's just a text message. The users can still do everything even when they don't accept the agreement.

D0nPiano:
Ah, okay, didn't know they were able to still do everything without accepting the agreement. Guess it really does not make sense now.

Issue #857, opened 2018-01-30
Fix image serving for cordova app


nicksellen:
Some images are rendering with the full domain e.g. `https://dev.karrot.world/media/__sized__/user__photos/photo_tytYiOv-3227be82cdb6ab52127c1936654cb57a.jpeg` - this is good, and how it should be.

Other images are broken as they are missing the full domain e.g. `/media/__sized__/user__photos/photo_07qXsXD-3227be82cdb6ab52127c1936654cb57a.jpeg`

Not sure why the difference right now.

tiltec:
Oh, that's my fault. I added some code to the frontend to strip the domain because it made some problems while developing locally (especially when proxying to dev.karrot.world). Maybe we should switch it off in deploment.

nicksellen:
You could just enable it under cordova. There is a global variable in the webpack config that you can check.

nicksellen:
> Oh, that's my fault. I added some code to the frontend to strip the domain because it made some problems while developing locally (especially when proxying to dev.karrot.world). Maybe we should switch it off in deploment.

It's also because the websocket sends the image urls back without the domain, so even if it didn't strip it, it still wouldn't work for the logged in user. Other users aren't affected by either of those things and their profile images still show.

So, we should make the backend consistent, either including the domain or not. I would be inclined to keep the domain in the URL (as potentially the images get served by another domain at some point, and that's something for the backend to know about). That would require modifying the websocket part.

What was the problem with local development specifically? I would be inclined then to solve that independently of anything else.

nicksellen:
I made it include domain via the socket in the backend https://github.com/yunity/karrot-backend/pull/476

tiltec:
I guess it should be fixed, but I couldn't try it out yet because https://dev.karrot.world/app.apk seems to be an invalid apk.

PR #858, opened 2018-01-31
Keep domain in image urls

nicksellen:
Will await feedback from @tiltec about issues that were experienced in development (see https://github.com/yunity/karrot-frontend/issues/857).


nicksellen:
If I try and upload an image locally I get a security/CORS related error:

```
vue-croppa.js?f3a8:824 Uncaught (in promise) DOMException: Failed to execute 'toBlob' on 'HTMLCanvasElement': Tainted canvases may not be exported.
    at VueComponent.generateBlob (webpack-internal:///1146:824:19)
    at VueComponent.boundFn [as generateBlob] (webpack-internal:///5:190:14)
    at eval (webpack-internal:///1146:839:31)
    at new Promise (<anonymous>)
    at VueComponent.promisedBlob (webpack-internal:///1146:837:14)
    at VueComponent.boundFn [as promisedBlob] (webpack-internal:///5:190:14)
    at VueComponent._callee$ (webpack-internal:///955:101:48)
    at tryCatch (webpack-internal:///213:62:40)
    at Generator.invoke [as _invoke] (webpack-internal:///213:296:22)
    at Generator.prototype.(anonymous function) [as next] (webpack-internal:///213:114:21)
```

Although doesn't seem like it would be related to the whether the returned image URLs include the domain. Have not investigated further right now.

nicksellen:
Aha, I see the issue. The component cannot use an image from another domain, and the initial image URL (from the existing image) is from the backend domain, and I think this prevents it from doing anything else later.

I've modified it to serve the profile image from localhost:8080 during development only.

Issue #852, opened 2018-01-19
On verifying new e-mail address, the old address is shown


mr-kojo:
If I change my e-mail address from `a@b.cd` to `b@c.de` and open the e-mail verification link, the page says:

> a@b.cd
> Your e-mail address was successfully verified!

It would be better if it showed

> b@c.de
> Your e-mail address was successfully verified!

I observed this on Firefox and Chrome, both with the release and the dev version of karrot.

paavininanda:
Hi i would like to work on this issue.

PR #859, opened 2018-02-02
Correct Email address is displayed upon verification

paavininanda:
- [x] No change in coverage
- [x] All the tests passed

Resolves https://github.com/yunity/karrot-frontend/issues/852


tiltec:
Seems right. I think `unverifiedEmail` is always defined, so that should work.

PR #867, opened 2018-02-07
First commit for the Back To Top Button

searchingfor:
Hello,
First time contributing on GitHub. Let me know if I have done anything wrong here, I tried my best to follow the rules. I appreciate any feedback.

Also as a side note, I'm not a fan of how the button hides under the footer once you scroll to the bottom. Wasn't sure how to fix it without messing with the footer.

Thanks!

### DIFF
--- a/config/index.js
+++ b/config/index.js
@@ -53,7 +53,7 @@ module.exports = {
     // auto open browser or not
     openBrowser: true,
     publicPath: '/',
-    port: 3000,
+    port: 8080,
 
     // If for example you are using Quasar Play
     // to generate a QR code then on each dev (re)compilation
--- a/config/index.js
+++ b/config/index.js
@@ -53,7 +53,7 @@ module.exports = {
     // auto open browser or not
     openBrowser: true,
     publicPath: '/',
-    port: 8080,
+    port: 3000,
 
     // If for example you are using Quasar Play
     // to generate a QR code then on each dev (re)compilation
--- a/src/App.vue
+++ b/src/App.vue
@@ -1,8 +1,10 @@
 <template>
   <!-- Don't drop "q-app" class -->
+
   <div id="q-app">
     <LoadingProgress />
     <router-view />
+    <BackToTopBtn />
   </div>
 </template>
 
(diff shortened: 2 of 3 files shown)


src/components/BackToTopBtn.vue
```diff
@@ -0,0 +1,35 @@
+<template>
+  <q-btn
+    v-back-to-top
+    round
+    class="fixed-bottom-right"
+    style="margin: 0 15px 15px 0; background-color: #4a3520"
```
tiltec: Can you move these styles into a scoped class?

Also, I think it's best if you use one of our theme colors, for example by adding `bg-primary` as class.
If none of the theme colors fits, we have to rework our theme ;)

src/components/BackToTopBtn.vue
```diff
@@ -0,0 +1,35 @@
...
+  QBtn,
+  QIcon,
+  QFixedPosition,
+  BackToTop,
+} from 'quasar'
+export default {
+  components: {
+    QBtn,
+    QIcon,
+    QFixedPosition,
```
tiltec: You don't use QFixedPosition, right? So it could be removed.

src/App.vue
```diff
@@ -11,7 +13,10 @@
  * Root component
  */
 import LoadingProgress from '@/components/Layout/LoadingProgress'
```
tiltec: Better use absolute instead of relative imports, and "@" as alias for "src/" (see line above)

searchingfor:
@tiltec,

What is the best way to get a mobile view of the app?

tiltec:
Depends on your browser, for example:
https://developer.mozilla.org/en-US/docs/Tools/Responsive_Design_Mode
https://developers.google.com/web/tools/chrome-devtools/device-mode/

config/index.js
```diff
@@ -53,7 +53,7 @@ module.exports = {
     // auto open browser or not
     openBrowser: true,
     publicPath: '/',
-    port: 8080,
+    port: 3000,
```
tiltec: Unrelated. Why did you change the port?

searchingfor: My browser was having some kind of glitch and I had to change the port to get it to stop. Sorry about that.

tiltec: Please remove before I can merge.

PR #875, opened 2018-02-09
Fix for #870

### LINKED ISSUE #870: Searchbar resizes after each keystroke
Whenever I type something into the search bar it seems to expand and contract very quickly.

searchingfor:
Closes #870.

Edit: Ah nevermind, I've found devices this doesn't work on.

### DIFF
--- a/src/components/Layout/KTopbarUI.vue
+++ b/src/components/Layout/KTopbarUI.vue
@@ -35,9 +35,7 @@
         >
           <q-icon name="fa-fw fa-arrow-right"/>
         </q-btn>
-        <div>
-          <Search style="margin-top: .2em; vertical-align: middle; width: 245px"/>
-        </div>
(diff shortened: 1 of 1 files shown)


tiltec:
Ping @D0nPiano because I'm not really a style expert.

Did it work fine on small screens, e.g. mobile phone?

D0nPiano:
@searchingfor thanks for taking care of #870

I think it would be best for it to not have a specific width (always annoying when it comes to mobile), therefore I would propose something like giving the `Search` component 100% width (and maybe removing the div around it?), and adding something like: 
```
  width: 100%
  min-width: 200px
  max-width: 400px
```
to the `.searchbar` class

I think it would solve the problem (didn't test it yet), and the search bar would still be bigger on desktop than on mobile.

If that doesn't work, than your approach seems better than the current solution though.

PR #879, opened 2018-02-15
enabled ctrl+enter as submit for textareas

Anjalizi:


src/components/Group/AgreementForm.vue
```diff
@@ -25,6 +25,7 @@
           @blur="$v.edit.content.$touch"
           type="textarea"
           :min-rows="20"
+          @keyup.ctrl.enter="send"
```
tiltec: How does it work? There is no send method defined on the controller, right? I guess it should be the same as the form submit action, "maybeSave".

src/components/Store/StoreEdit.vue
```diff
@@ -15,7 +15,7 @@
             <q-input
               v-model="edit.name"
               :autofocus="true"
-              autocomplete="off"
+              :autocomplete="off"
```
tiltec: Why does it need a binding instead of a plain static attribute?

Anjalizi: I will rectify this.

src/components/Group/AgreementForm.vue
Anjalizi: I will check on this as well

Anjalizi:
I have absolutely no idea on how to fix the codecov report issues. Please guide me and review the PR as well.

tiltec:
The Codecov failures can only be resolved by adding automated unit and integration tests. You can leave them away for this PR, as it's only a small change.

tiltec:
Did you check if it works with the right Ctrl key? I think last time I checked it just worked with the left Ctrl key.

src/components/Settings/ProfileEdit.vue
```diff
@@ -93,6 +93,12 @@ export default {
       }
     },
   },
+  methods: {
+    maybeSave () {
```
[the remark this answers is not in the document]
Anjalizi: I should remove the maybeSave method from ProfileEdit.vue and should write 
`
@keyup.ctrl.enter="send"
`
instead?
By canSave do you mean maybeSave? Is that a typo?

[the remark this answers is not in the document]
tiltec: You add this code:

```
  methods: {
    maybeSave () {
      if (!this.canSave) return
      this.save()
    },
  },
```

`canSave` is not defined in this component, so the code will not work. But you don't need the code at all, because you can directly use `save` as you wrote: `@keyup.ctrl.enter="send"`.

`canSave` is used in other components to check if the user filled all necessary form fields and prevent saving if this is not the case.

src/components/Group/GroupEdit.vue
```diff
@@ -47,6 +47,7 @@
                 v-model="edit.description"
                 type="textarea"
                 :min-rows="3"
+                @keyup.ctrl.enter="maybeSave"
```
tiltec: There's another `textarea` above this one ("edit.publicDescription"), can you add the handler there as well?

src/components/Pickups/PickupSeriesEdit.vue
```diff
@@ -138,6 +139,7 @@
           v-model="edit.description"
           type="textarea"
           max-length="500"
+          @keyup.ctrl.enter="send"
```
tiltec: `send` is not defined in this component, did you mean `maybeSave`?

Issue #861, opened 2018-02-04
Make fullscreen map available for desktop


tiltec:
The fullscreen map is currently only available to mobile users, by clicking on the bottom nav "Map" button: 

![image](https://user-images.githubusercontent.com/4410802/35777892-14e7bfcc-09b6-11e8-92cd-9f161006dce3.png)

https://dev.karrot.world/#/group/6/map (link only works if you joined the _01_testgroup_with_maps_and_password_ before)

Tasks:

- add links to the fullscreen map on desktop view, too
- make the fullscreen map look nice on all screen sizes
- (optional) make sure that the user ends up in the same map view when using browser back navigation
- (optional) sync the current zoom level and position to the URL, so that copy/paste-ing the URL puts the user into the same map view

LyraXaph:
I would like to try this.

PR #872, opened 2018-02-08
Disable save photo button if photo wasn't changed

LyraXaph:
Ready for review :)  Fixes #860

PR #880, opened 2018-02-16
Make fullscreen map available for desktop

LyraXaph:
Fix for #861, still working on tasks 3 and 4 (optional ones).

Had trouble with tests again, tried to work around it by manually updating snapshot files.

### DIFF
--- a/src/components/Map/StandardMap.vue
+++ b/src/components/Map/StandardMap.vue
@@ -76,11 +76,11 @@ export default {
       type: Boolean,
       default: false,
     },
-    centerUrl: {
+    forceCenter: {
       type: Object,
       default: null,
     },
-    zoomUrl: {
+    forceZoom: {
       type: Number,
       default: null,
     },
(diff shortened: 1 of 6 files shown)


tiltec:
Were the snapshot problems time-related?

LyraXaph:
Yes I think so, same as last time.. (#872)

src/pages/Map.vue
```diff
@@ -24,6 +25,17 @@ export default {
...
   },
+  data () {
+    return {
+      zoomLvl: null,
+    }
+  },
+  methods: {
+    mapZoomed (zoomLvl) {
+      this.$router.replace({ query: {zoom: zoomLvl} })
+      this.zoomLvl = Number(zoomLvl)
```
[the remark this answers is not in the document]
LyraXaph: I was thinking of using it to pass the value to the map component. But I don't think its a good solution (still WIP).

LyraXaph:
I think I am done with this, not sure if the checks in the StandardMap component that I added are the nicest solution tho.

src/components/Map/StandardMap.vue
```diff
@@ -75,6 +76,14 @@ export default {
       type: Boolean,
       default: false,
     },
+    centerUrl: {
```
tiltec: I'd call those props (including the parent ones) `forceCenter` and `forceZoom` to make their behavior clearer. They don't carry a URL in them, after all.

src/components/Sidenav/SidenavMapUI.vue
```diff
@@ -10,6 +10,22 @@
       slot="tools"
       class="tools"
     >
+      <q-btn
```
tiltec: Ah, I found out why the link doesn't work! It seems that nesting `router-link` inside `q-btn` doesn't work. It has to be the other way round. Unfortunately, this creates a problem where the icon would disappear on hovering, so I'd propose to add a `click` handler to `q-btn` and remove the `router-link` completely:

```
@click="$router.push({name: 'map', params: {groupId: currentGroup.id}})"
```

tiltec:
I added the snapshots and rebased to master :)

PR #888, opened 2018-02-19
Chrome Pickup Sort Fix

### LINKED ISSUE #873: Wrong pickup list order in Chrome
It should be sorted by date, ascending order. It works in Firefox, but not in Chrome. It also affects our mobile app:

image

Investigate how this happens and explore ways to fix it.

searchingfor:
Couple things: 
1. I don't know if doing the sort within the component is bad practice/not what you want.
2. For the line  'return this.pickups.slice(0).sort(compare)', I added the slice(0) because the linter would not accept it without me doing that. I don't know a better solution or if if this is even doing anything.

OK, Cheers 👍 

closes #873

### DIFF
--- a/src/store/modules/pickups.js
+++ b/src/store/modules/pickups.js
@@ -33,7 +33,7 @@ export default {
       }
     },
     all: (state, getters, rootState, rootGetters) => {
-      return state.idList.map(getters.get)
+      return state.idList.map(getters.get).sort(sortByDate)
     },
     filtered: (state, getters) => {
       return getters.all.filter(e => !state.storeIdFilter || (e.store && e.store.id === state.storeIdFilter))
--- a/src/store/modules/pickups.js
+++ b/src/store/modules/pickups.js
@@ -33,7 +33,7 @@ export default {
       }
     },
     all: (state, getters, rootState, rootGetters) => {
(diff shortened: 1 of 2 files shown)


src/store/modules/pickups.js
```diff
@@ -239,6 +239,8 @@ export function isWithinOneWeek (pickup) {
   return pickup.date < new Date(+new Date() + 6096e5)
 }
 
-export function sortByDate (a, b) {
-  return a.date > b.date
```
tiltec: Would it be sufficient to replace the line with this:
```
return a.date - b.date;
```

It seems that `a.date > b.date` behaved wrong because it returned true or false instead of integers.

searchingfor:
Interestingly, if I don't use a sort function at all in modules/pickup.js (after I have removed the one I added), it is now magically sorted in chrome and firefox? Any idea why this is? @tiltec

tiltec:
@searchingfor The backend return the pickups as sorted list, so that should work out of the box. The additional sorting becomes important when handling with data updates while the page is open, for example a change in pickup time or the creation of a new pickup.

tiltec:
Great, that looks pleasing to my eyes and seems to work fine :) Thanks @searchingfor!

Issue #877, opened 2018-02-12
Add filtering option for groupPreview


D0nPiano:
When you are logged out, there is no way of filtering all groups on karrot. My proposal would be adding a search bar underneath `groupPreview`'s "Which group do you want to join?" headline for when you are logged out.

tiltec:
I would like a location filter, perhaps we can get an approximate location of the user (GeoIP and/or browser location API) and show groups close to the user first.

tiltec:
> When you are logged out, there is no way of filtering all groups on karrot. My proposal would be adding a search bar underneath groupPreview's "Which group do you want to join?" headline for when you are logged out.

seems implemented in master, will close.

PR #892, opened 2018-02-22
Added searchbar with filter

searchingfor:
@D0nPiano @tiltec 

Made this as a starting point concerning #877. Didn't address the geofilter aspect of it yet. Let me know any feedback you have.

Thanks! :)

### DIFF
--- a/src/components/GroupJoin/GroupGalleryUI.vue
+++ b/src/components/GroupJoin/GroupGalleryUI.vue
@@ -45,12 +45,20 @@
     >
       {{ $t('JOINGROUP.WHICHGROUP') }}
     </h4>
+    <!-- <Search
+      class="searchbar"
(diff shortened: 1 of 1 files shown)


D0nPiano:
Looks pretty good! :+1: Two notes:
  * You should probably remove the HTML-comment `<!-- <SEARCH...` and the according import & component
  * Are you willing to fiddle with the design some more? fixing the margins, adding padding to the box, maybe putting it on a card (which also fixes margins) or giving the search a white background?

If you are very motivated and like css stuff maybe you could also tinker with [Vue's List Transitions](https://vuejs.org/v2/guide/transitions.html#List-Transitions). This is definitely optional though - haven't read into it much.

tiltec:
(I updated the snapshots directly in master)

Issue #890, opened 2018-02-20
Add edit button to profile page


djahnie:
So far editing one's personal profile is done via accessing 'settings' in the three dot menu in the upper right (desktop view) or the three bar menu in the upper left (mobile view).

It would be much more intuitive to have an edit button directly on the profile page. The green pencil button, like the one we use in the store view, would be the logical choice.

NaelaTarannum:
We would like to give this a try

djahnie:
Great! Go for it! :smile_cat:

agarwalpooja:
Hello @djahnie, I am working on this issue but I am unable to find the green pencil button.Can you give a picture or the font awesome icon name or the exact place where it is used in store view.

tiltec:
@agarwalpooja Try out [Vue Devtools](https://github.com/vuejs/vue-devtools), it's a browser extension that gives you nice overview and insight. I'm sure you can find the button with it :)

![vuejsdevtool](https://user-images.githubusercontent.com/31616/36394399-6e364c92-15b4-11e8-8901-b0f988a3d973.png)

agarwalpooja:
Thank you.

agarwalpooja:
![screenshot from 2018-02-25 12-50-24](https://user-images.githubusercontent.com/30632125/36639119-f5b6615a-1a2a-11e8-8d6c-b913eaf6598e.png)


I have added an edit button which on click is redirecting to settings page. Is this button icon fine or its alignment, color needs to be changed.

tiltec:
Great that you made it work! I guess we can discuss a bit more on the design, could you open a Pull Request for it?

agarwalpooja:
Yup, sure.

On Feb 25, 2018 3:28 PM, "Tilmann Becker" <notifications@github.com> wrote:

Great that you made it work! I guess we can discuss a bit more on the
design, could you open a Pull Request for it?

—
You are receiving this because you were mentioned.
Reply to this email directly, view it on GitHub
<https://github.com/yunity/karrot-frontend/issues/890#issuecomment-368296998>,
or mute the thread

agarwalpooja:
![screenshot from 2018-02-26 01-40-14](https://user-images.githubusercontent.com/30632125/36642596-af504d92-1a67-11e8-9f1e-18e1d0543eaf.png)


[...]

PR #895, opened 2018-02-25
added edit button in profile page

agarwalpooja:
Refs #890


tiltec:
I think it would be good to use the round green pencil button that we also use on the store page:

image

I'm a bit unsure where to position it, I have the feeling it should be top-right of the user map if we take the store edit page as example. @D0nPiano maybe your input would be helpful :)

agarwalpooja:
Okay. Thank you. I will make the changes accordingly.

tiltec:
Preview pictures are in #890

tiltec:
I don't feel up to make the design decision, what do you think about it @djahnie?

tiltec:
Your changelog is quite big (over 2700 changed lines), can you check what happened there?

agarwalpooja:
Actually I was trying to update the forked repository, as it was showing it
is 58 commits behind the actual repository. While pulling upstream 71 files
changed and then I was not able to resolve this so committed the changes.
Can you help me with updating my repository with the actual repository.

nicksellen:
@agarwalpooja https://help.github.com/articles/syncing-a-fork/ is a guide for how to manage keeping your fork up to date with our upstream repository. There's quite a few concepts to get your head around, so don't worry if it is a bit confusing at first.

agarwalpooja:
Okay. Thank you.I will try this.

agarwalpooja:
Thank you for this github article. It was really helpful.

src/components/User/Profile.vue
```diff
@@ -113,7 +121,10 @@ p.subtitle
   width 200px
   float right
   max-width: 40%
-
+.on-right
+  margin-left 480px
```
D0nPiano: This is problematic for smaller screens. You can try out what it looks like on mobile using your browsers Resposive Design Mode (Firefox: https://developer.mozilla.org/en-US/docs/Tools/Responsive_Design_Mode)
It should probably be something like "float: right" and maybe some margin (10px or so) to it.

agarwalpooja:
Yup previously i used float right, but using float right the icon appears
as in the second image which you liked previously.
I have tried using it at various positions but with float right I dont
think It is possible.I will try again.

tiltec:
image

Mobile layout is a bit broken, you can use the responsive design mode of your browser to see and fix this. For example: https://developer.mozilla.org/en-US/docs/Tools/Responsive_Design_Mode

agarwalpooja:
Okay, Thank You.

D0nPiano:
@tiltec  Haha, was pretty much exactly what I replied earlier. Obviously had the same thought :smile: 

@agarwalpooja Mhm, wired. Just tried it out and:
```
.map
  height 200px
  width 200px
  float right
  max-width: 40%
.on-right
  float right
  margin-right 10px
.inlinee
  display inline
```

works fine for me. Which browser are you using?

agarwalpooja:
Firefox

agarwalpooja:
It's good to know that it worked.

agarwalpooja:
@D0nPiano

.map
  height 200px
  width 200px
  float right
  max-width: 40%
.on-right
  float right
  margin-right 10px
.inlinee
  display inline

You were right it is working.

I have added these changes. Are any more changes required?

agarwalpooja:
Okay :)

tiltec:
Ooops, it seems that we added the button to _all_ user profile pages, not just for the user that is logged in...

src/components/User/Profile.vue
```diff
@@ -39,13 +39,21 @@
...
         </strong>
         <a :href='"mailto:" + user.email'>{{ user.email }}</a>
       </div>
+      <q-btn
+        icon="fa-pencil"
+        small
+        round
+        class="inlinee on-right"
+        color="secondary"
+        @click="$router.push('/settings')"
```
tiltec: I missed this. We prefer named routes instead of referencing the URL directly

src/components/User/Profile.vue
```diff
@@ -39,13 +39,21 @@
...
         <strong>
           <i class="fa fa-fw fa-envelope-o on-left" />
         </strong>
         <a :href='"mailto:" + user.email'>{{ user.email }}</a>
       </div>
+      <q-btn
+        icon="fa-pencil"
+        small
+        round
+        class="inlinee on-right"
```
tiltec: `inlinee` sounds confusing to me

agarwalpooja: Ok I will change this class name.

src/components/User/Profile.vue
agarwalpooja: Okay. I will try to resolve this.

agarwalpooja:
So we have to add this edit button only for  the user who is logged in.I will try to resolve this.

tiltec:
Somebody noticed it in #945 and said they would work on it, so best coordinate with them.

PR #896, opened 2018-02-25
Group preview animaions

D0nPiano:
Was tinkering with transition groups a bit to improve the groupGallery

- [x] Added animation
- [x] Added groupPreview component on GroupGallery page (like the new historyList) and changed routes accordingly
- [x] Smaller improvements to Layout of GroupPreview

### DIFF
--- a/src/__snapshots__/storyshots.spec.js.snap
+++ b/src/__snapshots__/storyshots.spec.js.snap
@@ -658,11 +658,11 @@ exports[`Storyshots GroupGallery signup view 1`] = `
     </div>
     <!---->
     <!---->
-    <h4 class="text-primary">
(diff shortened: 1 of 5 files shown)


D0nPiano:
Just looked through the snapshot changes. They are all groupPreview related. I think they are fine.
Probably it's just so much since it's 15 Preview cards which at least changed in indent (since they are now in `<transition-group>`), each being 25 lines long its about 375 changed lines from the transition group alone.

tiltec:
Ah alright! Yes, it's indeed only the changes to GroupGallery that make the snapshot diff so big :)

tiltec:
Looks great, thanks for your work! Hope the Bachelor thesis is going fine as well :+1:

PR #901, opened 2018-02-26
bugfix for maps

LyraXaph:
This should fix the maps not showing properly in some cases.

### DIFF
--- a/src/components/Map/StandardMap.vue
+++ b/src/components/Map/StandardMap.vue
@@ -146,7 +146,7 @@ export default {
       }
     },
     zoom () {
-      if (!Number.isNaN(this.forceZoom) & this.forceZoom) {
+      if (this.forceZoom) {
         return this.forceZoom
       }
       if (!this.preventZoom && !this.bounds) {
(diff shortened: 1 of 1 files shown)


src/components/Map/StandardMap.vue
```diff
@@ -146,7 +146,7 @@ export default {
       }
     },
     zoom () {
-      if (!Number.isNaN(this.forceZoom)) {
+      if (!Number.isNaN(this.forceZoom) & this.forceZoom) {
```
nicksellen: You probably meant `&&`(see https://stackoverflow.com/a/7310120).

Although `isNaN(null)` returns false, and as `null` is the default value it is not useful here - you can just check if the value is `null` to decide whether to use it or not.

LyraXaph: You are right that it is not useful, I think it used to be a problem before when switching to full screen  view of the map (because upon switching forceZoom has the value NaN), but I couldn't reproduce it now, so I guess I will remove it.

nicksellen: `isNaN` is quite a confusing function really, check the examples on https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/isNaN#Examples

e.g. `isNaN({});` returns true...

PR #903, opened 2018-02-26
GroupGallery Map

D0nPiano:
Added a map to GroupGallery. Might make sense for #877. Looks fairly different on mobile, since there were some trubbles with the footer and z-index.

ToDos:
- [x] Add Markers
- [x] Code Cleanup

map

### DIFF
--- a/src/components/GroupJoin/GroupGalleryCardsUI.vue
+++ b/src/components/GroupJoin/GroupGalleryCardsUI.vue
@@ -41,7 +41,7 @@
     </transition>
     <transition name="slide-toggle">
       <div
-        v-if="expanded && filteredMyGroups.length>0"
+        v-if="(expanded && filteredMyGroups.length>0) || (filteredOtherGroups.length == 0 && filteredMyGroups.length == 1)"
         class="join-groups">
         <h4 class="text-primary">
           {{ $t('JOINGROUP.MY_GROUPS') }}
@@ -89,7 +89,7 @@
       </q-card>
     </transition>
     <transition name="slide-toggle">
-      <div v-if="expanded && filteredOtherGroups.length > 0">
+      <div v-if="(expanded && filteredOtherGroups.length > 0) || (filteredOtherGroups.length == 1 && filteredMyGroups.length == 0)">
         <transition-group
           name="list-complete"
           v-if="showOtherGroups"
--- a/src/components/GroupJoin/GroupGalleryCardsUI.vue
+++ b/src/components/GroupJoin/GroupGalleryCardsUI.vue
@@ -6,7 +6,7 @@
       v-for="group in groups"
       :key="group.id"
       class="list-complete-item inline-block col-xs-12 items-stretch"
-      :class="groups.length == 1 ? '' : 'col-sm-6 col-lg-4'">
+      :class="groups.length == 1 ? '' : $q.platform.is.desktop ? 'col-lg-6 col-xl-4' : 'col-sm-6 col-lg-4'">
       <GroupGalleryCard
         v-if="groups.length != 1"
         :class="{highlight: group.id === currentGroupId}"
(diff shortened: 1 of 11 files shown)


D0nPiano:
More changes than expected... :smile:  Should finally be done for now though!

Looks like this now:
### Desktop
map1
map2

### Mobile
mapmobile

### Possible To Dos (for other PRs)
- [ ] make group map use new "full width" slot in main Layout
- [ ] if user moved the map / zoomed, filter groups accordingly
- [ ] if user moved the map / zoomed, filter groups accordingly
- [ ] But the actual cards into a seperate component, rename GroupCards to GroupCardsLayout
- [ ] Enable searching for locations
- [ ] Find a way to make popups on map more pretty

tiltec:
Wow, great changes! Looks good! A few nitpicks:

image

This could be aligned better, also I would prefer the arrow-up/arrow-down that we use in the sidebar instead of "x/+" icons.

image

Three group cards next to each other look a bit cramped on my laptop, I guess it's just too small. But something that could be improved.

src/components/GroupJoin/GroupGalleryCardsUI.vue
```diff
@@ -0,0 +1,257 @@
...
+            class="underline"
+          >
+            {{ $t('JOINGROUP.LOGOUT_MESSAGE.LOG_IN') }}
+          </router-link>
+        </i18n>
+      </q-alert>
+    </transition>
+    <transition name="slide-toggle">
+      <div
+        v-if="(expanded && filteredMyGroups.length>0) || (filteredOtherGroups.length == 0 && filteredMyGroups.length == 1)"
```
tiltec: The condition is getting quite long, better make it a `computed` value.

src/components/GroupJoin/GroupGalleryCardsUI.vue
```diff
@@ -0,0 +1,257 @@
...
+      <q-card v-if="!previewOpened && !showMyGroups">
+        <q-search
+          @change="$emit('search', arguments[0])"
+          class="searchbar"
+          v-model="search"
+        />
+      </q-card>
+    </transition>
+    <transition name="slide-toggle">
+      <div v-if="(expanded && filteredOtherGroups.length > 0) || (filteredOtherGroups.length == 1 && filteredMyGroups.length == 0)">
```
tiltec: < insert comment about complicated logic in templates here > ;)

D0nPiano: Haha :laughing:

tiltec:
This also needs consideration how to handle inactive groups (per #915 ). I added a quick fix to hide inactive groups from the list in #925, but maybe if someone explicitly searches for a group it should still show it? Inactive groups will anyway get shown if someone searches in the topbar.

No blocking issue, I'd merge this PR without filtering inactive groups.

For the testing, I'm not sure if I would like to merge without tests. Complicated logic tends to break easily on future edits and this can slow down development. I had quite a bad experience when I tried to edit the `PickupUsers` component (the one that calculates the number of slots depending on page width) and after a day of work, I gave up and did only the most minimal change. Might be my own stupidity, but maybe I would have been more successful if I had started with writing tests for it.

D0nPiano:
Wrote a couple of tests now - especially for expanded / not expanded and for clicking on the preview button and the according change (not showing the other cards anymore, creating a GroupPreview component)

I saw the active field in /api/groups - but there are no inactive groups yet, right? Is there a way to make one inactive for testing purposes yet? Would also just wait a bit otherwise then...

Hm, yeah, that does sound annoying with the pickups - what did you want to change?

tiltec:
> I saw the active field in /api/groups - but there are no inactive groups yet, right? Is there a way to make one inactive for testing purposes yet?

True. We would have to mark them in the database, there is no API planned for that. Also I want to rewrite the `active` field into a `status` field to accomodate the playground group too -> #914 
So we rather merge this and add the inactive & playground handling later.

> Hm, yeah, that does sound annoying with the pickups - what did you want to change?

 I fixed a bug where v-for would receive a negative number to iterate over. While doing that, I wanted to move some of the logic from the template into computed properties to make the component easier to debug.

tiltec:
Merging this will break the end-to-end test that runs before deploying to production, just a reminder to myself to adapt the test soon.

Issue #900, opened 2018-02-26
Sort groups by number of members


djahnie:
In the group preview overview the groups are sorted alphabetically:
![image](https://user-images.githubusercontent.com/17573771/36689100-c9f23468-1b2e-11e8-84a4-de541d9e18fb.png)
I think it would make much more sense to sort them from many to few members. Especially since we have so many groups which are not actually used. I'd prefer to have the real groups on the top and not hidden below a whole lot of test groups...

djahnie:
I just realized that the order already _is_ from biggest amount of users to smallest when one is not logged in, but when logged in and part of multiple groups the list looks like the one in the screenshot above. Meaning:
- groups the user is a part of are sorted alphabetically
- groups the user is not a part of are sorted from biggest amount of users to smallest

So it's only about fixing the order of the groups a user is part of. :)

tiltec:
I think the joined groups should continue to be sorted by name, I assume it's irrelevant for most users which of their joined groups has most members.

You're member of many groups, which I assume is a very special use case of the software. I don't have statistics at hand, but I assume that regular users are part of only a few groups (1-3). In this case, I'd expect the default alphabetical sorting, because I'd switch between groups for other reasons than their member count.

Would it help if I extract the average number of groups per user on karrot.world?

djahnie:
Hmm you could be right, @tiltec. I forgot that my use case is quite special. Sorry if I made you do unnecessary edits, @searchingfor.

PR #904, opened 2018-02-27
Now sorts by member count

searchingfor:
For #900


tiltec:
Did you do this for consistency? I think the joined groups should continue to be sorted by name, I assume it's irrelevant for the user which of their joined groups has most members.

PR #906, opened 2018-02-27
Conversation notification settings

tiltec:
Closes https://github.com/yunity/karrot-backend/issues/490
Needs https://github.com/yunity/karrot-backend/pull/500

Adds a notification setting to the group wall.

It supports to trigger the conversation mute via URL parameter for supporting unsubscribe links.
I made it work when the user logs in during the process, as I guess it's a typical use case for people to stop their e-mail subscription when they haven't visited for a long time and their session expired.


tiltec:
image

PR #916, opened 2018-02-28
Fix Search.vue

varsha6319:


tiltec:
You added another search box below the existing search box. That doesn't fix the problem. It's always shown, even when search results exist.
Can you think of other ways to solve the problem?

As a sidenote, can you reference the issue in your first message in the PR? like this:
`Closes #123`

This helps me and other keeping track of the connection between issues and PRs.

tiltec:
-> https://github.com/yunity/karrot-frontend/pull/920

Issue #868, opened 2018-02-07
Mark and remove inactive users from a group


tiltec:
As mentioned in #853, it would be nice to remove inactive users from a group:

> due to inactivity: notify and mark as inactive after 1 month (grey out in member list), notify again after 2 months, remove from group after 3 months

Questions were raised by @djahnie 

- What counts as activity? (read or write access, imho read is enough)
- What is the process and what are the timeframes?

I think activity should be counted _per group_ and read/write access should both be counted as activity. It will therefore be possible to get removed from a group while being active in another one. It will also be possible to stay active by just visiting the group from time to time.

djahnie:
I agree on your definition of activity and that it has to be counted per group.

What are the first steps to implementing something like this? I guess a new field in the database that has the binary choice of being set to 'active' or 'inactive' is needed, as well as some code, that counts the time since the last activity of a user. This is more of a backend task, am I right?

(In case this is not obvious, I'm trying to find more starter tasks...^^)

nicksellen:
This also ties a bit into my ponderings on how to record activity in general (so it's clearer which people/groups are actively using it) - my ponderings led me to think in terms of 3 levels of activity:
- registered user (signed up, but do nothing)
- online user (they log in, click around)
- active user (they do pickups!)

Maybe it's also useful to distinguish between users that are active online (writing messages, editing things, etc) and passive ones (just click around a bit), but maybe not...

Historic data is better stored in influxdb I think and exported periodically, but the application also would/might/should have some historic data too, where it makes sense for the logic of the application (I think the application should not read from influxdb).

Perhaps a storage format for inside the application would be an activity table and record into time buckets per user+group, e.g. `recordActivity(user_id, group_id, time_bucket, activity_type)` where _time_bucket_ might be a day , and _activity_type_ might be `online|offline`. the data could be sent to influxdb db at the same time, then the application can delete from the activity table when it makes sense for the application logic (e.g. you only actually need to know the last activity).

So, actually, rethinking as I write, if we need to only know the last activity then could store that directly in the user/group join table (`last_online_activity_at`, `last_offline_activity_at`), and write to influxdb as before.....

... _or_ even just repurpose the history table, and check when their last activity was (although that does not record things like logging in, but could (and just filter it from the normal view), but could end up with too much data in there that we don't really need.

The activity table could have counters for each time bucket row too if we wanted to keep track of how much activity too, not just whether there was some.

Seems enough thoughts for now.

[...]

PR #919, opened 2018-02-28
Mark user as active in group when they visit a group-related page

tiltec:
Related to #868

### DIFF
--- a/src/services/api/groups.js
+++ b/src/services/api/groups.js
@@ -46,4 +46,8 @@ export default {
   async conversation (groupId) {
     return convertConversation((await axios.get(`/api/groups/${groupId}/conversation/`)).data)
   },
+
+  async markUserActive (groupId) {
+    return axios.post(`/api/groups/${groupId}/mark_user_active/`)
+  },
 }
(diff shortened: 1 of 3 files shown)


djembejohn:
OK, I did a test of this. It works perfectly when I click on all the side bar options on the group page

The only issue I found was that when you add messages on the wall it doesn't register the user as being active. I suppose a genuine user could conceivably sit with a group page open and just chat on the wall?

I also get a Server Error when I add a message on the wall, which might be causing the problem: 

backend_1   | Internal Server Error: /api/messages/

backend_1   | Traceback (most recent call last):

backend_1   |   File "/app/foodsaving/utils/email_utils.py", line 270, in prepare_email_content

backend_1   |     text_template = get_template('{}.text.jinja2'.format(template))

backend_1   |   File "/app/env/lib/python3.5/site-packages/django/template/loader.py", line 19, in get_template

backend_1   |     raise TemplateDoesNotExist(template_name, chain=chain)

backend_1   | django.template.exceptions.TemplateDoesNotExist: conversation_message_notification.text.jinja2

nicksellen:
I think for this coarseness of activity, changing pages should be sufficient.

Issue #864, opened 2018-02-04
Make search show info box if no results were found


tiltec:
The topbar search doesn't show any information when no results were found. It would be nicer to show a message "No results found.".

![image](https://user-images.githubusercontent.com/4410802/35778455-c6cd4f56-09be-11e8-8dc2-1e9a8c7ed8e5.png)

Refs #817

kallepallimounikasmitha:
I want to work on this.

tiltec:
Great!

tiltec:
@kallepallimounikasmitha How is your work going?

varsha6319:
I have solved this issue, can I send the pull request

tiltec:
@varsha6319 Yes please :)

PR #920, opened 2018-03-01
fix-search.vue

NaelaTarannum:


NaelaTarannum:
solves issue #864

src/components/General/Search.vue
```diff
@@ -13,6 +13,7 @@
         @selected="selected"
       />
     </q-search>
+    <span v-if="!results.length" >No results.</span>
```
nicksellen: All the text needs to be internationalized so we can manage the translations.

Have a look at the other files for examples of how to do this, e.g. from CurrentUser.vue there is:

```vue
<span>{{ $t('PICKUPLIST.ITEM.LEAVE') }}</span>
```

src/components/Group/Navbar.vue
```diff
@@ -9,6 +9,20 @@
       class="col"
       ref="groupnav"
     >
+      <div class="">
```
nicksellen: Can remove attribute if it is empty

src/components/Group/Navbar.vue
```diff
@@ -9,6 +9,20 @@
...
+        <q-transition
+          duration="510"
+          name="turn-in"
+          appear
+        >
+          <div class="photoCard shadow-2">
+            <ProfilePicture
+              :group="group"
```
tiltec: The changes in Navbar.vue seem unrelated to the referenced issue. What was your intention here?
Also,`ProfilePicture` doesn't have a `group` prop, but needs a `user` prop, so this code seems unlikely to work.

src/components/General/Search.vue
NaelaTarannum: Will try

tiltec:
Closing this in favor of #965

Issue #831, opened 2018-01-05
Display a foodsharing logo somewhere?


djahnie:
I realized that many of the people, who contact us come via foodsharing.de. They expect something foodsharing.de-related and would feel way more at ease if they could clearly see that there's a close connection. Even if I keep telling them that this connection exists, we never display it openly.

Through this g0v thing I thought about the possibility to similarly display a foodsharing logo in the footor of karrot. Just the fork. And with some kind of explanatory writing like 'Inspired by' or 'Bringing X worldwide'. Something like that.

This would probably also ease the confusion we could come across when recruiting new people for foodsharing in Wurzen but then leading them to a different page. Here a more obvious connection between the two projects would be really beneficial. :)

tiltec:
This is primarily about karrot.world I guess? Not the software itself, which would affect all people who (will) self-host. I guess we could add a customization option #676 where people can change footer text.

djahnie:
Yes true, karrot.world is what I'm talking about and not the software itself.
If there was an easy to use customization option I'd totally do it myself! ^^

tiltec:
Hm, it probably will be some server-side setting. Although we could think of using the Django admin interface for site admin tasks. For this we would need to make some users into superusers that have read-write access to all the data..

D0nPiano:
I think it's a very good idea to include the logo somewhere.

I would suggest adding a file in `src/themes/` for karrot.world (like `app.karrot.styl`) and add it via css / app variables. The file could be included/excluded in the build process then as far as I understood how the theming works.

tiltec:
> I would suggest adding a file in src/themes/ for karrot.world (like app.karrot.styl) and add it via css / app variables. The file could be included/excluded in the build process then as far as I understood how the theming works.

This seems a bad idea to me because then other sites will have to maintain a fork of our git repo with their custom images in it and run their own build process.

I think the workflow should rather be like this: get bundle from https://github.com/yunity/karrot-frontend/releases, copy to server, done!
The customizations should either happen in the backend and served via API, or be defined in a file `site.json` that gets loaded from the server (similar to `about.json`). This makes it much easier to ship code updates.

djahnie:
So... can we decide on something and then maybe even hand this task to a Rails Girl..? :wink:

tiltec:
I would propose the `site.json` file. It could look like this:

```
{ "siteLogo": "https://karrot.world/site/logo.png" }
```

[...]

PR #929, opened 2018-03-02
Add link to info page

djahnie:
- in footer of desktop version
- in sidenav of mobile version
- remove the 'general' header in sidenav
- make sublabels in sidenav translatable

### DIFF
--- a/src/locales/locale-en.json
+++ b/src/locales/locale-en.json
@@ -1,6 +1,7 @@
 {
   "GLOBAL": {
     "GITHUB_NOTE": "Development",
+    "FSWW_NOTE": "Read our info page!",
     "FACEBOOK_NOTE": "Join our Facebook Group!",
     "WARNING": "Warning",
     "BETA_WARNING": "This website is under active development and everything is subject to change. We can't guarantee for the safety and privacy of the data you enter (but we try our best).",
--- a/src/locales/locale-en.json
+++ b/src/locales/locale-en.json
@@ -157,7 +157,7 @@
   },
   "SIDENAV": {
     "INFO_SUB": "How to build your foodsaving group",
-    "BLOG_SUB": "All News around karrot.world",
+    "BLOG_SUB": "All news around karrot.world",
     "GIT_SUB": "See our code!"
   },
   "LOGIN": {
--- a/src/locales/locale-en.json
+++ b/src/locales/locale-en.json
@@ -155,6 +155,11 @@
     "USERPROFILE": "Visit your profile",
     "CHANGE_GROUP": "Change group"
   },
+  "SIDENAV": {
+    "INFO_SUB": "How to build your foodsaving group",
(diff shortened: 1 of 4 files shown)


src/locales/locale-en.json
```diff
@@ -154,6 +155,11 @@
     "USERPROFILE": "Visit your profile",
     "CHANGE_GROUP": "Change group"
   },
+  "SIDENAV": {
+    "INFO_SUB": "How to build your foodsaving group",
```
tiltec: Does it make sense to have those translatable? The infopage is only available in English.

src/locales/locale-en.json
```diff
@@ -154,6 +155,11 @@
...
+    "BLOG_SUB": "All News around karrot.world",
```
tiltec: should be "news" not "News"?

src/locales/locale-en.json
djahnie: I still think it makes sense to translate the text inside Karrot. And if only for consistency...

src/locales/locale-en.json
djahnie: Ahaha that's what I was thinking! But in the older version from you it was written like that so I simply kept it that way...^^

src/locales/locale-en.json
```diff
@@ -154,6 +155,11 @@
     "USERPROFILE": "Visit your profile",
     "CHANGE_GROUP": "Change group"
   },
+  "SIDENAV": {
+    "INFO_SUB": "How to build your foodsaving group",
+    "BLOG_SUB": "All news around karrot.world",
```
tiltec: Seems a bit weird to have references to `karrot.world` inside our code base, especially when we want to make karrot more customizable. It was there before, but now it will also get translated...

djahnie: True. I'll change it.

tiltec: At some point, we need to tackle the issue of site customizations. My last proposal in https://github.com/yunity/karrot-frontend/issues/831#issuecomment-363807645 didn't receive much feedback, but it's clearly not ideal (for example it's not easily translatable).
Or maybe it's not an issue right now and can be tackled in a more distant future. Then we would just add the things to our repo, even though they are specific to karrot.world.

src/locales/locale-en.json
```diff
@@ -154,6 +155,11 @@
     "USERPROFILE": "Visit your profile",
     "CHANGE_GROUP": "Change group"
   },
+  "SIDENAV": {
+    "INFO_SUB": "How to build your foodsaving group",
+    "BLOG_SUB": "All news around the karrot software",
```
tiltec: Maybe just "News about the karrot"?

djahnie: Sounds weird to me. And unclear.

PR #930, opened 2018-03-02
Show hint if the user can't receive notification emails

### LINKED ISSUE #921: Send notifications only to verified addresses
Lots of invalid email adresses might not bounce, but end up unwanted in someone elses mailbox. I think we should only send out notifications if the users e-mail address is verified.

tiltec:
...when their e-mail address is not verified.

Closes #921


D0nPiano:
The button also needs the `hoverScale` class to get bigger when hovered. (Should also be added to the "bell" notification button). Rest looks good :+1:

tiltec:
Ah, I didn't know about `hoverScale`. Maybe it's a good idea to make our own `KActionButton` that has all styles included?

Issue #924, opened 2018-03-01
Enable group summary emails


nicksellen:
Group summary emails (#847)are implemented but not sent out yet. They probably need a button to enable/disable them, a few options:

1. reuse the wall notifications option (but would not allow user to have summary emails, but not wall notifications)
2. introduce an option in the user settings menu for global summaries (but would not allow user to subscribe to for summary emails from some groups only)
3. introduce a group notifications page where you can configure all notifications about a group (wall messages, and summary emails, and in the future pickup notifications)
4. introduce a global notifications page where you can set everything in 3 but for all groups on one page

I would be most inclined towards 3 (and therefore add another field in the db (or turn the existing one into an array)).

tiltec:
Seems good!

How about another section in the group settings page, for example "Your personal group settings"? It's quite similar to 3, but would avoid introducing another page. I don't think it would be too overloaded, but that's subjective.

You will need another field in the db, as the current one is per conversation, not per group.

Slightly related: I think it would be helpful if the e-mail would contain an unsubscribe link that doesn't need login. I was often annoyed by other websites that forced my to recover my password in order to unsubscribe. As that's also related to conversation notifications, I should probably open another issue...

nicksellen:
> group settings page

tiltec:
Agree, better create a new page then.

PR #933, opened 2018-03-03
Add group settings page for notification options

nicksellen:
Addresses issue in https://github.com/yunity/karrot-frontend/issues/924


nicksellen:
groupsettings

nicksellen:
Ah, it should also have a nice thing saying you need to verify email address to actually get them too... in general I think that could be more prominent.

PR #934, opened 2018-03-04
Use reworked auth API

mr-kojo:
**Steps:**
- [x] E-mail verification
- [x] E-mail change
- [x] Password change
- [x] Password reset
- [x] Account deletion


mr-kojo:
ToDos for password-change:

- [ ] Show an alert on success
- [ ] Clear fields on success 

ToDos for email-verification:

- [ ] Test verification if not logged in

nicksellen:
@mr-kojo do you have time to continue this, or should I assign it to me? I think it's quite close now, looks good!

mr-kojo:
Sorry for the delay, was busy with other things. Didn't forget it, though :) I'll continue to work in this this weekend.

src/components/Settings/ChangePassword.vue
```diff
@@ -59,14 +59,13 @@ export default {
     }
   },
   methods: {
-    save () {
+    async save () {
```
mr-kojo: This new implementation of the `save` method no longer triggers an event for the parent component to call the backend. The reason is that this doesn't work in combination with `async`/`await`. 
As a result, the `changePassword` method of the `Settings` component (the parent component) is no longer used.
Do you consider this bad style? Is there a nicer way of clearing the password fields?

src/components/Settings/ChangePassword.vue
```diff
@@ -59,8 +61,16 @@ export default {
   methods: {
     save () {
       const { oldPassword, newPassword } = this
-      this.$emit('save', { oldPassword, newPassword })
```
mr-kojo: This new implementation of the `save` method no longer triggers an event for the parent component to call the backend. The reason is that this doesn't work in combination with a Promise.
As a result, the `changePassword` method of the `Settings` component (the parent component) is no longer used.
Do you consider this bad style? Is there a nicer way of clearing the password fields?

nicksellen: It's not the style we've used so far, which would be to set some state (which might be status), but I don't feel like we've quite got all the common use cases covered - it seems very long winded for cases like this, and not really useful. It might be there is a reasonable similar approach that has that logic in a connect component, I don't think we've explored all the different options that those take.

Although maybe there is not time now to investigate further approaches, and that this is fine for now.

One thing though, it should use the async/await syntax which is equivelent to using the promises explicitly. For that you need to declare the whole `save()` function async (`async save()`) then just await on `this.saveUser()` (if there is an error it should throw, so should not need to check the result.

src/store/modules/auth.js
```diff
@@ -93,7 +93,10 @@ export default {
 
       async changePassword ({ commit, state, dispatch }, data) {
         await auth.changePassword(data)
-        dispatch('logout')
+        dispatch('alerts/create', {
+          type: 'changePasswordSuccess',
+        }, { root: true })
+        return true
```
nicksellen: .... so can get rid of this `return true`

src/components/Settings/ChangePassword.vue
nicksellen: The vuejs guy points out you can emit a callback in the event too... https://github.com/vuejs/vue/issues/5443#issuecomment-294147285

tiltec: We already pass actions as props in some other places (e.g. https://github.com/yunity/karrot-frontend/blob/f7377683baa1a4cbc1a328278f691f6e271c1499/src/components/Conversation/Conversation.vue#L120 ), so it seems fine to do the same here. Emitting a callback could be easier to understand though.

tiltec: Agree with switching `.then` to `await`, we almost always use `async`/`await` in our codebase (could refactor the rest at some point...)

mr-kojo: Aah, emitting a callback - that's clever. 
I implemented it this way now (6fd91c3).

src/pages/Settings.vue
```diff
@@ -59,15 +59,14 @@ export default {
...
-        clearPasswordFields()
+    async changePassword (oldPassword, newPassword, done) {
+      await this.$store.dispatch('auth/changePassword', { oldPassword, newPassword })
+      if (!this.changePasswordStatus.hasValidationErrors) {
+        done()
       }
     },
     ...mapActions({
       saveUser: 'auth/save',
-      changePassword_: 'auth/changePassword',
```
mr-kojo: I replaced this line by the explicit call
`this.$store.dispatch('auth/changePassword', { oldPassword, newPassword })`
for the sake of better readability.

src/store/modules/users.js
```diff
@@ -10,6 +10,7 @@ function initialState () {
     entries: {},
     idList: [],
     activeUserId: null,
+    resetPasswordSuccess: false,
```
mr-kojo: Unrelated change

src/components/Login/PasswordReset.vue
```diff
@@ -26,7 +26,7 @@
           class="error"
```
mr-kojo: Unrelated change

src/components/Login/PasswordReset.vue
```diff
@@ -26,7 +26,7 @@
...
         >
           <i class="fa fa-exclamation-triangle"/>
-          {{ $t('PASSWORD.RESET.CODE_ERROR') }}
+          {{ $t('GLOBAL.INVALID_LINK') }}
         </div>
```
mr-kojo: Unrelated change

src/locales/locale-en.json
```diff
@@ -272,8 +277,7 @@
     "RESET": {
       "TITLE": "Reset password",
       "OK": "Save",
-      "SUCCESS": "Your new password has been saved.",
-      "CODE_ERROR": "Sorry, the link you've opened is invalid."
+      "SUCCESS": "Your new password has been saved."
     },
```
mr-kojo: Unrelated change

mr-kojo:
On my local machine, `yarn test` complains:
```code
[vuex] unknown getter: users/requestDeleteAccountStatus
```

What can I do about this?

tiltec:
That happens if the store is not properly mocked in tests. I guess you added the `users/requestDeleteAccountStatus` getter to one of the components that is tested by Storyshots (`*.story.js` files). We usually don't include vuex in storyshots, therefore you need to either mock it or load the store implementation. Have a look at the other `.story.js` files, they have some store mocks and partial store implementations.

src/components/Settings/ChangePassword.vue
```diff
@@ -1,25 +1,25 @@
 <template>
```
mr-kojo: Unrelated

mr-kojo:
Should be ready to merge.
Some client-side form validation can be added later on (see #937).

tiltec:
image

The sticky alert is a bit intense. Maybe it needs a different solution....

tiltec:
I'd merge it without the changes above, it can be fixed later in `master`

mr-kojo:
> The sticky alert is a bit intense. Maybe it needs a different solution....

That's right. I've already documented it here: #936

PR #951, opened 2018-03-07
Edit button removed from user profile who is not logged in.

chocos10:


src/components/User/Profile.vue
```diff
@@ -55,6 +55,7 @@
         round
         class="inlinee on-right"
         color="secondary"
+        v-if="user.id==currentUserId"
```
tiltec: Each `user` object already contains the information if it's the currently logged in user: https://github.com/yunity/karrot-frontend/blob/master/src/store/modules/users.js#L29

No need to access the store directly from this component. It makes it easier to test when all data is getting passed in via props.

(By the way, you almost never want to use the `==` operator in JavaScript, instead use `===`: https://stackoverflow.com/questions/1995113/strangest-language-feature/1998224#1998224 )

chocos10: Changes updated :+1:

Issue #914, opened 2018-02-28
Add a playground group in the group list


djahnie:
Right now the group list is the landing page. Probably a lot of people come to karrot.world and just want to try it out. It would be great if they didn't have to create an account, or if they wouldn't randomly join open groups. My proposal is the following:

- [x] Take one of the unused groups, rename it to 'Playground' and write a login in the public description.
- [x] Display the playground group as the very first in the group list.
- [ ] Add a yellow frame around it to make it even more salient.

Any resistance to me tackling this?

D0nPiano:
I really like the idea! A couple of questions though.

- How are we keeping the playground user from entering other groups?
- Is renaming the group sufficient or should it be translatable?
- Are we keeping people from changing the password for the demo group?

Therefore, I am unsure if it wouldn't be easier to just deploy a demo version of karrot to demo.karrot.world and add a demo user (or multiple ones) there, since the demo user cannot mess with the production database then. The database could also be copied and replaced to the original then from time to time. If you feel confident to prevent the demo user from changing non-demo data though, I am happy either way.

djahnie:
Oh, very good points you have there, @D0nPiano!
- I didn't think of that, you're right. But actually, users who just wan't to play around will probably not make the effort of joining a whole lot of groups, I guess. If they wanted to, they could always make an account for themselves and just hop into any unprotected group they want to, so this is not really a big deal for me... (Also, as soon as we have the autoremoval of inactive users (#868) this won't be grave anyways...)
- Having it translatable would be fancy, for sure! But how would this work? Normally group names are not translatable, so how can we just make this one group name different?
- Sounds reasonable, yes. Okay, looks like this group wull have unique features in more than one way...

I'd love to have the playground directly in production, so that the demo experience is closest to what actual users have. That's also why I'm more in favor of this solution than linking to dev.karrot.world and giving out foo for testing purposes. I think it could confuse some users...

D0nPiano:
Yeah, makes sense. I just usually think I can't destroy anything when I am on a demo page, wherefore I just click around randomly. Especially because you are anonymous as a demo user and don't even have to provide an email address for signing in.

I think having it translatable is just as complicated as adding a yellow frame. The playground group should probably be names "Playground", and in vue.js the store could then filter out the group and do something like
```
if (group.name === 'Playground') {
   group.name === $t('PLAYGROUND.TITLE')
}
```

[...]

PR #960, opened 2018-03-11
Add playground group frontend

tiltec:
Early stages of #914

image

image

image


tiltec:
- disable group name and password editing for playground group (frontend only for now)
- add "group.isPlayground" property
- make playground group name translatable
- show hint when user is in playground group
- make playground group card green
- add a checkbox to signup form to automatically join playground group

The playground group is shown in the group list
- in "my groups" if the user is part of the playground group
- at the top if the user is not logged in
- at the top if the user is logged in, but not part of any group
- at the bottom if the user is logged in and part of any other group than the playground group
- if the search term matches the playground group name (the translated name)

image

Design changes
- increase minimum size of GroupGalleryCard
- slightly change behaviour of group search

Refactoring:
- remove duplicate state ("previewOpened", "search")
- move GroupPreview into GroupGalleryCardsLayout (from GroupGalleryCards)
- use "group.isMember" instead of passing "currentGroupId"
- move component and prop declarations to the top
- fix stories and tests

tiltec:
The group list and map now hides inactive groups by default.

I'm now working on the backend side, disabling notifications for the playground group and preventing password changes via API.

@djahnie and me think that it's nice to have the playground group name and public description translatable. This PR does the translation in the frontend (just the name for now), which hides the _real_ group name that comes from the API. This seems a bit unexpected for developers who use the API.
Therefore, I'm tempted to move the translation into the backend. Possibly the serializer is a good place for it.

Issue #820, opened 2017-12-31
Upgrade to fontawesome 5


tiltec:
[Quasar Extras](https://github.com/quasarframework/quasar-extras#readme) was upgraded to Fontawesome 5, which is a complete overhaul of FA4 which we currently use.

Some icons were renamed: https://fontawesome.com/how-to-use/upgrading-from-4
I think also the style changed a bit.

Do we want to do this upgrade?

Alternatively, we could install FA4 as additional dependency and not use the one bundled with Quasar Extras.

tiltec:
I think we should try to do the upgrade to FA5 and see how it looks.

Steps:
- (optional) add the "[version 4 shim](https://fontawesome.com/how-to-use/upgrading-from-4)" and submit a pull request containing a few representative screenshots to give other contributors the chance for feedback. Probably need to temporarily disable the `import 'quasar-extras/fontawesome'` line in `main.js`.
- upgrade the package `quasar-extras` to the newest version
- rename usages of font-awesome (new prefixes, sometimes changed name): https://fontawesome.com/how-to-use/upgrading-from-4

Anjalizi:
I would like to take up this task.

agarwalpooja:
hey.. @tiltec I am working in this issue. I have downloaded the files for font awesome 5 and have given its link in index.html.And i have changed the prefix and icon name as per FA5.
But now the icons are removed from the page.

anishaagrawal:
hey.. @tiltec i am working on this issue. I have downloaded the files for font awesome 5. And i have given the link for font awesome files in index.html .I havereplaced the old prefixes and icons name with the new one.
But now the icons are removed from the page.!!!! can u help me to resolve this.

tiltec:
Hi @anishaagrawal, can you open a `WIP` pull request containing your changes? Then I can have a closer look!

anishaagrawal:
i have pulled the files changed for upgrading to font awesome 5.... please check it.

ananya-agrawal:
since there has been inactivity in this issue for some time, I want to give it a shot

Issue #913, opened 2018-02-28
Upgrade to quasar 0.15


tiltec:
Lots of changes: https://github.com/quasarframework/quasar/releases
Upgrade and check if the page still looks the same and things work. Maybe a good chance to add more end-to-end tests?

tiltec:
Let me quote here some parts of the release notes.

> The Quasar CLI has been rewritten from scratch to allow for a stellar development experience (Mobile App developers and Electron will fall in love with it!). Only one starter kit is required in order to handle websites, PWAs, Mobile Apps and Electron Apps.

> Take some time to read all "Guide" pages once again. It will help you understand the true power of Quasar v0.15 and what you can do with it.
http://quasar-framework.org/guide/

---

> Observe the new project structure. Start to port out files to the new project folder, taking into account the far superior structure. Using the new starter kit will allow you to take advantage of future seamless upgrades! In any case, do not simply copy your /src folder over to the new starter kit.

Quasar 0.15 comes with a new directory structure, which they recommend to use. It doesn't seem very different to our existing one: http://quasar-framework.org/guide/app-directory-structure.html
The main different I can spot is then introduction of a separate `src-pwa` and `src-cordova` folder. But there seems to be more:

> You'll notice the new starter kit doesn't provide a /build or /config folders. They are no longer required. Everything can be easily configured from /quasar.conf.js now. You don't need to know Webpack.

I fear that this is less powerful than our existing approach where we configure webpack directly and more like the `create-react-app` approach where you need to _eject_ in order to have full power.

---

> No main.js? Yes. It's no longer there because you don't need it anymore. For initialization code and importing libraries into your website/app, read about App Plugins.

> In earlier Quasar versions, to run code before the root Vue instance was instantiated, you could alter the /src/main.js file and add any code you needed to execute.
> There is one problem with this approach. With a growing project size your main.js file most likely would get very cluttered and hard to maintain, which breaks with Quasars concept of encouraging developers to write maintainable and elegant cross-platform applications.
> An app plugin is a simple JavaScript file which needs to export a function. Quasar will then call the exported function when it boots the application.

I guess that makes the project less of a _Vue_ project and more of a _Quasar_ project with every custom functionality bundled into plugins. Might be good because it's more separated.

--- 

Other relevant changes:

- QLayout had major changes: 
- Alert and Toast have been merged into new Notify plugin
- QSideLink is gone
[...]

PR #961, opened 2018-03-12
Upgrade font awesome to v5 

ananya-agrawal:
issue no: #820

### DIFF
--- a/package.json
+++ b/package.json
@@ -52,7 +52,7 @@
     "markdown-it": "^8.4.0",
     "markdown-it-emoji": "^1.4.0",
     "markdown-it-link-attributes": "^2.0.0",
-    "quasar-extras": "0.0.8",
+    "quasar-extras": "1.0.2",
     "quasar-framework": "^0.15.8",
     "raven-js": "^3.20.1",
     "reconnecting-websocket": "^3.2.1",
--- a/src/__snapshots__/storyshots.spec.js.snap
+++ b/src/__snapshots__/storyshots.spec.js.snap
@@ -183,7 +183,7 @@ exports[`Storyshots GroupEdit create 1`] = `
                     </div>
                 </div>
                 <div class="q-field row no-wrap items-start q-field-responsive">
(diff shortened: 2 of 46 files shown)


nicksellen:
Looks good!

You can fix the failing snapshot tests by running `yarn test -u` and then add/commit/push the resulting snapshot files.

Also looks like you might need to rebase on top of the latest master, and resolve the conflict it mentions.

A few little details that seem to be broken (looking at the [netlify deploy preview](https://deploy-preview-961--karrot.netlify.com/#/group/4/wall) and comparing to https://dev.karrot.world/):

Some of the circular buttons have slightly misaligned icons now (they also all seem a little bigger?):

broken1

broken3

The map pins are missing their icons, and when the store markers are enabled the green check mark is missing.

broken2

tiltec:
The misaligned icons seem weird, so far Quasar took care of aligning them properly. Maybe there's an update in Quasar 0.15 that handles this? #913

ananya-agrawal:
whether the alignment should be done by adjusting the height and the width of element ?

nicksellen:
It would be to nice to have them aligned, but the implementation could be temporary solution until the quasar 0.15 upgrade is done.

I didn't look at the existing CSS, but `position: relative` allows you to define an offset relative to it's position (in combination with top/right/bottom/left attributes), so might be a solution with that.

tiltec:
I came quite far with upgrading to Quasar 0.15 in the last days, so you could merge https://github.com/yunity/karrot-frontend/pull/974 into your branch and see if the icons are aligned nicely by default. Then we don't need lots of custom inline styles.

Note this section in the Quasar upgrade guide regarding font awesome icons: http://quasar-framework.org/guide/quasar-upgrade-guide.html#Icon-Packs

ananya-agrawal:
please review it once

ananya-agrawal:
@tiltec please review

ananya-agrawal:
@tiltec do review, done required changes.

PR #965, opened 2018-03-13
Fix search box

chocos10:
fixes issue #864

### DIFF
--- a/src/locales/locale-en.json
+++ b/src/locales/locale-en.json
@@ -6,7 +6,8 @@
     "WARNING": "Warning",
     "BETA_WARNING": "This website is under active development and everything is subject to change. We can't guarantee for the safety and privacy of the data you enter (but we try our best).",
     "CONNECTION_INTERRUPTED": "Error: connection interrupted!",
-    "SERVER_ERROR": "Server error"
+    "SERVER_ERROR": "Server error",
+    "SEARCH_NOT_FOUND": "Not Found"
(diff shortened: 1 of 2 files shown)


src/locales/locale-en.json
```diff
@@ -6,7 +6,8 @@
     "WARNING": "Warning",
     "BETA_WARNING": "This website is under active development and everything is subject to change. We can't guarantee for the safety and privacy of the data you enter (but we try our best).",
     "CONNECTION_INTERRUPTED": "Error: connection interrupted!",
-    "SERVER_ERROR": "Server error"
+    "SERVER_ERROR": "Server error",
+    "GLOBAL.SEARCH_NOT_FOUND": "Not Found"
```
tiltec: `GLOBAL` is duplicate here, since it's already its namespace. You should just name it  `SEARCH_NOT_FOUND`, then use it in the code with `GLOBAL.SEARCH_NOT_FOUND`.

Maybe `NOT_FOUND` instead of `SEARCH_NOT_FOUND` is also better, since it's a pretty generic expression.

tiltec:
It feels a bit unexpected that the "Not found" label can be clicked on and the search closes completely after the user does so. Maybe you find a way to disable this, but it's not a blocker. I would leave it as future improvement then.

chocos10:
@tiltec  i hope this solves the issue :) please review it

PR #974, opened 2018-03-19
Upgrade to Quasar 0.15 (without new CLI)

### LINKED ISSUE #913: Upgrade to quasar 0.15

tiltec:
Closes #913 

I upgraded Quasar to the latest version. It needed many changes, but it works and looks almost the same as before :smile: 

I didn't upgrade to the new CLI and the new project structure yet, as this requires moving and rewriting lots of files. But we might do it in future.

I didn't set up Quasar i18n, because we don't use any component that needs it: http://quasar-framework.org/components/internationalization.html


nicksellen:
If we still need an eventbus you can make one with:

```js
const Events = new Vue()
```

src/store/plugins/presenceReporter.js
```diff
@@ -13,7 +13,7 @@ export default store => {
     .forEach(event => document.addEventListener(event, resetTimer))
 
   // listen on tab & window visibility, mostly works on desktop browser
-  Events.$on('app:visibility', state => {
+  /* Events.$on('app:visibility', state => {
```
tiltec: Could be replaced with http://quasar-framework.org/components/app-visibility.html

tiltec: ... but need to check if it also works without the quasar plugin

tiltec:
I worked a bit more on it, solved the most obvious warnings and switched to the new QLayout component.

A few selected issues that I noticed:

- Toasts need to be replaced with notify
- Dialogs don't open
- AvailablePickups on group wall don't fold out
- Pickups don't fold out on pickup manage page
- Lots of styling issues everywhere, especially where we have custom styling on top of quasar components

I'll have a look at the Dialog component next, that's one of those new plugins that are supposed to be activated in the new `quasar.conf.js` file - which we would only get with the new Quasar CLI and the new directory structure. Maybe there's a way around it though.

tiltec:
Functionality should be mostly done, design needs fixing.

tiltec:
Ready for review! Have a look at https://deploy-preview-974--karrot.netlify.com/ and find the remaining problems :)

nicksellen:
Great work!

newquasarui1

Two things:
1. content of the search box (the magnifying glass and the text "search" is aligned a bit low)
2. the close icon has fallen down a bit

tiltec:
If there's no resistance until then, I'll merge tomorrow afternoon.

PR #977, opened 2018-03-20
redirects to user profile after submitting changes on the settings page

### LINKED ISSUE #968: Redirect to personal profile after changing it
When submitting changes on one's personal profile right now the user stays on the personal settings page. It would be much more intuitive to get redirected to the profile to see the changes one made.

neuroleptic:
Resolves #968 

Redirects user to their personal profile page after they save changes on their personal settings page.

### DIFF
--- a/src/store/modules/auth.js
+++ b/src/store/modules/auth.js
@@ -100,6 +100,7 @@ export default {
       async changeEmail ({ commit, dispatch }, email) {
         const savedUser = await authUser.save({ email })
         commit('setUser', { user: savedUser })
+        router.push({ name: 'user', params: { userId: savedUser.id } })
       },
     }),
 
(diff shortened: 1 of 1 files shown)


tiltec:
Seems good from my POV. Feedback from @mr-kojo might be useful, as you are working on related topics.

tiltec:
As there was no further feedback, I'll merge it and then see what people say :)

mr-kojo:
@tiltec Sorry, I didn't notice your feedback request because I unintentionally disabled Github notifications a couple of days ago 🙈 

Actually, I'm not so sure if users want to be redirected to the profile page. They might instead want to do some other changes. E.g., if I've signed up to karrot and go to the Settings page, I might first upload a profile pic. When I click on _Save Changes_, I'm confused because I'm redirected to the profile page although I would have wanted to enter a phone number, too.
So in fact I wouldn't do any redirect on the settings page...

Issue #988, opened 2018-03-29
Loading button animation no longer working


mr-kojo:
I noted that the quasar buttons don't show this `loader` animation anymore (which is triggered by the `statusMixin`). 

@tiltec, might this be the result of the quasar update?

tiltec:
Indeed, the API changed: http://quasar-framework.org/components/button.html#Button-with-Progress
I can update this, but it might be a nice starter task too :)

PR #990, opened 2018-03-29
loader animation updated

chocos10:
Fixes this #988

### DIFF
--- a/docs/forms.md
+++ b/docs/forms.md
@@ -167,6 +167,6 @@ e.g. the status about a `save` action would be available as `saveStatus` or the 
     - uses `hasAnyError` and `anyFirstError` to show any server-side validation error?
     - uses `hasError(field)` & `firstError(field)` methods to check for field-specific server-side validation errors?
     - uses `hasNonFieldError` & `firstNonFieldError` to show errors unrelated to fields?
-    - sets `:loader="isPending"` on submit button?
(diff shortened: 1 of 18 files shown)


mr-kojo:
When you're done, please run `yarn test -u` and commit the updated UI snapshots so that the tests pass.

tiltec:
Do we still need to set `:value`  on each button?

Quasar documentation is a bit confusing to me. http://quasar-framework.org/components/button.html#Button-with-Progress mentions the method with `:loading`, http://quasar-framework.org/components/button.html#Controlling-the-Button-for-Form-Submission mentions another method with `v-model`.

chocos10:
@tiltec  even i was confused so i didn't removed that .

chocos10:
@tiltec  What should i do remove :value or let it be for now ?

tiltec:
@chocos10 What happens if you remove it?

chocos10:
@tiltec for now it does not disturb anything on the following page but i don't know if the :value is used somewhere else in the codebase.

chocos10:
@tiltec  for now i have removed :value as i didn't find any problem after removing that.

tiltec:
Cool! Can you update the snapshots again? Then it should be ready to merge!

chocos10:
@tiltec done :+1:

Issue #937, opened 2018-03-04
Dynamically disable/enable submit buttons 


mr-kojo:
Buttons used for form submission should be disabled programmatically
- when client-side validation fails, and
- while awaiting the response of the backend.

This applies to the following buttons (feel free to extend the list):

- [x] Change email address
- [ ] Change password

tiltec:
Button should also be disabled if there's nothing to save (if there are no changes made to the form)

mr-kojo:
@chocos10 right, @tiltec has already mentioned that in his comment above ;-)
And I think this is part of client-side validation anyway (which is why I didn't create an extra bullet point for it).

chocos10:
I thought @tiltec  was talking about something else .

mr-kojo:
@chocos10 I requested two changes to your pull request (please see respective comments in the code).
Sorry I couldn't review earlier...

PR #992, opened 2018-03-31
email and password buttons disabled

chocos10:
Fixes this  #937 by disabling/enabling submit buttons. @tiltec  please review it

### DIFF
--- a/src/components/Settings/ChangeEmail.vue
+++ b/src/components/Settings/ChangeEmail.vue
@@ -34,7 +34,7 @@
       <q-btn
         color="primary"
         @click="save"
-        :disable="!hasEmailChanged"
+        :disable="!hasEmailChanged || !hasPassword"
         :loading="isPending"
         :value="isPending"
       >
@@ -58,7 +58,7 @@ export default {
       return this.newEmail && (previousEmail !== this.newEmail)
     },
     hasPassword () {
-      return this.password
+      return !!this.password
     },
   },
   props: {
--- a/src/components/Settings/ChangeEmail.vue
+++ b/src/components/Settings/ChangeEmail.vue
@@ -34,7 +34,8 @@
       <q-btn
         color="primary"
         @click="save"
-        loader
+        :disable="!hasEmailChanged"
+        :loading="isPending"
         :value="isPending"
       >
         {{ $t('BUTTON.CHANGE_EMAIL') }}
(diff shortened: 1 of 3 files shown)


src/components/Settings/ChangePassword.vue
```diff
@@ -50,6 +51,14 @@ import statusMixin from '@/mixins/statusMixin'
 export default {
   components: { QField, QInput, QBtn },
   mixins: [statusMixin],
+  computed: {
+    hasNewPassword () {
```
tiltec: I'd expect a variable that starts with `has` to return a boolean

chocos10: I hope these changes you were expecting :) @tiltec

src/components/Settings/ChangeEmail.vue
```diff
@@ -51,6 +52,15 @@ import VerificationWarning from '@/components/Settings/VerificationWarning'
 export default {
   components: { QField, QInput, QBtn, VerificationWarning },
   mixins: [statusMixin],
+  computed: {
+    hasEmailChanged () {
+      const previousEmail = this.user.email || this.user.unverifiedEmail
```
tiltec: I think `user.email` is always set, isn't it? Then you don't need to look up `user.unverifiedEmail`.

chocos10: @tiltec i will check it and do the necessary changes :+1:

chocos10: @tiltec  ya user.email works fine :) updated the changes thank you.

chocos10:
@tiltec  done :+1:

src/components/Settings/ChangeEmail.vue
mr-kojo: No, it has to be 
```code
const previousEmail = this.user.unverifiedEmail || this.user.email
```
because otherwise it's not possible to restore one's previous email address before verifying the new one.

src/components/Settings/ChangePassword.vue
```diff
@@ -34,7 +34,8 @@
       <q-btn
         color="primary"
         @click="save"
-        loader
+        :disable="!hasNewPassword || !hasOldPassword"
```
mr-kojo: Please also disable the button if the new password equals the old one (i.e., if the password hashes are equal).

chocos10: @mr-kojo  according to me that should be validated in the backend after submission for security purpose i talked to my mentor(for rgsoc) about this even he suggested to do the same as most of the websites follow. Whatever your suggestions i will proceed accordingly?

src/components/Settings/ChangeEmail.vue
chocos10: Sorry didn't saw that will change it now.

PR #993, opened 2018-04-01
add profile pic to search result 

ananya-agrawal:


tiltec:
Can you use the `ProfilePicture` component for this? Then the profile pictures will look uniform across the page. Also, it provides the random art if the user didn't upload a picture yet.

ananya-agrawal:
No, using quasar autocomplete we can't use ProfilePicture component. In that case the entire autocomplete feature have to be written again from scratch instead of using quasar autocomplete.

tiltec:
Yes indeed. Quasar autocomplete doesn't seem flexible enough for the purpose... But rewriting the topbar search could be a good exercise and might be good anyways, as the topbar search is not _exactly_ an autocomplete, but a bit more.

Would you be up for it?

PR #996, opened 2018-04-05
Allow to give feedback without entering weight

### LINKED ISSUE #976: Make it possible to give feedback without entering weight
Currently, the feedback form is quite focused on entering the weight:

image

It should be possible to give feedback without weight, because people might not want to estimate the amount. Entering a weight of "0 g" doesn't really fit and will drastically change the statistics.

The backend already supports comment-only feedback.

ananya-agrawal:
Closes #976


ananya-agrawal:
@tiltec please review it

tiltec:
Thanks! The changes you made are good, but now the UI looks a bit like it has an error:

image

Maybe it would be good to add a toggle button ("I want to enter the amount of food that we were able to save"), and only show the amount picker when the user chooses to.

Besides that, having feedback without amount makes some other bugs more apparent too:

- "NaN" weight value in feedback list (should maybe show no scale?)

image

- Two "NaN" values while editing feedback (could be solved with the toggle suggestion above)

image

PR #1000, opened 2018-04-08
email and password buttons disabled.

chocos10:
Fixes this #937 by disabling/enabling submit buttons. @mr-kojo  i have updated the changes you suggested in #992


src/components/Settings/ChangePassword.vue
```diff
@@ -58,6 +58,9 @@ export default {
     hasOldPassword () {
       return !!this.oldPassword
     },
+    hasPasswordChanged () {
+      return !(this.oldPassword === this.newPassword)
```
mr-kojo: Looks good.
One minor thing, though, regarding code style: You can simplify 

```code
return !(this.oldPassword === this.newPassword)
```
to
```code
return this.oldPassword !== this.newPassword
```

chocos10: i will update the changes @mr-kojo  :+1:

mr-kojo:
It seems you once again forgot to apply the other change I requested:

https://github.com/yunity/karrot-frontend/pull/992/files/7f6c8d80cf2cff6c0f0a61f402327dd1aeef4994#diff-c01de46328c8e96e2d052f67cbd5042fR57

Please apply this change, then it should be fine.

Issue #994, opened 2018-04-02
Open/Close message below search box should be Show/HIde


inktrap:
IMHO the Open/Close messages below the search box should be Show/HIde because that is what they do.

inktrap:
Would it be beneficial to look through the whole app at some point for similar improvements, or is this irrelevant for now?

tiltec:
Can you clarify what this is about, e.g. with a screenshot?

> Would it be beneficial to look through the whole app at some point for similar improvements, or is this irrelevant for now?

I think it's important, improving UX is a constant focus for karrot.

inktrap:
![karrot-world-open-close](https://user-images.githubusercontent.com/1449471/38386197-3114434a-3914-11e8-97fc-30751b13d384.png)

Since the result list is part of the same window, it shows/hides that part, the window stays open. That would be opinion.

tiltec:
I think it's reasonable to change it to show/hide and keep representing the action, not the current state.

Maybe the main reason why it's open/close is because we had it already translated: https://github.com/yunity/karrot-frontend/blob/master/src/locales/locale-en.json ;)

codebyuma:
@tiltec would it be ok if I attempted to fix this issue?

tiltec:
@codebyuma please go ahead :)

tiltec:
Ah, I need to clarify that. There's two steps for adding messages:

1. add them to `locale-en.json`
2. upload to transifex

Step 2 should only happen once the changes are merged into `master`, so you don't need care about it now.

codebyuma:
@tiltec awesome - ok I'll likely push up my PR tonight for review 👍

codebyuma:
PR is ready for your review: https://github.com/yunity/karrot-frontend/pull/1003

PR #1003, opened 2018-04-11
#994 - use 'show' and 'hide' for searchbox

codebyuma:
* adds `show` and `hide` to local-en.json
* updates `GroupGalleryUI` component to use the new terms
* updates snapshots to account for the updated component


inktrap:
This looks good, as far as I can see.

Two more things:

We would have to change/add the other languages too, so they can be translated by Transifex, right? What happens if something like 'HIDE' is not in one of the language files? Is the english fallback used?

The android build fails, but that looks like a password issue?!

tiltec:
> What happens if something like 'HIDE' is not in one of the language files? Is the english fallback used, or will this lead to an error?

@inktrap 
It will fall back to English, so there's no problem if the `locale-xx.json` files are not 100% up-to-date.

PR #1004, opened 2018-04-14
Notifications/Alerts revisited

### LINKED ISSUE #998: Use toasts instead of alerts where appropriate
In most of our use cases, quasar toasts are more appropriate than alerts.

mr-kojo:
Fixes #998 

So I thought about how to go about notifications.

**Conceptual proposal:**

We distinguish between _banners_ and _toasts_.

_Banners_
These are notifications that are potentially shown for a long period of time. Currently, this applies to `playgroundGroupInfo` and `awaitingAgreement`. 

_Toasts_
These are notifications that are shown only for a very short period of time or that require immediate interaction. This applies to all other notifications we have implemented so far, which are basically just sucess or error notifications, e.g., `inviteAcceptError` and `groupLeaveSuccess`. 

**Implementation proposal:**

_Banners_
There is no need to change the implementation for them, we just keep showing them fixed at the top of the screen, right below the navigation bar. See 288fc32 

_Toasts_
Quasar has a nice plugin for toasts, they call it [Notify](http://quasar-framework.org/components/notify.html). Basically, it wraps the `QAlert` component, adds some smooth animations to it, takes care of queuing and provides a very simple interface. I implemented a `notifyMixin` and a demo usage in 79caee9. 
In contrast to a banner a toast dismisses itself automatically, either after a configurable timeout or after clicking a button on it. This is very convienient because it thus doesn't need a vuex store. At the same time, it's the only drawback  in my opinion. This is because the toast has to be triggered using a callback (`done()`) instead of dismissing an action. It seems to me that there is no other reasonable way to use quasar's notify plugin. But in my opinion it would be okay. Remember we had already used a `done()` callback anyway to clear the password fields after password change. 
Please let me know what you think about it.


mr-kojo:
@tiltec Okay. Glad you like it.

@nicksellen Ah yes, that would be a nice solution, then we could indeed trigger toasts using actions. 
However, I'm afraid this wouldn't help to get rid of the other callback which already exists (the "clear-password-fields-after-password-change" callback). So we would still have two different ways to trigger something... Therefore I think I'll just stick to the current approach for now. Thanks for the suggestion. Hope you didn't get sunburnt 🌞

src/store/modules/toasts.js
```diff
@@ -0,0 +1,8 @@
+export default {
+  namespaced: true,
+  actions: {
+    show ({ commit }, toast) {
+      // The corresponding plugin takes care of the rest.
```
tiltec: Why use a plugin when you can directly write the code in here?

src/store/modules/banners.js
```diff
@@ -14,35 +14,35 @@ export default {
...
       }
 
-      return alerts
+      return banners
     },
   },
   actions: {
-    create ({ commit }, alert) {
-      return commit('create', alert)
+    create ({ commit }, banner) {
```
tiltec: Is this still used anywhere?

src/components/Layout/BannersUI.vue
```diff
@@ -0,0 +1,89 @@
...
+    playgroundGroupInfo () {
+      return {
+        color: 'positive',
+        icon: 'fa-child',
+        message: 'GROUP.PLAYGROUND_INFO',
+      }
+    },
+
+    bannerClasses () {
+      return this.$q.platform.is.mobile ? {
```
tiltec: Did you test the banner on mobile?

src/store/modules/toasts.js
mr-kojo: Because I didn't want to import quasar directly into a store module.
I thought it might be a bit cleaner if I do the quasar stuff in a plugin...

src/components/Layout/BannersUI.vue
mr-kojo: No, I spent quite some time trying to set up cordova but there were loads of issues so I asked @nicksellen to fix them first. 

However, I didn't change the banner UI anyway, so with respect to the banners there's no difference to the current production version of karrot.

src/store/modules/banners.js
mr-kojo: You mean banners that are created by actions?
No, but maybe in the future...

src/store/modules/toasts.js
tiltec: There's no big difference, when someone does `import store from "@/store" this also loads all plugins. 

Maybe it would be slightly simpler to test because one could load the module without the plugin, on the other hand it's quite easy to mock the Quasar dependency.

src/components/Layout/BannersUI.vue
tiltec: I think the only banner that's ever shown on mobile is the group agreement one (the playground banner is hidden because it was blocking topbar access). So maybe we can ignore it.

You don't need Cordova, you can use the responsive design view to test it, e.g. https://developer.mozilla.org/en-US/docs/Tools/Responsive_Design_Mode

src/store/modules/toasts.js
mr-kojo: So should I refactor it to remove the plugin? For me, both options are okay.

src/store/modules/banners.js
tiltec: I'm usually in favor of removing unused code, especially when it's untested. It's not lost, if the need arises one could still get the from the file history.

src/store/modules/toasts.js
tiltec: No strong need to rewrite, and the comment explains it well why the action is empty. With the rewrite, we wouldn't need a comment at all though ;)

tiltec:
Do you feel like adding a paragraph to the karrot docs about this? https://docs.karrot.world/

src/store/modules/banners.js
mr-kojo: Fair enough. Done in 7ddeda3

mr-kojo:
> Do you feel like adding a paragraph to the karrot docs about this?

Good idea. Done via #1008

Issue #786, opened 2017-11-23
Add a notice for unsupported browsers (Safari <9.1, IE <11)


tiltec:
I think it would be helpful to inform users that their browser is unsupported by quasar and we can't do much about it except asking upstream for support.

This mostly about older Apple devices and older Microsoft browsers: http://quasar-framework.org/guide/browser-support.html

---

This is an example error that we get from unsupported browsers:
(It's an error in the quasar browser detection code, funnily. Maybe at least this error can get fixed upstream?)

https://sentry.io/foodsaving-worldwide/karrot/issues/408157538/

```
TypeError: 'undefined' is not a function
  at onResize (/js/vendor.54a0396f8b977e976cd2.js:11)
  at n (/js/vendor.54a0396f8b977e976cd2.js:1)
  at None (/js/vendor.54a0396f8b977e976cd2.js:11)
  at None (/js/vendor.54a0396f8b977e976cd2.js:1)
  at rt (/js/vendor.54a0396f8b977e976cd2.js:1)
...
(1 additional frame(s) were not displayed)
```

Hardiki:
I am taking up this issue. @tiltec

peterkshultz:
Makes sense. I had tried implementing it with `Vue.use` in `src/main.js`, but wasn't seeing the behavior I was hoping for there either.

Given the size of the bundle, I'll try another implementation so long as @Hardiki is no longer working on this.

peterkshultz:
@tiltec: Given that older browsers don't support Quasar, it probably doesn't make sense to use one of Quasar's components (like a modal) to display the notice. 

Would a simple Javascript window `alert()` make sense?

Hardiki:
Hi @peterkshultz, Sorry for the delay message, I am working on it, I have done some work on it. Will share the code with you soon.

peterkshultz:
Sounds great @Hardiki !

Hardiki:
Here is the link to my Github repo here I have a bit of code, hope this helps. https://github.com/Hardiki/browserdetect

PR #1005, opened 2018-04-14
Add alert for outdated browsers

peterkshultz:
Fix for issue [#786](https://github.com/yunity/karrot-frontend/issues/786). Alerts users of Safari < 9.1 and IE < 11 that their browser is out of date.

Uses [`browser-detect`](https://www.npmjs.com/package/browser-detect), which comes in at [5.3 kB](https://bundlephobia.com/result?p=browser-detect@0.2.25) minified and gzipped. Relatively large given its purpose, but more reliable than checking the user agent string in Javascript.


peterkshultz:
CircleCI fails look like they're happening because `browser-detect` can't be recognized. Do I need to add information about that package to somewhere other than `package.json`?

tiltec:
You still have some references to `detect-browser` in the code, probably should be `browser-detect`.

peterkshultz:
@tiltec: Should I be worried about the Android integration test failure here? I've observed it fails on other PRs, even those that [don't modify code](https://github.com/yunity/karrot-frontend/pull/1007).

src/pages/MainLayout.vue
```diff
@@ -112,6 +118,16 @@ export default {
     defaultShowSidenavWidth () {
       return 992
     },
+    browserDetect () {
```
tiltec: I guess this list is not exhaustive? So we would need to add more devices in future.

peterkshultz: The original link to Quasar's browser support in #786 is broken, but I was able to find one [here](http://v0-14.quasar-framework.org/guide/browser-support.html). Aside from Android, IE and Safari are the only ones listed with version numbers, so I figured those were the only ones to check.

Want me to add Android too?

tiltec: No need to add Android right now. I guess we should add something later that tells them there's an app available (after we released the karrot app in Play store)

src/pages/MainLayout.vue
```diff
@@ -1,5 +1,10 @@
 <template>
-  <div>
+  <div v-if="browserDetect">
+    <p>You are using an unsupported browser. Please <a
```
tiltec: Can you make it translatable? A significant number of people have no grasp of English language.

browser-update.org is also multi-lingual, so that makes sense to me.

Oh by the way, can you set the link to open in a new window? Don't forget to add `noopener` to avoid XSS attacks: https://mathiasbynens.github.io/rel-noopener/

peterkshultz: Certainly! Are you referring to adding the [`translate`](https://www.w3.org/International/questions/qa-translate-flag) flag to the `<p>` and `<a>` elements? Or is there something I should do related to Transifex?

tiltec:
Have a look at other strings in the karrot code - they are in `locale-en.json` and get translated with `vue-i18n`.

You don't need care about transifex.

src/pages/MainLayout.vue
```diff
@@ -1,5 +1,13 @@
 <template>
-  <div>
+  <div v-if="browserDetect">
```
tiltec: `browserDetect` seems a not very fitting name, how about `unsupportedBrowser`?

src/pages/MainLayout.vue
```diff
@@ -1,5 +1,13 @@
...
+    <p> {{ $t('OUTDATED.FIRST_MESSAGE') }} <a
```
tiltec: You could use the [vue-i18n functional component](https://kazupon.github.io/vue-i18n/en/interpolation.html) to get slightly nicer translation strings, like here:

https://github.com/yunity/karrot-frontend/blob/61200a6eba32ca0daafd845961d183b79e350f7f/src/components/Settings/VerificationWarning.vue#L6-L18
with this message https://github.com/yunity/karrot-frontend/blob/61200a6eba32ca0daafd845961d183b79e350f7f/src/locales/locale-en.json#L333

PR #1015, opened 2018-05-07
Add sparkpost docs

nicksellen:
I added some initial documentation for the sparkpost stuff as I went through it. Almost certainly not totally correct, but a start!

### DIFF
--- a/docs/sparkpost.md
+++ b/docs/sparkpost.md
@@ -34,7 +34,9 @@ I'm not sure if we need to also configure the bounce domain bit.
 
 You also need to setup the inbound email domain for "reply to this email" functionality. By convention we use the main application domain prepended with "replies." e.g. `karrot.world` -> `replies.karrot.world`.
 
-For this domain you need to give it some MX records so they will arrive at Sparkpost [TODO: provide link or details].
+For this domain you need to give it some MX records so they will arrive at Sparkpost. See the [Sparkpost docs](https://www.sparkpost.com/docs/tech-resources/inbound-email-relay-webhook/#add-mx-records) for which records to add.
+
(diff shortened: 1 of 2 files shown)


nicksellen:
Would benefit from a review from @tiltec - but also would be happy to merge it sooner rather than have it 100% correct. It has a disclaimer at the top already...

docs/sparkpost.md
```diff
@@ -0,0 +1,58 @@
...
+
+Keep hold of those two API keys as you'll need to store them as ansible secrets.
+
+Later, during the ansible deployment a webhook and a relay setup (inbound email configuration) will be created for you.
+
+## DNS records
+
+When you create a sending domain in Sparkpost, it will tell you which DNS record to add. You want to add the TXT one to verify domain ownership.
+
+I'm not sure if we need to also configure the bounce domain bit.
```
tiltec: I think if we don't, the sparkpost does something for us? Would be worth checking at some point.

docs/sparkpost.md
```diff
@@ -0,0 +1,58 @@
...
+
+You also need to setup the inbound email domain for "reply to this email" functionality. By convention we use the main application domain prepended with "replies." e.g. `karrot.world` -> `replies.karrot.world`.
+
+For this domain you need to give it some MX records so they will arrive at Sparkpost [TODO: provide link or details].
```
tiltec: This was a bit tricky in Namecheap. Setting the MX records for a subdomain somehow disabled the wildcard A records…

nicksellen: Added a note about that, and the link to sparkpost docs.

PR #1016, opened 2018-05-10
980 icon of deleted user is clickable

### LINKED ISSUE #980: Icon of deleted user is clickable
After a user deleted her profile, other users can still click her user icon (e.g., on a group's wall), which redirects to the `groupPreview` page.
Instead, it would be nice to display a uniform icon for all deleted users which isn't clickable but shows a message on mouse-hover (something like _"this user has been deleted"_).

pogopaule:
Closes #980.

## Without this fix
screenshot at may 10 20-04-57

## With this fix + layout improvements of questionmark
screenshot at may 10 22-17-08

screenshot at may 10 22-17-42

### DIFF
--- a/src/components/ProfilePictures/ProfilePicture.vue
+++ b/src/components/ProfilePictures/ProfilePicture.vue
@@ -34,14 +34,16 @@
         :style="pictureStyle"
       />
     </div>
-    <span v-else>
-      <span>?</span>
+    <div
+      v-else
+      class="deletedUser"
+      :style="deletedUserStyle"
+    >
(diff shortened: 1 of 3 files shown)


src/components/ProfilePictures/ProfilePicture.vue
```diff
@@ -84,4 +89,8 @@ export default {
 .randomArt
   display block
   overflow hidden
```
tiltec: I guess we should a stylus linter again, this starts to look a bit random ;)

PR #1019, opened 2018-05-19
Translate Karrot Frontend to Brazilian Portuguese

trmendes:
Next time, adding some new translation or fixing those I'm pulling now, I'll use transifex.

This is the json file translated to Brazilian Portuguese. It's a basic translation but as soon as I get familiar with Karrot I'll make the changes we need to have a more accurated translation.

### DIFF
--- a/src/locales/locale-pt_BR.json
+++ b/src/locales/locale-pt_BR.json
@@ -318,7 +318,7 @@
     "PICKUP_MODIFY": "modificou as configurações de uma coleta",
     "PICKUP_DELETE": "deletou uma coleta",
     "SERIES_CREATE": "crirou uma nova coleta recorrente",
-    "SERIES_MODIFY": "modificou as configurações para uma coleta recorrente",,
+    "SERIES_MODIFY": "modificou as configurações para uma coleta recorrente",
     "SERIES_DELETE": "deletou uma coleta recorrente",
     "PICKUP_DONE": "fez uma coleta na loja {storeName}",
     "PICKUP_MISSED": "A coleta na {storeName} não foi feita!",
--- a/src/locales/locale-pt_BR.json
+++ b/src/locales/locale-pt_BR.json
@@ -1,0 +1,397 @@
+
+  "GLOBAL": {
+    "GITHUB_NOTE": "Desenvolvimento",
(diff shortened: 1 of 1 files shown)


src/locales/locale-pt_BR.json
```diff
@@ -0,0 +1,397 @@
...
+    "GROUP_JOIN": "entrou no grupo",
+    "GROUP_LEAVE": "deixou o grupo",
+    "STORE_CREATE": "crirou a loja {name}",
+    "STORE_MODIFY": "modificou as configurações da loja {name}",
+    "STORE_DELETE": "deletou a loja {name}",
+    "PICKUP_CREATE": "criou uma nova coleta pontual",
+    "PICKUP_MODIFY": "modificou as configurações de uma coleta",
+    "PICKUP_DELETE": "deletou uma coleta",
+    "SERIES_CREATE": "crirou uma nova coleta recorrente",
+    "SERIES_MODIFY": "modificou as configurações para uma coleta recorrente",,
```
tiltec: Duplicated comma at the end of the line

tiltec:
Hey @trmendes I had to revert these changes in https://github.com/yunity/karrot-frontend/pull/1030 because the JSON file is invalid. Transifex rejected it with `Exception: Could not import file: Extra data: line 2 column 11 - line 398 column 1 (char 11 - 16120)`.
I guess it's because you are missing an opening curly braces in line 1. Can you fix this problem and open another PR?

Did you test the translations in a dev environment? I'm wondering, because I think that invalid JSON should have thrown an error at this point already.

trmendes:
@tiltec I probably made a mistake when saving on vim. Probably hit x and deleted the first { before global.

I'm so sorry about it.

About testing it. It's my first time at this project so I still not sure if I'm doing things right. (I can see now I'm not :O)

To test it I changed the index.js file into the locales dir to add the pt_BR support  and ran 'yarn run dev' to check the log messages. Am I'm missing something here?

Thanks for your help. I'm going to wait for your message before I do any other PR here.

Thanks again

tiltec:
Ah, so the problem was introduced after you tested it! No worries, can happen to everyone :)

While you're at fixing it, you could add Portuguese to the UI language selector and commit this change too. `locales/index.js` should be the file.

PR #1021, opened 2018-05-19
WIP enter recurring pickup recurrence rule via natural language

pogopaule:
@tiltec here is the WIP RP you asked for

### DIFF
--- a/package.json
+++ b/package.json
@@ -57,6 +57,7 @@
     "quasar-framework": "^0.15.8",
     "raven-js": "^3.20.1",
     "reconnecting-websocket": "^3.2.1",
+    "rrule": "2.2.9",
     "twemoji": "^2.5.0",
     "typeface-cabin-sketch": "^0.0.54",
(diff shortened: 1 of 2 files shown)


tiltec:
Thanks! It seems like a nice addition in usability, but it also adds [~12kB gzipped](https://bundlephobia.com/result?p=rrule@2.2.9) to the bundle. Though it might only get loaded when the PickupSeriesEdit page get opened, thanks to webpacks lazy loading.

It also adds complexity to translations, I'm unsure if our translators can handle these regexes well: https://github.com/jakubroztocil/rrule/blob/master/lib/nlp.js#L940
Another option would be to leave it untranslated, which might be better than having to input "rrule code".

My proposal would be to identify which recurrence rules our users typically need and add these as extra options between "weekly" and "custom". The full rrule power could stay as is for power users, but maybe the hint text should be rephrased (to make it clear that it expects "rrule code", not natural language).

PR #1022, opened 2018-05-20
go to previous page after saving group

### LINKED ISSUE #999: Redirect to previous page after editing group description
The group edit page can get opened from multiple places:

- Group overview page (in sidebar, the "three dots menu")
- Group description page (one of the action buttons)
- If no store or group location is set: group map in sidebar

pogopaule:
fixes #999

### DIFF
--- a/src/store/modules/groups.js
+++ b/src/store/modules/groups.js
@@ -69,7 +69,7 @@ export default {
         const data = await groups.save(group)
         commit('update', data)
         dispatch('currentGroup/update', data, { root: true })
-        router.push({ name: 'group', params: { groupId: group.id } })
+        router.go(-1)
       },
 
       async join ({ commit, dispatch, rootGetters }, { id: groupId, password }) {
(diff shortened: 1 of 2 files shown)


src/store/modules/groups.js
```diff
@@ -69,7 +69,7 @@ export default {
         const data = await groups.save(group)
         commit('update', data)
         dispatch('currentGroup/update', data, { root: true })
-        router.push({ name: 'group', params: { groupId: group.id } })
+        router.go(-1)
```
tiltec: This does nothing if the user opens the edit link directly and saves the form, e.g. https://dev.karrot.world/#/group/6/edit

It does the wrong thing if the user opens this link from another website (e.g. a chat application) in the same tab. Then it would direct the user back to the chat application after saving the form.

As the browser `history.length` does not help us very much there (it only shows the number of pages the user has visited in the current tab, but we don't know if the pages before have been on karrot), it might make sense to introduce a state that indicates if the current page is the first karrot page the user is visiting.

It could be implemented as [vue-router navigation guard](https://router.vuejs.org/en/advanced/navigation-guards.html), checking if the `from` state is defined. For [redirects](https://router.vuejs.org/en/essentials/redirect-and-alias.html), navigation guards are only executed for the destination, so it's necessary to check if the `from` state is still as expected.

tiltec:
Closing this because of https://github.com/yunity/karrot-frontend/issues/999#issuecomment-399747726

PR #1031, opened 2018-05-28
Add support to BR Portuguese

trmendes:


trmendes:
```json
ptBR: {
....
dateFnsLocale: () => import('date-fns/locale/en'),
}
```

I left it as en because there is no ptBR at date-fns. Do you know something about it? I couldn't find it

tiltec:
You could use their `pt` locale: https://github.com/date-fns/date-fns/tree/v1/src/locale/pt

trmendes:
pt stands for Portuguese from Portugal. I wasn't sure about it.

tiltec:
Yeah, it's a different locale, that's for sure. But maybe still better than English for now? Are the differences big?

You can still add a `pt_BR` translation to datefns later and then switch

By the way, we didn't add separate locales for de_CH, de_AT or cn_TW yet, because no user complained so far. As it's quite some work to keep the translations up to date when we add more features to karrot (don't forget all the e-mail templates in the backend!), it makes sense to me to use the pt_BR locale also for pt. Or maybe there's an automatic browser fallback? Don't know...

tiltec:
The snapshot tests fail because the `status-frontend.json` file was updated and the language options are not mocked out in tests (yet). 
Can you run `yarn test -u` and commit the changed file to make the test pass?

PR #1032, opened 2018-05-29
#968 bug redirect on user settings save

### LINKED ISSUE #968: Redirect to personal profile after changing it

alekpentchev:
Closes #968 

I have resolved the issue with the user being redirected to the profile page after updating settings.
Now it does not redirect. Instead the user will see a notification at the bottom of the page (I used Quasar framework to stay in the same environment).
Notification for requests with errors also added - a red notification at the bottom.

Everything should be fine - checked on Chrome, Firefox and Edge. Tests run - no problems.


tiltec:
Thanks @alekpentchev, looks good!

From a code style perspective, it would be good to use the "Toasts" infrastructure that we recently added: https://docs.karrot.world/notifications.html

It's a tiny abstraction on top of Quasar's `Notify`, but it allows us to define a default configuration.

src/store/modules/auth.js
```diff
@@ -110,8 +110,21 @@ export default {
 
     ...withMeta({
       async save ({ dispatch }, data) {
-        const savedUser = await dispatch('backgroundSave', data)
-        router.push({ name: 'user', params: { userId: savedUser.id } })
+        try {
+          await dispatch('backgroundSave', data)
+          dispatch('toasts/show', {
+            message: 'Your changes have been saved!',
```
tiltec: The only thing left would be to make these messages translatable. We use `vue-i18n` for this. You can import our instance from `@/i18n` and use the `i18n.t` function. The messages are defined in `locales/en.json`.

alekpentchev:
All change requests were addressed. 
Messages are now translatable and I have also added some translations.

tiltec:
Thanks! Our translations are managed by transifex. Unfortunately, transifex doesn't have any "merge" capabilities, so it's a bit complicated if translations get added outside of transifex. I'll copy-paste it manually to transifex after this PR is merged to master.

If you want to add more translations in future, please add them in transifex directly! -> https://www.transifex.com/yunity-1/karrot/translate/

PR #1036, opened 2018-06-10
2018 06 deps upgrade

tiltec:
Puh, that was some hard work upgrading dependencies!

- node 10.3 has a problem with promises, which only shows up with recent versions of `core-js`. This was a hard one. Be sure to get node >10.4 or stay with an older version
- some change caused the `QInfiniteScroll` component to hit the recursive limit of vue when the page is loading. I worked around that by adding a `Vue.nextTick` as delay to the `done` call.
- `vue2-leaflet` had major changes concerning component names and event names. It also had a completely unexpected and undocumented change how it renders its components. The `mounted()` function needs to be called, otherwise `this.ready` is not set to true and no map component shows. `vue-test-utils` doesn't call this properly (unexpected in itself!), so it added an `await Vue.nextTick()` to affected tests. The storyshots with maps in them are severely incomplete, but the `nextTick` trick didn't help there unfortunately
- our `test/helpers.js` mount function included a global stub for `window.getComputedStyle()`, which unfortunately was quite limited. This introduced a conflict with an undocumented change in Quasar.
- the name of `vue-test-utils` was changed to `@vue/test-utils`. It also received major badly documented changes, which made lots of our tests fail or throw errors. I didn't upgrade, instead I would rather wait for a proper release with documentation.

I'll merge this PR now to give people the opportunity to test it on dev.karrot.world.

EDIT:
Upgrading `clone-deep` to v4 causes `yarn build` to fail:

```
Unexpected token: keyword (const) [./node_modules/clone-deep/index.js:28,0]
```

Seems like it doesn't get transpiled by babel...

P.S.:
Updates to webpack 4 are still tbd. Maybe it would be a chance to switch to Quasar CLI, but that would mean restructuring frontend files a bit.


tiltec:
Oops, CI couldn't create the APK:

```
* What went wrong:
Could not resolve all files for configuration ':releaseCompileClasspath'.
> Could not find runtime.jar (android.arch.lifecycle:runtime:1.0.0).
  Searched in the following locations:
      https://jcenter.bintray.com/android/arch/lifecycle/runtime/1.0.0/runtime-1.0.0.jar
> Could not find common.jar (android.arch.core:common:1.0.0).
  Searched in the following locations:
      https://jcenter.bintray.com/android/arch/core/common/1.0.0/common-1.0.0.jar
```

I didn't try it locally.

I'll be back on Thursday, will have a closer look then.

tiltec:
Back to work! Android build seems to fail in `master` too: https://circleci.com/gh/yunity/karrot-frontend/10711
I suspect it's not our fault, but we need to deal with it... @nicksellen any insight how this can happen? I'll try to reproduce it locally.

PR #1037, opened 2018-06-14
Add pickup conversations

nicksellen:
Initial UI work on pickup conversations.
TODO:
- [x] actually fetch proper data about pickup
- [x] work out where to store state of whether to show it or not
- [x] add links to pickups to open side detail/conversation
- [x] work out how mobile view should work
- [x] fix the negative margin issue at the bottom (because of the carrot logo poking up, it makes it funny...)
- [x] don't always jump to the bottom after update (e.g. reaction update should not jump to bottom, and could be annoying if you are just flicking through old messages)
- [x] work out pagination/infinate scroll stuff (looks like quasar infinite scroll is only designed to work downwards...)
- [x] ~~maybe get more info/actions in the top bar bit, it looks a bit small to me right now~~
- [x] work out how to handle marking messages as read
- [x] put text into translation system
- [x] autofocus conversation input on opening detail sidebar
- [x] ~~consider whether an overall conversation/notification thing is needed to avoid missing messages~~ (not in this PR)
- [x] handle case where user is no longer signed up to pickup/conversation and so it should go away
- [x] make route redirect to group page on desktop with the side panel open
- [x] add tests


nicksellen:
UI looks like this for now:

pickup-conversation2

nicksellen:
Hmm, for some reason the netlify preview doesn't show anything interesting yet...

tiltec:
Looks great already - but lots of thoughts necessary for this! I'm happy to give some input in the next days, what do you think are the showstoppers before it can get merged to master?

nicksellen:
Would be worth getting https://github.com/yunity/karrot-backend/pull/544 merged first - it's much simpler, and allows us to use dev backend and/or netlify preview.

I think the main usability showstopper could be missed messages - as we don't have a conversation notification overview page, new messages won't be very visible (they do get sent by email and push notification, but the desktop view wouldn't make it very apparant that there is a new message).

Otherwise, mostly details. Pagination is a bit annoying and looks like I'll need to reimplement a bunch of the quasar infinite scroll thing :/ ... also read messages, would like it a bit more automatic than the wall, but didn't work anything out yet.

nicksellen:
I added automatic read-marking - basically, if you are not away, then we mark the most recent received as read, as they arrive. If you are away, we don't.

nicksellen:
Oh, infinate scroll wasn't so tricky to implement, given we already have the nice cursor/status stuff implemented.

nicksellen:
The [netlify preview](https://deploy-preview-1037--karrot.netlify.com/) is working now :) - well, apart from that netlify doesn't proxy websockets yet...

tiltec:
Cool! Tried it out, works :)

About missed messages: not a problem right now, can add a "to catch up" page in future. Maybe as banner on the group wall as "you have new messages". Can also change the favicon (add a red dot) and add an `(!)` at the beginning of the browser titlebar. We can also get browser push notifications. Maybe I should write this into some issue instead of here ;)

About email notifications: would be cool if there was a hyperlink that jumps directly to the pickup conversation view. Maybe with a route parameter?

nicksellen:
Ok, I added a mobile page for pickup conversations (well pickup detail I termed it - with a vague idea that it can be a more general detail page for different things, with or without conversations as part of them).

I think I'll try and make it so that the pickup detail link will redirect to the group page with the pickup conversation open, or something like that. I wanted to avoid having to store the side detail state in the url too. Which is maybe not ideal, but it's quite complicated to support a link parameter (as it can be on any page).

nicksellen:
Thanks for feedback! I did the following:
* made the whole pickup clickable (I left the "Open Chat" link too as I don't think people will know the click the pickups to start with, and a button looked a bit odd...) - oh, I forgot to add "Open Chat" to i18n
* made the pickup detail route nested in the group/store (simplifies the action as we have the group/store ids)
* changed the toolbar header thing as suggested
* added breadcrumbs for group/store

src/components/General/Detail.vue
```diff
@@ -0,0 +1,289 @@
...
+      send: 'conversations/send',
+      mark: 'conversations/mark',
+      toggleEmailNotifications: 'conversations/maybeToggleEmailNotifications',
+      toggleReaction: 'conversations/toggleReaction',
+      fetchMore: 'conversations/fetchMore',
+    }),
+    markRead (messageId) {
+      if (!this.conversation) return
+      this.mark({
+        id: this.conversation.id,
```
tiltec: During testing, I sometimes see requests to `/api/conversations/undefined/mark/` when opening the pickup chat while the page is loading.
The check above should prevent it, but maybe it doesn't? Could need some debugging, or could also be ignored...

tiltec:
The backend doesn't send out notification emails for pickup conversation messages, does it? It raises this exception: https://sentry.io/foodsaving-worldwide/backend/issues/582451427

nicksellen:
Ah, good point. Will do a backend fix for that. At least the error message was nice and clear :)

tiltec:
Ready to merge?

tiltec:
image

tiltec:
Still needs something to toggle pickup chat notification emails.

tiltec:
image

nicksellen:
Looks good, does it wrap nicely if there are lots of people in there? (which would be quite rare for a pickup...)

tiltec:
image


Seems to work nicely without fiddling. Not super space efficient, but ok for the purpose.

tiltec:
I'm done for today, seems to be overly ready for dev.karrot.world :)
@nicksellen merge as you like!

PR #1042, opened 2018-06-20
Performance improvements

nicksellen:
A few performance improvements after some informal profiling (see https://medium.com/@brockreece/unlock-performance-tracing-in-vue-3b2c8f619cdc).

A few details:

`RandomArt` now uses the `use` feature of `<svg>`, so we have a global svg object that define sub objects, that can be reused around the page (see the svg element with id `RandomArtRootSVG` at the bottom of the body).

Emoji elements are also cached - we use `twemoji` to create an element, and then `Element.cloneNode(true)` to make them each time we need it.

Async components vue supports components to be [async](https://vuejs.org/v2/guide/components-dynamic-async.html#Async-Components), a bit like for the routes, but for local registration. I changed two things to only load their popover content when the popover is actually open (`LocaleSelect` and `ConversationAddReaction`).

I changed a bunch of `<q-tooltip>` instances to just use regular html `title` attribute.

After leaving group wall page, i truncate the wall conversation to the first 10 messages.


nicksellen:
> Even though you say it's informal, did you record quantitative improvements?

Yes, but not in a rigourous way. I basically loaded various pages with the profiler enabled, looked at the things that took most time, fixed/removed/improved them, then ran the profiler again.

You can try the subjective difference visiting https://deploy-preview-1042--karrot.netlify.com/ vs https://dev.karrot.world/ - focusing on the wall, members, and pickups pages.

nicksellen:
The other area of performance improvements I'd like to see are whole page load times, here's a screengrab of a performance profile (you can get the json file to load into chrome from [here](http://nicksellen.co.uk/upld/Profile-20180622T150548-group-6-pickups.json)).

The big blob of orange in the middle is the main page rendering, the scattered bits to the right seem to be re-rendering bits of the pickup list (not entirely sure why). The whole blank area to the left of the main orange blob is app initialization - we wait to download loads of data before rendering anything. And most of these things aren't actually needed on this page (e.g. pickup feedback items, group wall conversation).

I think it's too much for this PR, so should probably create another issue, but the goal would be to render the main skeleton of the app as early as possible, then trigger loading of the data for that page, then only after that fetch any related data we might want (but could push more of that to routing triggers).

profile

tiltec:
Agree, a quick first render could be nice :) Although we will never be as fast as SSR-pages.

src/store/modules/conversations.js
```diff
@@ -297,8 +310,9 @@ export default {
         }
       }
     },
-    setCursor (state, { conversationId, cursor }) {
+    setCursor (state, { conversationId, cursor, first = false }) {
       Vue.set(state.cursors, conversationId, cursor)
+      if (first) Vue.set(state.entries[conversationId], 'firstCursor', cursor)
```
tiltec: `state.entries[conversationId]` gets overwritten pretty often, e.g. when somebody writes a new message in that conversation, the server sends out that update via websocket (for updating `unreadMessageCount`), which triggers `setConversation`. So my guess is that `firstCursor` will be undefined even when it shouldn't be. I didn't confirm this behavior though.

Also, I quite often get this error when loading pickup conversations:

```
Cannot set reactive property on undefined, null, or primitive value: undefined
TypeError: right-hand side of 'in' should be an object, got undefined
Stack trace:
set@webpack-internal:///5:1032:1
setCursor@webpack-internal:///328:718:9
...
```
Seems that `state.entries[conversationId]` is undefined, I just wonder why...

nicksellen: Yeah, the `firstCursor` was a bit of a quick hack really. I think I would actually be up for loading this kind of data on entering the route, and clearing it when leaving, but that needs more change.

Another idea would be to store each cursor inside the message too, so when we want to clear out some conversation, can find a point at which we have the cursor available to get the next entries.

Or keep all cursors aroun in the `state.cursors` storage so we can find them again when needed. Probably with a reference to the message id they correspond to.

It seems `state.entries[conversationId]` could be undefined because the messages and conversation are loaded in parallel [1], so if the messages "win" the race, it'll try and set something on the non-existant conversation object.

Although it currently loads the conversation twice, once through the pickup related endpoint, and once through the conversation endpoint, so might be better to only load it in the `conversations` module via a `fetchConversationForPickup()` method.

[1] from `detail.js`
```js
await Promise.all([
  dispatch('conversations/fetchConversation', conversation.id, { root: true }),
  dispatch('conversations/fetch', conversation.id, { root: true }), // gets the messages
])
```

Issue #973, opened 2018-03-19
Improve group sidenav


djahnie:
I'm just now writing a new walkthrough and thus have a closer look at Karrot from all angles. I'm now at the chapter 'Navigation', and I guess we all agree that this is not perfect...

So here are some concrete suggestions!

The group sidenav is pretty redundant imho. That's what it looks like rn:
![image](https://user-images.githubusercontent.com/17573771/37613547-936e21ec-2ba8-11e8-8b6b-66a39ac6ddae.png)

How about this:
- Move 'edit group' into the group settings (which so far only hold the email notification settings)
- Remove 'show public group info', since it's already shown in the private description
- Remove 'invite member to group', since it's already accessible on the member page
- Move 'leave group' to the personal profile (only visible on your own profile, e.g. as a collapsible list with an option to directly switch to a different group or leave it - details for this would be a different issue. But moving it makes a lot of sense, since the EfA people did not find the option to leave a group at all...)
- Now we can get rid of the three dot menu! :smiley:

ananya-agrawal:
I want to do this issue.

djahnie:
I'm not sure if this is ready to just be tackled like this - I first would like to get more feedback, e.g. from @nicksellen and @D0nPiano. I'm not even sure @tiltec has said anything so far.

Also, didn't you just grab issue #898 @ananya-agrawal? I didn't see a pull request concerning the public group info thingy. Please don't claim new issues before finishing (or giving up on) the one you already started working on.

tiltec:
@ananya-agrawal From my POV, this topic needs discussion before code changes can start. What do you think about the points that @djahnie proposed?

My comments:

> Remove 'show public group info', since it's already shown in the private description
> Remove 'invite member to group', since it's already accessible on the member page

These options are there for backwards compatibility, but maybe we can remove them by now. Actually I don't know. I don't have a good concept how we can do UI changes gradually, without confusing existing users too much.

Additionally, the "show public group info" functionality in the private description page is a bit inferior, since it can't be linked to (it just toggles the page content). So if we remove the button in the "three dots menu", there's no way to reach the public group info page. I guess we should first modify the button in the private group info page to at least change the URL appropriately, or have both descriptions on the same page below each other.

> Move 'edit group' into the group settings (which so far only hold the email notification settings)

You mean move the button into the group settings page, or the whole group edit form? I think moving the button there could be good, but it comes at the cost of needing two clicks to reach the group edit form.

[...]

Issue #958, opened 2018-03-11
Improve navigation for typical user flows


tiltec:
Yesterday at the Gothenburg hackathon presentation I noticed that karrot already has quite a lot navigation elements, with sometimes unclear priorities. It could be good to write down some typical karrot user flows and develop a navigation concept with clearer guidance.

I'd consider this issue as done when we identified 2-3 typical things that most people do on karrot and  streamlined the navigation by raising the visibility of those navigation elements and hiding some of the less important pages.

(sorry, no pictures yet)

Related: Janina conducted our first formal usability test and wrote down some notes about it: https://slack-files.com/T0B6WCFM5-F9M5KMN8G-db8a31ec5d

Related issues:
#683 about a redesigned sidenav bar
#905 about a tutorial for karrot

djahnie:
How do you imagine coming up with typical user flows? I could write up some examples and then ask in the facebook group if that's close to reality or where the differences are if there are any. I just fear that we won't get as detailed answers as we'd hope for... I'll still give it a try, I guess.

Any suggestions what to start with?

tiltec:
As this is being revisited in #1047, I'll close this issue.

Issue #683, opened 2017-10-31
Sidenav Restructure Proposal


alangecker:
![screenshot_2017-10-31_01-09-10](https://user-images.githubusercontent.com/10246027/32219585-eeb50a44-be2e-11e7-986a-78e74d71ce71.png)

## further ideas
- Store/User-toggles on top of the map on hover
- Maximize map button

D0nPiano:
Is it meant to be the new sidenav for desktop?

tiltec:
I like your work on this!

alangecker:
![image](https://user-images.githubusercontent.com/10246027/32238769-0172dee2-be69-11e7-9e79-afe0a9a82e0c.png)

tiltec:
This is quite important to me, maybe I'll have a try at implementing the design proposal. Would be happy about support :)

tiltec:
This is being superseded by the experiment in #1047

PR #1047, opened 2018-06-23
WIP: New sidenav

nicksellen:
In the spirit of https://github.com/yunity/karrot-frontend/issues/683

Should also consider:
- [Improve navigation for typical user flows #958 ](https://github.com/yunity/karrot-frontend/issues/958)
- [Improve group sidenav #973](https://github.com/yunity/karrot-frontend/issues/973)

A few thoughts/questions that came up:
- should stores really have their own sections for pickups/feedback/etc or should it be filters on the group versions of those pages
- seems quite nice to have indented nested submenus for lesser used things that relate to the main item
- feedback could be nested under pickups
- maybe also nested under pickups, could be "My upcoming pickups" / "Pickups in need of someone..." / "manage pickups"
- description might not really deserve it's own menu
- the "edit group" button is on the description page, but that's for editing everything about the group
- maybe could have group name somewhere on left nav
- would be nice to open user profile in the context of the current group (not sure what to do if they are not in the current group)
- I missed off lots of things that were in the toolbars, they would need to go somewhere too

newsidebar


tiltec:
Maybe most of the toolbar contents could get moved to an "action bar" on top of the page, similar to the actions buttons in store pickup view.

image

For the group wall, we could move the 3 alert bars into the toolbar, with a combination of icon+number per alert.

The map toggles (user on/off, stores on/off, fullscreen) could become something similar to leaflet map controls, floating inside the map.

tiltec:
Group description, members and history could be joined into one "Details" sidenav entry?

DougInAMug:
@nicksellen Do I interpret your screenshot correctly, that you are planning to have Store Pickups, Feedback and History nested below the Store? This would 'fix' the UI confusion Jess and I had when using Karrot yesterday. She mentioned that even switching the existing panes (i.e. having the Stores pane above the Store Pickup/Fedback/History) would also be better.

(Further to the Foodsharing Edinburgh meeting I attended 2018-07-02)

nicksellen:
@DougInAMug yes that's how I did it there, but so far this is just an idea, I wanted to spend a bit more time thinking more deeply about user navigation as it is certainly a bit confusing at the moment. Actually when I showed people here that new UI they didn't see the Pickups/Feedback/History as nested. Another approach there is to use filters on the Group versions of those pages.

I think we need to introduce some new concepts or hierarchies to make the common things very clear and easy, and put the less frequently used things deeper in the navigation. Also, a bit more nesting (e.g. pickup feedback could be nested under pickups).

Input very welcome though! (especially with nice drawings!).

nicksellen:
Closing in favour of https://github.com/yunity/karrot-frontend/pull/1071

PR #1048, opened 2018-06-23
Slightly unify color usage

tiltec:
Biggest change is a freshly implemented leaflet marker icon component. The old one didn't let me choose the Quasar colors.

- differentiate colors `secondary` and `positive` better
- reduce usage of ugly `tertiary` color
- reduce footer size and make look lighter

image

Address picker supports custom marker colors and font icons:
image


src/services/svgUtils.js
```diff
@@ -184,6 +184,8 @@ function getRootDefs () {
     if (!rootSVG) {
       rootSVG = document.createElementNS(SVGNS, 'svg')
       rootSVG.setAttribute('id', 'RandomArtRootSVG')
+      rootSVG.style.position = 'absolute'
```
tiltec: @nicksellen I sometimes had a big white space at the bottom, this fixed it. Should I leave it in here or is there a better solution?

nicksellen: Ah I see, I didn't realise it would have taken up any space in the DOM. If it fixes it then fine!

nicksellen:
The footer text looks a bit small

tiltec:
I'll merge, but I have no resistance against making the footer font a bit bigger and the icon in the center (was only centered in mobile view before, not on desktop).

PR #1049, opened 2018-06-24
Data loading tweaks

nicksellen:
Here is a version of conversation loading that always loads messages on demand.

It:
- removes the concept of `activeConversation` in `conversations` module
- adds concept of caching conversation id mapping for groups and pickups
- will leave the conversation data in state
- will clear out message data in state
- `<Conversation>` will render ok whilst message data is loading - made most actions load the data async, so page rendering continues
- should simplify cursor handling (re: https://github.com/yunity/karrot-frontend/pull/1042#pullrequestreview-131414740)

... might try it for feedback too, as I think this is not needed so often, and it causes a lot of requests (as it fetches each related pickup too, as they aren't ready on first page load)


tiltec:
Looks good! I didn't review in great detail, but the changes I saw make sense to me.

At some point maybe the paginationModule could be extended to also support the conversations use case:
https://github.com/yunity/karrot-frontend/blob/2ad6c8428d5dcb7d4af1164dd2540e674fed0d2e/src/store/helpers.js#L211

src/components/Conversation/Conversation.vue
```diff
@@ -3,51 +3,52 @@
     <q-alert v-if="data.fetchStatus.hasValidationErrors">
       {{ data.fetchStatus.validationErrors }}
     </q-alert>
-    <template v-if="hasLoaded">
+    <template>
```
tiltec: Element is unneeded

PR #1052, opened 2018-06-27
Improve group map

### LINKED ISSUE #902: Add Filtering options to full screen map
It would be nice to be able to filter the full screen map for users & stores, like the minimap on desktop.

(optional) Bonus points for adding filters for different store statuses. Not quite sure if that should be done in store or in the component though (?)

### LINKED ISSUE #975: Rework maps: group map, fullscreen map, gallery map
We have three types of map that fulfill different purposes, but could be improved. Here are some issues with the current implementation:

- Fullscreen map (which is the same as mobile map) doesn't have an option to show users and hide stores -> #902
    - When switching from group map to fullscreen map, it's unexpected for the user that the map view and map content changes -> state would need to be synchronized between them
- Group map and fullscreen map could show other groups in some way, e.g. as bubbles. This would bring it closer to the gallery map
- There is no way to find out the location of a store on mobile -> #863 
- At the hackathon in Göteborg, the fullscreen map was sometimes a bit "jumpy" on Chrome. Although the computer was also rather slow, it's good to keep in mind that not all computers are as fast as our dev machines ;)

tiltec:
Closes #902 
Closes #975 

Adds controls
Adds back button
Adds optional group markers (to see other groups)
Adds context menu as shortcut to create new store at a location
Fixes having too many marker shadows
Refactor marker definitions to one place (markers.js)

Controls:
image

Context menu:
image


src/components/Map/GroupMap.vue
```diff
@@ -103,23 +128,34 @@ export default {
...
+    },
+    selectedMarkers () {
+      if (this.selectedStore) {
+        const markers = []
+        if (hasLocation(this.selectedStore)) {
+          markers.push(storeMarker(this.selectedStore))
+        }
         if (this.showUsers) {
-          ids.push(...this.usersWithLocation.map(user => user.id).map(this.userMarkerId))
+          markers.push(...this.usersWithLocation.map(userMarker))
```
nicksellen: This is a duplication of marker creation as another one is created below in the `markers()` function. It would be unclear which is actually used without looking how they are used in the map. In this case it just pulls the marker id out later on.

An improvement might be to have a `selected` property on the marker, and remove the `selected-markers` prop to the map.

Issue #893, opened 2018-02-24
Add browser push notifications


nicksellen:
Simplest approach would be to use the existing FCM features in the backend to push to browsers. Ask me if you want to know more about how it works, can also refer to trustroots code for this https://github.com/Trustroots/trustroots (which I wrote).

peterkshultz:
Hi all! Cool that @jskollin and I take this on?

tiltec:
Hi @peterkshultz thanks for your initiative & sounds good to me! But I'm not an expert on this topic, so you may want to get in contact with @nicksellen about it. He just started to go on a cycling tour, so it might take a few days to respond :biking_man: ;)

Did you work with browser notifications and service workers before?

nicksellen:
I'm still here! Should have it internet access generally anyway.

PR #1057, opened 2018-06-29
Add browser push notifications

nicksellen:
Implements #893 

Also needs backend PR first https://github.com/yunity/karrot-backend/pull/556


nicksellen:
I fixed up some of crap (oops!). Currently there is a CORS error registering the token at a Google URL so needs a bit more work.

tiltec:
Throws some errors. When I click on the toggle:

image

The toggle stays enabled afterwards.
When I disable and enabled again, the browser asks me for permission. Then:

image

I ran a fresh `yarn dev` beforehand, so the service worker should have been built.

nicksellen:
Ah, the CORS error was just because of my browser plugin, push registration looks to be successful so far. Although the FCM_SENDER_ID was wrong, but I think the new way should have fixed it (I was passing it wrong in circleci).

From your screenshots, you don't have the latest code running, that error should be fixed now.

nicksellen:
Oh, and for your local one to work properly, you need to set the FCM_SENDER_ID env variable. I didn't really document that properly.

tiltec:
I found the sender ID on https://console.firebase.google.com/project/karrot-dev/settings/cloudmessaging/ios:world.karrot.dev .. let's see if it works!

nicksellen:
Ah, it works for me on the netlify preview now :)

tiltec:
Works on netlify too for me, but can't get it to work locally :(

PR #1081, opened 2018-08-16
exchange info and warning colors

djahnie:
## What does this PR do?

Exchanges info and warning colors without visible changes in the UI and only where it semantically makes sense. (One untreated occurrence is VerificationWarning.vue in which the warning color was already in use - it would have felt wrong to change it to info now...)


tiltec:
Looks good! The snapshot checks fail though, you need to update them with `yarn test -u`

PR #1085, opened 2018-08-20
Make all big round chat buttons tertiary

djahnie:
## What does this PR do?

Use tertiary as chat color, like we started doing in group-applications.

## Checklist

- [x] no unrelated changes

### DIFF
--- a/src/components/User/Profile.vue
+++ b/src/components/User/Profile.vue
@@ -69,7 +69,7 @@
             <q-btn
               small
               round
-              color="secondary"
+              color="tertiary"
               icon="fas fa-comments"
               class="hoverScale"
               @click="$emit('detail', user)"


tiltec:
I'm not sold on the idea that the chat button should have a special color. I can see that you had a conflict on the group application page (green being the accept color), but from my POV, it would be better to find a different solution there. 
For example, the chat button on the group application item could be removed and the whole box could be made clickable. Similar to how it's made in the pickup item, but with additional hover highlight to make it clearly visible that the item can be clicked.

My rationale for this: "Accept" or "Decline" are secondary actions, while clicking on the item is a primary action.

djahnie:
How would that hover highlight look like? The pickup chat is really hard to find, because people are not used to look for it. New and unexpected things need to be very obvious, otherwise they won't be found. I know that you want to have as few colors as possible, but to me having a button is the most salient solution I could think of.

PR #1086, opened 2018-08-20
Clear applications from state after leaving groupsGallery

djahnie:
## What does this PR do?

Prevents display of my own applications to other groups when entering from groupsGallery.

## Checklist

- [x] no unrelated changes


djahnie:
I'm unsure if it really behaves like it should now, but when I clicked around and tried to replicate the unwanted behavior I was unsuccessful... So maybe it works like this..?

src/store/modules/groupApplications.js
```diff
@@ -100,6 +103,9 @@ export default {
     update (state, application) {
       Vue.set(state.entries, application.id, application)
     },
+    clear (state) {
+      state.entries = {}
```
tiltec: We usually used the `initialState` function to clear the state of a module, for example here:
https://github.com/yunity/karrot-frontend/blob/1831610517ee3d7247c33df53999da8405f6eb42/src/store/modules/currentThread.js#L152

With this, we don't need to update the `clear` mutation when a new property gets added to initialState.

PR #1126, opened 2018-11-05
Switch back to webpack dev server

nicksellen:
## What does this PR do?

We/I tried out the alternative hip webpack tools but they are both deprecated now:
- https://github.com/webpack-contrib/webpack-serve
- https://github.com/webpack-contrib/webpack-command

This PR switches back to the establishment tools:
- https://github.com/webpack/webpack-dev-server
- https://github.com/webpack/webpack-cli


nicksellen:
I would merge soon @tiltec, seems to work ok :)

PR #1139, opened 2018-11-23
Add info button about user inactivity

djahnie:
## What does this PR do?
It puts an info button in the user list to explain what it means to be inactive.
I'm unsure actually if I put the right info in the dialogue - happy about input here! And also if the way I put the new translation strings into the locales is appreciated or not (the other way would have been to just put it into GROUP, but that's super huge already, soo....) 
(It's a tiny change that didn't break any test...)

## Links to related issues
No issue but two users asking independently for more detailed information about inactivity criteria.

## Checklist

- [x] no unrelated changes


src/locales/locale-en.json
```diff
@@ -514,5 +514,9 @@
     "HEADER_LINK": "foodsaving worldwide community",
     "LAST_UPDATED": "Last updated {relativeDate}"
   },
+  "INACTIVITY": {
+    "WHAT": "What does it mean to be inactive?",
+    "HELP": "Everyone who did not log into the group for 30 days or more counts as inactive. These users will not receive notification emails anymore."
```
tiltec:
```suggestion
    "HELP": "Everyone who did not visit the group for {dayCount} days or more counts as inactive. These users will not receive notification emails anymore."
```

- a group visit counts as activity
- make it parameterized so we can change the setting without invalidating the translation message

src/users/components/UserList.vue
```diff
@@ -92,6 +117,13 @@ export default {
       if (!this.filterTerm || this.filterTerm === '') return list
       return list.filter(u => u.displayName.toLowerCase().includes(this.filterTerm.toLowerCase()))
     },
+    inactivityInfo () {
+      Dialog.create({
+        title: this.$t('INACTIVITY.WHAT'),
+        message: this.$t('INACTIVITY.HELP'),
```
tiltec: Need to add `dayCount` as argument.

Actually, we should get the value from backend to keep settings in sync. We could add it to the group detail endpoint. It's probably a task for me then...

src/users/components/UserList.vue
```diff
@@ -18,12 +18,28 @@
...
       >
+        <template slot="header">
+          <q-item-side>
+            <q-item-tile icon="fas fa-bed" />
+          </q-item-side>
+          <q-item-main
+            :label="$t('GROUP.INACTIVE')"
+            :sublabel="inactiveSublabel"
+          />
+          <q-item-side>
```
tiltec: On very small screens, this can hide the "Inactive" label. Maybe reduce the default margin/padding on mobile to make it less likely.

src/locales/locale-en.json
djahnie: Yes, right, makes sense.

src/users/components/UserList.vue
djahnie: Yep. Tell me when it's there, then I'll change it. :)

src/users/components/UserList.vue
djahnie: I'm not sure I understand you correctly. You mean that the info button would hide the label? How small would that screen need to be..? o.0
I don't even know how to make that happen in mobile view of the browser, I couldn't find a device that small in the list... :sweat_smile:

tiltec: Depends on the language you chose, e.g. to hide the Polish word partially it needs 250 pixel. Maybe we just ignore it.

src/users/components/UserList.vue
tiltec: It's there! `group.memberInactiveAfterDays`

tiltec:
@djahnie you can click to merge :)

Issue #1068, opened 2018-07-25
Add Content Security Policies and set up reporting


tiltec:
@nicksellen proposed to set up some CSPs for karrot-frontend to enhance security. They consist of an HTTP header that we should probably configure in ansible and add a reporting service.

Documentation about CSP: https://developer.mozilla.org/en-US/docs/Web/HTTP/CSP
Our ansible playbooks without CSP: https://github.com/yunity/yuca
Using sentry as a CSP reporting service: https://docs.sentry.io/learn/security-policy-reporting/

My idea would be to start with report-only policies, to get better insight in the effects that an enforced policy would have. I already started a sentry project that could be used for it: https://sentry.io/foodsaving-worldwide/karrot-csp-report-only/

nicksellen:
Hmm, there are some fiddly complications. Mostly because I would like to get the git sha in the report-uri as the sentry "release". But because the ansible project is independent from the frontend deployment it does not have the git sha available. A few options:

1. deploy frontend via ansible in such a way we get the git sha available and can modify the nginx config - don't like this one so much as it's nice to deploy without having to modify nginx configuration and reload it
2. don't include the _release_ tag in the report-uri, so still put configuration in ansible project - seems release might be useful though so we don't chase old reports
3. have a fixed report-uri that goes to a url that redirects to the correct one (assuming that csp is ok with that):
  a. to a backend api url - but this doesn't know the frontend sha
  b. to an html page with a meta refresh redirect - this would allow it to be built by the frontend project
4. set the Content-Security-Policy using an html meta tag instead of an http header - not the recommended approach, but it certainly the easiest as we have all the information available at that point. One downside is internet explorer does not support it.

I think I'll go with 4 for now, as it's so simple.

nicksellen:
Oh, 4 is not possible as you cannot use a `report-uri` (or it's replacement `report-to`) in a `<meta>` tag.

https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/report-uri

nicksellen:
I progressed, but still a few decisions, and maybe some related tasks.

So I tried option 3c (not mentioned above) - which is to have a little nodejs proxy server that listens for `/csp-reports` then reads `about.json` and forwards it onto the right sentry url (including the release and environment).

[...]

PR #1141, opened 2018-11-23
Avoid eval

nicksellen:
## What does this PR do?

CSP is upset about eval, and apparantly if we make this change it'll be happy again :)

See https://github.com/webpack/webpack/issues/6461#issuecomment-394318550

## Links to related issues

#1068 

## Checklist

- [x] no unrelated changes

### DIFF
--- a/build/webpack.config.js
+++ b/build/webpack.config.js
@@ -70,6 +70,9 @@ module.exports = {
     chunkFilename: 'assets/js/[id].[chunkhash].js',
     pathinfo: false,
   },
+  node: {
+    global: false,
+  },
   resolve: {
     extensions: [
       '.mat.js', // for use inside quasar


build/webpack.config.js
```diff
@@ -70,6 +70,9 @@ module.exports = {
     chunkFilename: 'assets/js/[id].[chunkhash].js',
     pathinfo: false,
   },
+  node: {
```
tiltec: We could also try `node: false` and see if all dependencies cope with it.

nicksellen: Seems ok so far! It builds, and clicking around locally is ok. I'll push the change.

nicksellen:
Nope, has an error. On https://deploy-preview-1141--karrot.netlify.com/ when trying to edit my profile.

> ReferenceError: "process is not defined"

`vuelidate` tries to access `process.env.BUILD` in `./vuelidate/src/withParamsBrowser.js`

Seems they should fix it, see https://github.com/monterail/vuelidate/issues/365, but we can force it. I guess "web" is the correct value.

PR #1144, opened 2018-11-24
More csp fixes

nicksellen:
## What does this PR do?

Still showing errors for eval/inline. Have:
- set `node: false` for service worker script
- turned off html minification (as the csp report has a line number which isn't very useful when it's all on one line :) )
- specified netlify headers so hopefully can debug this in branches and don't have to merge to master each time

## Links to related issues

#1068

## Checklist

- [ ] no unrelated changes

### DIFF
--- a/build/webpack.config.js
+++ b/build/webpack.config.js
@@ -157,7 +157,7 @@ module.exports = {
     new HtmlWebpackPlugin({
       filename: 'index.html',
       template: 'src/base/index.html',
-      minify: true,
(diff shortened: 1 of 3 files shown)


nicksellen:
Firefox still complains, but doesn't seem to actually try sending a report.

@tiltec I suggest we merge this. And then maybe removing the "report-only" part for dev and see how it goes...

tiltec:
Do we need to turn html minification back on?

PR #1150, opened 2018-12-14
Fix Selecting many days in recurring pickups breaks layout

### LINKED ISSUE #1133: Selecting many days in recurring pickups breaks layout
When creating a new recurring pickup and selecting multiple days it can break the layout:

image

We should check if Quasar can handle this case. If yes, maybe we did something to break it and should fix it. If not, we should open an issue at their repository and perhaps implement a workaround.

adrianodantasit:
Closes #1133

## What does this PR do?

This pull request fixes the layout break when you select multiple days in a new recurring pickup.

I checked the Quasar Documentation, typed some extra text in the Select component example (as you can see in the image bellow) and could check they don't support yet text truncation in a Select component. So I added the CSS class "ellipsis" [Quasar Documentation Ellipsis](https://quasar-framework.org/components/visibility.html) in the Select component responsible for the text and the issue is fixed.

Typing a text longer than the container also breaks the layout in Quasar Documentation

## Checklist

- [ ] no unrelated changes

### DIFF
--- a/src/pickups/components/PickupSeriesEdit.vue
+++ b/src/pickups/components/PickupSeriesEdit.vue
@@ -45,6 +45,7 @@
             toggle
             v-model="byDay"
             :options="dayOptions"
+            class="ellipsis"
           />
         </q-field>
       </div>


tiltec:
Hey @adrianodantasit, thanks your initiative!

Before it looked like this:

image

Now it looks like this:

image

Seems an improvement to me! Would be cool to actually *see* the ellipsis `...` if that's possible!

Other than that, the tests fail because the snapshots aren't up-to-date. You could run `yarn test -u` to update them, or I could run it before merging.

---

I had a quick try and set the `white-space: normal` property to `.q-input-target`:

image

But I'm not sure if this is better - if somebody wants to see all details, they can just click on the input field.

adrianodantasit:
Thanks @tiltec , it's a big pleasure help somehow such a nice project. 

Well I inspected the web console and I realized what's going on.
There's a div which it bigger than its parent. Look:

Parent:
parent

Child:
child


That happened because the child div don't have a width in any of its classes, so I appended an extra class to the child only for control the width, setting it to width: 100% and this already fixed the text overlaying the arrow drop down. 

But it didn't show the ellipsis just doing that, I suspect it's because the text is inside a flex container without any child, so I removed the display: flex property from the container:
`<div class="col q-input-target ellipsis justify-start">Monday, Tuesday, Wednesday, Thursday, Friday, Saturday, Sunday</div>`

With those 2 steps I achieved a nice result check bellow:

final-result

So I think this is an issue in the Quasar framework generating this child bigger than its parent, what's your opinion? append an extra class just to control the width? open an issue in Quasar github?

tiltec:
Thanks for your praise! :smile:

Look what I found: https://github.com/quasarframework/quasar/issues/2795
Seems the problem only shows up in Firefox and should be fixed with the next Quasar release.

I would close this PR and https://github.com/yunity/karrot-frontend/issues/1133 and wait some days for a new release.

adrianodantasit:
Yeah I really meant it. :)
I've never used Vue so I'm kinda lost in this project but as soon as I get more familiar with it I'll help more.
I agree with close this PR.

One extra question, not sure if it's appropriate ask here so if not just close the PR.
Could you say some good references for study about tests? Like books, courses...

tiltec:
> Could you say some good references for study about tests? Like books, courses...

I should say I'm really not an expert. Mostly I write tests because I'm not sure if I implemented something correctly. Or there is a bug and I write a test before I fix it, to prevent regression.

Generally speaking, the [test pyramid](https://martinfowler.com/bliki/TestPyramid.html) gives good guidance, and [this is a much more in-depth article](https://martinfowler.com/articles/practical-test-pyramid.html).

Issue #1148, opened 2018-12-12
Keep up with new eslint-vue rules


tiltec:
By today's update, eslint-vue has some new rules to keep us busy (and make the codebase nicer).

I commented most of them out to not distract myself too much from #1147, so this could be fixed up:

https://github.com/yunity/karrot-frontend/blob/13f1c6bcb09ab1369b88c8cf0df4e20733ec6b64/.eslintrc.js#L44-L54

1. Comment out one exception
2. Run `yarn fix`
3. Check if it could be fixed automatically, otherwise edit files manually while having `yarn dev` running to get direct feedback
4. Repeat from (1)

djahnie:
@taistadam started with the first rule already.

tiltec:
Two rules have been applied, to-be-done ones can be found at the bottom of https://github.com/yunity/karrot-frontend/blob/master/.eslintrc.js

PR #1163, opened 2019-01-03
Tribute to the lint god: component-name-in-template-casing

taistadam:
Related to issue #1148 : solves component-name-in-template-casing rule by transforming all into PascalCase, with exception set for i18n component that the lint god wasn't so clever about.

- manual testing ok


lwm:
:fireworks: :fire: :fireworks:

PR #1178, opened 2019-01-07
Display pickups in progress to its collectors

### LINKED ISSUE #1172: Delay pickup and conversations dissapearance by ~30 minutes
> feature request: the pick ups and the conversations disappear exactly at the pick up time, Is it possible to make it half an hour llater, so people can still use the chat if somebody is late and also read the describtion? that would be great!

Just pushing this one out from the #karrot-dev channel.

djahnie:
Closes #1172 

Displays pickups 30 minutes after their occurring time _for those who are signed up to it_.
The chat is still reachable then.
The member cannot leave the pickup anymore after time of pickup.
Designs of hovering and empty slots not perfect but quite fiddly... probably good enough as it is :)

## Checklist

- [ ] needs more manual testing
- [x] no unrelated changes

### DIFF
--- a/src/pickups/components/CurrentUser.vue
+++ b/src/pickups/components/CurrentUser.vue
@@ -2,13 +2,15 @@
   <a
     @click.stop="$emit('leave')"
     class="user-slot-wrapper"
+    :pickup="pickup"
     :style="{ width: size + 'px', height: size + 'px' }"
-    :title="$t('PICKUPLIST.ITEM.LEAVE')"
   >
     <div class="hoverShow">
       <i
+        v-if="!inProgress(pickup)"
         :style="{fontSize: (size - 9) + 'px'}"
         class="fas fa-fw fa-times"
+        :title="$t('PICKUPLIST.ITEM.LEAVE')"
       />
     </div>
     <div class="hoverHide">
@@ -34,11 +36,19 @@ export default {
       type: Object,
       required: true,
     },
+    pickup: {
+      type: Object,
+      required: true,
+    },
   },
   components: {
     ProfilePicture,
   },
-
+  methods: {
+    inProgress (pickup) {
+      return (pickup.date <= new Date())
+    },
+  },
 }
 </script>
 
--- a/src/pickups/components/CurrentUser.vue
+++ b/src/pickups/components/CurrentUser.vue
@@ -2,12 +2,11 @@
   <a
     @click.stop="$emit('leave')"
     class="user-slot-wrapper"
-    :pickup="pickup"
     :style="{ width: size + 'px', height: size + 'px' }"
   >
     <div class="hoverShow">
       <i
-        v-if="!inProgress(pickup)"
+        v-if="!pickup.hasStarted"
         :style="{fontSize: (size - 9) + 'px'}"
         class="fas fa-fw fa-times"
         :title="$t('PICKUPLIST.ITEM.LEAVE')"
@@ -43,11 +42,6 @@ export default {
   },
   components: {
     ProfilePicture,
-  },
-  methods: {
-    inProgress (pickup) {
-      return (pickup.date <= new Date())
-    },
   },
 }
 </script>
--- a/src/pickups/components/PickupItem.vue
+++ b/src/pickups/components/PickupItem.vue
@@ -33,10 +33,10 @@
           <b class="text-negative">{{ $t('PICKUPLIST.PICKUP_DISABLED') }}</b>
         </div>
         <div
-          v-if="inProgress(pickup)"
+          v-if="pickup.hasStarted"
           class="q-my-xs"
         >
-          <b class="text-orange">{{ $t('PICKUPLIST.PICKUP_ONGOING') }}</b>
+          <b class="text-orange">{{ $t('PICKUPLIST.PICKUP_STARTED') }}</b>
         </div>
         <div
           class="q-my-xs multiline"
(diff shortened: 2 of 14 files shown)


src/pickups/components/CurrentUser.vue
```diff
@@ -2,13 +2,15 @@
   <a
     @click.stop="$emit('leave')"
     class="user-slot-wrapper"
+    :pickup="pickup"
```
tiltec: not needed

src/pickups/components/CurrentUser.vue
```diff
@@ -34,11 +36,19 @@ export default {
...
+      type: Object,
+      required: true,
+    },
   },
   components: {
     ProfilePicture,
   },
-
+  methods: {
+    inProgress (pickup) {
```
tiltec: if you need this in other places as well, makes sense to move it into the `enrich` getter

could be renamed to `ongoing` to unify with locale message

src/pickups/components/PickupItem.vue
```diff
@@ -32,6 +32,12 @@
         >
           <b class="text-negative">{{ $t('PICKUPLIST.PICKUP_DISABLED') }}</b>
         </div>
+        <div
+          v-if="inProgress(pickup)"
```
tiltec: use enriched property

src/pickups/components/PickupItem.vue
```diff
@@ -87,20 +93,25 @@ export default {
...
     leave () {
-      Dialog.create({
-        title: this.$t('PICKUPLIST.ITEM.LEAVE_CONFIRMATION_HEADER'),
-        message: this.$t('PICKUPLIST.ITEM.LEAVE_CONFIRMATION_TEXT'),
-        ok: this.$t('BUTTON.YES'),
-        cancel: this.$t('BUTTON.CANCEL'),
-      })
-        .then(() => this.$emit('leave', this.pickup.id))
-        .catch(() => {})
+      if (this.pickup.date > new Date()) {
```
tiltec: `!pickup.ongoing`

src/pickups/datastore/pickups.js
```diff
@@ -36,8 +37,9 @@ export default {
     },
     upcoming: (state, getters) => {
```
tiltec: better rename to something like `upcomingAndRecentlyStarted`

tiltec: ...or split into a separate getter `started`

taistadam:
still need to:
- [ ] settings to include previous 30 minutes pickups in both API and datastore
- [ ] remove started pickups from PickupsManage component

src/pickups/datastore/pickups.js
```diff
@@ -1,11 +1,22 @@
...
 import pickups from '@/pickups/api/pickups'
 import { createMetaModule, withMeta, isValidationError, withPrefixedIdMeta, metaStatusesWithId, metaStatuses } from '@/utils/datastore/helpers'
+import { pickupRunningTime } from '@/pickups/settings'
+import subMinutes from 'date-fns/sub_minutes'
 
 function initialState () {
   return {
     now: new Date(), // reactive current time
-    entries: {},
+    entries: {
```
tiltec: test data should be removed

src/pickups/datastore/pickups.js
```diff
@@ -31,13 +42,15 @@ export default {
         group,
         collectors: pickup.collectorIds.map(rootGetters['users/get']),
         feedbackGivenBy: pickup.feedbackGivenBy ? pickup.feedbackGivenBy.map(rootGetters['users/get']) : [],
+        hasStarted: pickup.date <= new Date(),
```
tiltec: This is not fully reactive, you need to compare with `state.now` instead of `new Date()`.

src/pickups/datastore/pickups.js
```diff
@@ -31,13 +42,15 @@ export default {
...
         ...metaStatusesWithId(getters, ['save', 'join', 'leave'], pickup.id),
       }
     },
     upcoming: (state, getters) => {
       return Object.values(state.entries)
-        .filter(p => p.date >= state.now)
+        .filter(p => p.date >= subMinutes(state.now, pickupRunningTime))
         .map(getters.enrich)
+        .filter(p => p.date >= state.now || (p.isUserMember && p.date <= state.now))
```
tiltec: you could use `p.hasStarted` here to make it easier to understand:

```suggestion
        .filter(p => !p.hasStarted || (p.isUserMember && p.hasStarted))
```

src/pickups/datastore/pickups.js
tiltec: This is not really resolved, `upcoming` still contains started pickups. Would propose the rename, seems simpler at this stage (compared to having a separate getter).

src/pickups/datastore/pickups.js
```diff
@@ -47,7 +60,9 @@ export default {
       return getters.byCurrentGroup.filter(({ store }) => store && store.isActiveStore)
     },
     byActiveStoreOneTime: (state, getters) => {
-      return getters.byActiveStore.filter(e => !e.series)
+      return getters.byActiveStore
+        .filter(e => !e.series)
+        .filter(p => !p.hasStarted)
```
tiltec: A comment could be helpful to understand why started pickups are only filtered at this stage and not elsewhere.

src/pickups/datastore/pickups.js
djahnie: Ah right! We forgot...

Issue #1110, opened 2018-10-08
Unify and simplify vuex modules


tiltec:
In the last months, I noticed some inconsistencies and unnecessary complexity growing in our vuex modules:

- `set` mutations sometimes replace, sometimes updates state. They should always replace state.
- `update` mutations sometimes accept one, sometimes multiple entries. They should always handle multiple entries if the module itself handles multiple ones.
- filtering by scope (group, store, user etc) should always be done in getters. If there's stale data or performance problems, filtering can additionally be done in actions.
- it seems easier to perform sorting in getters instead of mutations. So far, we did sorting of messages and history entries in mutations, but maybe doing it in getters simplifies code without noticable performance loss.
- [x] avoid pass-through actions, instead use mutations directly
- [x] some duplication could be removed, e.g. multiple places have a reactive current time (pickups, notifications, DateAsWords although this should stay independent of vuex). 

There's also at least one outright bug:
- PaginationMixin: if two backend requests are pending, the last one that succeeds defines the cursor.  This could cause unwanted effects if the scope of the list changed (e.g. history switched between two stores). We should either take care to cancel unneeded requests or handle one cursor per scope.

Related issues:
- https://github.com/yunity/karrot-frontend/issues/1091 talks about message store only
- https://github.com/yunity/karrot-frontend/issues/1066 might mean moving vuex module files around

I'll have a deeper look at this during the next months...

tiltec:
This is mostly solved, except for this

> PaginationMixin: if two backend requests are pending, the last one that succeeds defines the cursor. This could cause unwanted effects if the scope of the list changed (e.g. history switched between two stores). We should either take care to cancel unneeded requests or handle one cursor per scope.

But as nobody complained about it so far and it can be simply solved by reloading the page, I will close this now.

PR #1184, opened 2019-01-08
Rename to currentStore and currentUser

tiltec:
In the spirit of https://github.com/yunity/karrot-frontend/issues/1110

Goal is to unify with `currentGroup`.

- activeStore, selectedStore -> currentStore
- activeUser, selectedUser -> currentUser


tiltec:
currentUser probably conflicts with the logged in user (although the site works fine in manual testing).
A different name could be useful... _loggedInUser_? _authUser_?

PR #1195, opened 2019-01-11
Add unsubscribe view

### LINKED ISSUE #1174: Allow to mute conversations without being logged in
I got reports from users who had problems with muting some of their conversations, because they weren't logged in and either had to log in or even reset their password.

The frontend has some logic in place to execute the _mute_ action after login: https://github.com/yunity/karrot-frontend/blob/fd5c7785dc1ce2af51d001af0b08a6d249e7bb64/src/authuser/datastore/auth.js#L52
But it probably breaks if they have to reset their password.

It would be cool to make it easier for the users to receive less emails. We can use Django's signing tool as [in the incoming-email feature](https://github.com/yunity/karrot-backend/blob/aad44ebf48c1d84412d65a29298184a52ca539eb/foodsaving/webhooks/api.py#L16) or [our VerificationCode](https://github.com/yunity/karrot-backend/blob/f0e69964df464de79460dd817eaef62be7388ff0/foodsaving/userauth/models.py#L19) to send out a link that allows people to trigger this action, without being logged in.

I lean toward using Django's signing tool, as it doesn't need to keep state in the database. I would encode the user id, conversation id and maybe the type of action, to allow adding more logged-out actions in future. There's also a [signer that can limit the valid duration](https://docs.djangoproject.com/en/2.1/topics/signing/#verifying-timestamped-values).

More details yet to be figured out!

nicksellen:
Closes #1174 

## What does this PR do?

Adds a view for unsubscribing from notifications. The link is only available in the footer of emails, see https://github.com/yunity/karrot-backend/pull/618 for the backend details of how this works.

## Checklist

- [x] no unrelated changes

### DIFF
--- a/package.json
+++ b/package.json
@@ -44,6 +44,7 @@
   "dependencies": {
     "@sentry/browser": "^4.3.0",
     "axios": "^0.18.0",
+    "base62": "^2.0.0",
     "clone-deep": "4.0.1",
     "date-fns": "^1.29.0",
     "deep-equal": "^1.0.1",
--- a/package.json
+++ b/package.json
@@ -44,7 +44,6 @@
   "dependencies": {
     "@sentry/browser": "^4.3.0",
     "axios": "^0.18.0",
-    "base62": "^2.0.0",
     "clone-deep": "4.0.1",
     "date-fns": "^1.29.0",
     "deep-equal": "^1.0.1",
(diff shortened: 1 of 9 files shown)


package.json
```diff
@@ -44,6 +44,7 @@
   "dependencies": {
     "@sentry/browser": "^4.3.0",
     "axios": "^0.18.0",
+    "base62": "^2.0.0",
```
tiltec: can be removed, as the decoded timestamp is not used anymore

nicksellen:
I removed the timestamp decoding now, if it needs to be done again it is available here ---> https://github.com/yunity/karrot-frontend/blob/7a18f0f6c54e9981ee22c135aea29a30191111b4/src/unsubscribe/utils.js

Issue #1111, opened 2018-10-08
When creating or editing feedback, weight number is off the screen


tiltec:
On the PickupFeedback form (can be accessed with http://localhost:8080/#/group/1/give-feedback and similar), when entering a weight, the weight number is weirdly moved off the screen:

![image](https://user-images.githubusercontent.com/4410802/46615978-d7351000-cb19-11e8-92f4-03ae7c152a0d.png)

It should be shown inside the form field.

farmanp:
Came here to say this is a pattern in design called "content jumping" - will try and take a look how to discontinue the issue

stevenkuipers:
Hi, 

if it's okay with you, I would like to take this.

tiltec:
Hi @stevenkuipers, sounds good! I'm happy to answer questions about this :)

stevenkuipers:
Hi @tiltec, I was not able to reproduce the bug where the number moves offscreen. Can you clarify what browser you were / are using?

PR #1221, opened 2019-01-21
Wrapper for images does not overflow images to stop content…

stevenkuipers:
Wrapper for images does not overflow images to stop content below it from jumping around. Also added a transition to the images.

## What does this PR do?

As mentioned in #1111, using the slider on a screen that's not so wide makes the content jump around because the images keep getting added and removed. This is a small ui improvement to counter this jarring effect.

## Links to related issues

https://github.com/yunity/karrot-frontend/issues/1111

## Checklist

- [ ] no unrelated changes

### DIFF
--- a/src/feedback/components/AmountPicker.vue
+++ b/src/feedback/components/AmountPicker.vue
@@ -18,7 +18,6 @@
           :amount="value"
         />
         <div
-          style="margin-left: .6em"
           class="col"
         >
           <div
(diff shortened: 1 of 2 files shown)


tiltec:
Revisiting this: I would choose `@change` instead of `@input` for the QSlider component. That updates the value only when releasing the mouse button. It is simpler than a transition and less jarring too.

PR #1229, opened 2019-01-23
Mark conversations as seen when opening the latestMessages menu

tiltec:
## What does this PR do?

It helps if users don't read all of their conversations to still get updates about new messages.

## Checklist

- [x] no unrelated changes

### DIFF
--- a/src/messages/api/conversations.js
+++ b/src/messages/api/conversations.js
@@ -38,6 +38,10 @@ export default {
       muted,
     })).data)
   },
+
+  async markAllSeen () {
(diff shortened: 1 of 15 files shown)


src/messages/api/conversations.js
```diff
@@ -38,6 +38,10 @@ export default {
       muted,
     })).data)
   },
+
+  async markMenuSeen () {
+    return (await axios.post(`/api/conversations/mark_menu_seen/`)).data
```
nicksellen: I would keep the concept of the "menu" out of the API.

PR #1291, opened 2019-02-12
Conflict resolution feature

### LINKED ISSUE #853: [Brainstorming] Process to remove user from group
Joining and leaving a group on karrot is easy, but only if the user actively does it. I think there might be two reasons where removal for other reasons seems valid: inactivity and kick request.

- due to inactivity: notify and mark as inactive after 1 month (grey out in member list), notify again after 2 months, remove from group after 3 months
- due to kick request: anybody can ask to remove another person from the group, using a process.

## Process proposal

1. Requester describes reasons why X should be removed from group
2. X can answer
3. After one day, all members from the group are added to the conversation and can vote if X should be removed from group
    - options: "yes", "no", "don't care"
    - additional yes/no selection: "Do you have enough information?" (_control proposal_)
    - voting period: until 50% of active group members voted, plus 3 days (arbitrary, to give everyone a chance to vote). Immediately ends if 100% active group members voted.
    - personal decision can be changed during voting period
    - results are only shown when vote is over (to prevent strategic voting and bias)
    - if less than 50% of active group members voted after 2 weeks, process is stopped and archived
4. Everybody can write comments into the conversation
5. Result is the comparison of "yes" and "no" votes
    - if more people say they don't have enough information than people saying "yes" or "no", then the process continues until it is not the case anymore (_extended voting period_), but only to a maximum of 2 weeks. Then the process gets archived, without removal
    - if "yes" < "no", process gets archived, without removal
    - if "yes" > "no", X gets removed as a group member

While the process is ongoing, X keeps being a member of the group but temporarily gets most rights revoked (to prevent revenge actions; only join/leave pickups and leave the group keep active). During the _extended voting period_, rights are reverted to how it was before the removal process (to avoid blocking someone without a decision).

## Further comments / questions

- Does this work with open groups and password-protected groups? The removed user could just join again. I guess it would need some kind of _approval process_ before joining the group
- Would it be reasonable to prevent new members (joined the group less than 2 weeks ago) from voting?
- Is it harmful if somebody wants to troll an active user by starting the removal process against them? The rights would get removed until to process is over (worst case for 2 weeks)
- The process is very complex. How can it be simplified without much chance for harm?
- The server admins always have full power and need to be trusted. Wouldn't it be OK to have trusted admins with almost full power per group? Most people are honest, should we really make process that complex for the few cases where people aren't?
- Can the process be generalized for other group decisions and would this add enough value to karrot to warrant the complexity?

## Mockup

I made a mockup of some parts of the process, you can visit the page here: https://karrot-remove-proposal.netlify.com/

Screenshots:

image
image
image

djahnie:
Closes #853 

## What does this PR do?
- will fill in later...

## Checklist

- [ ] no unrelated changes


src/issues/components/IssueResults.vue
```diff
@@ -0,0 +1,137 @@
...
+  },
+  props: {
+    issue: {
+      type: Object,
+      required: true,
+    },
+  },
+  computed: {
+    lastDecision () {
+      return this.issue.votings.find(v => max(v.expiredAt))
```
tiltec: I think this is not doing what it should do. [max](https://date-fns.org/v1.9.0/docs/max) expects an array and returns the latest date. `find` will give you the first list entry with a `expiredAt` set, so essentially it's the same as `this.issue.votings[0]`.

We want to find the latest date, so we could sort the array by expiredAt and take the first entry:

```
votings.slice().sort((a, b) => b.expiredAt - a.expiredAt)[0]
```
(I hope the sort order is right...)

Although my code would fail if there are no votings. That should not happen, but to failproof our code, we could add a check before.

src/issues/components/IssueVote.vue
```diff
@@ -0,0 +1,184 @@
...
+  data () {
+    return {
+      edit: null,
+    }
+  },
+  watch: {
+    options: {
+      immediate: true,
+      handler (current, previous) {
+        /*
```
tiltec: I _think_ it should be fine without the commented code. 
It means that the users' pending changes will get overwritten when a change comes in (e.g. a websocket message), so we need to check that case.

src/issues/components/IssueVote.vue
```diff
@@ -0,0 +1,184 @@
...
+        this.edit = cloneDeep(current)
+      },
+    },
+  },
+  computed: {
+    days () {
+      return distanceInWordsStrict(this.ongoingVoting.expiresAt, new Date())
+    },
+    progress () {
+      return 100 - (differenceInHours(this.ongoingVoting.expiresAt, new Date()) / 168 * 100)
```
tiltec: What is `168`?

src/issues/components/IssueVote.vue
```diff
@@ -0,0 +1,184 @@
...
+    },
+    options () {
+      return this.ongoingVoting.options
+    },
+  },
+  methods: {
+    getLabel (score) {
+      switch (score) {
+        case 0:
+          return 'neutral'
```
tiltec: These should be translatable

src/issues/components/IssueVote.vue
```diff
@@ -0,0 +1,184 @@
...
+    },
+    getTitle (type) {
+      return this.$t(`ISSUE.VOTING.${type.toUpperCase()}`, { userName: this.issue.affectedUser.displayName, groupName: this.issue.group.name })
+    },
+  },
+}
+</script>
+
+<style scoped lang="stylus">
+@import '~slidetoggle'
```
tiltec: this seems unused in here, probably can be removed

src/issues/datastore/issues.js
```diff
@@ -0,0 +1,86 @@
...
+      commit('saveScores', data)
+    },
+  },
+  mutations: {
+    setCurrentIssue (state, issueId) {
+      state.currentId = issueId
+    },
+    saveScores (state, results) {
+      for (let i = 0; i < results.length; i++) {
+        Vue.set(state.entries[state.currentId].votings[0].options[i], 'yourScore', results[i].score)
```
tiltec: This might address the wrong voting & wrong option: `votings[0].options[i]`

I'd be tempted to remove this logic and rely on websockets, but if we keep it, this should be fixed:
- the first voting is not always the current one
- don't rely on the sorting of options, rather access by option id

src/messages/datastore/detail.js
```diff
@@ -5,7 +5,7 @@ import router from '@/base/router'
 
 function initialState () {
   return {
-    scope: { type: null, id: null },
+    scope: {},
```
tiltec: I wonder why this has been changed.... maybe just revert?

src/notifications/components/Notifications.story.js
```diff
@@ -16,6 +16,17 @@ const place = factories.makePlace({
   group,
 })
 
+// TODO move this to enrichedFactories
+let issueIdCnt = 0
```
tiltec: Should be moved/merged with existing factories (was me)

src/users/components/ProfileUI.vue
```diff
@@ -62,6 +62,14 @@
           icon="fas fa-comments"
           @click="$emit('detail', user)"
         />
+        <QBtn
+          v-if="triggerCriteria"
```
tiltec: rename to `canReport`?

I think it would be nice to keep the button but explain _why_ the user can't be reported. maybe via tooltip (bad for mobile) or dialog.

src/users/pages/Profile.vue
```diff
@@ -3,6 +3,7 @@
     <Profile
       :user="user"
       :current-group="currentGroup"
+      @report="report"
```
tiltec: As the ProfileUI component is now tied to `vuex`, we could merge `Profile.vue` and `ProfileUI.vue` -> later?

Not so nice for storybook though, so maybe we can entangle it again...? TrustButton is also tied to the datastore however...

test/enrichedFactories.js
```diff
@@ -169,6 +172,80 @@ export const makePickupSeries = data => {
...
 }
 
+let votingIdCnt = 0
+export const makeVoting = data => {
+  return {
+    id: votingIdCnt++,
+    acceptedOption: 74,
+    expiresAt: addHours(new Date(), 106),
+    options: [{
+      affectedUser: 174,
```
tiltec: affectedUser should be null here

test/enrichedFactories.js
```diff
@@ -169,6 +172,80 @@ export const makePickupSeries = data => {
   }
 }
 
+let votingIdCnt = 0
+export const makeVoting = data => {
+  return {
+    id: votingIdCnt++,
+    acceptedOption: 74,
+    expiresAt: addHours(new Date(), 106),
+    options: [{
```
tiltec: we should add a `makeOption` factory to abstract a bit over this

test/enrichedFactories.js
```diff
@@ -169,6 +172,80 @@ export const makePickupSeries = data => {
...
+      affectedUser: 174,
+      id: 73,
+      sumScore: 7.0,
+      message: null,
```
tiltec: I removed the `message` field from the API just now, so that could be removed too

src/applications/components/ApplicationItem.vue
```diff
@@ -42,7 +42,7 @@
           >
             <RouterLink
               place="userName"
-              @click.native.stop=""
+              @click.native.stop
```
tiltec: Need to check if that still does the job. I think the goal was to _not_ open the application when clicking on the person who decided it.

src/base/datastore.js
```diff
@@ -45,6 +46,7 @@ const appModules = [
   topbar,
   users,
   utils,
+  issues,
```
tiltec: Should be sorted by a recognized alphabet ;)

src/notifications/components/notificationConfig.js
```diff
@@ -82,6 +97,14 @@ function getRouteTo (type, { group, user, place, pickup } = {}) {
...
     case 'pickup_moved':
       return group && place && pickup && { name: 'pickupDetail', params: { groupId: group.id, placeId: place.id, pickupId: pickup.id } }
+    case 'conflict_resolution_created':
+    case 'conflict_resolution_created_about_you':
+    case 'conflict_resolution_continued':
+    case 'conflict_resolution_continued_about_you':
+    case 'conflict_resolution_decided':
+    case 'conflict_resolution_decided_about_you':
+    case 'voting_ends_soon':
+      return group && issue && { name: 'conflictResolution', params: { groupId: group.id, issueId: issue.id } }
```
tiltec: I think the name should be `issueTabs` or so

src/base/routes/main.js
```diff
@@ -113,6 +115,34 @@ export default [
         },
         component: GroupWall,
       },
+      {
+        name: 'issueTabs',
```
tiltec: Rename to `issue`?

src/issues/api/issues.js
```diff
@@ -0,0 +1,45 @@
...
+  async vote (id, listOfVotes) {
+    await axios.post(`/api/issues/${id}/vote/`, listOfVotes)
+  },
+
+  async deleteVote (id) {
+    await axios.delete(`/api/issues/${id}/vote/`)
+  },
+}
+
+export function convert (val) {
```
tiltec: Needs to be extended to converting votings (and options if needed)

src/issues/components/ConflictSetupTitle.vue
```diff
@@ -0,0 +1,6 @@
+<template>
+  <h1>Start a bloody conflict</h1>
```
tiltec: should be translatable

src/issues/components/IssueListUI.vue
```diff
@@ -0,0 +1,76 @@
+<template>
+  <div class="bg-white">
```
tiltec: Maybe remove that div and have `<QList class="bg-white">`?

src/issues/components/IssueTabsUI.vue
```diff
@@ -0,0 +1,153 @@
...
+                :user="issue.affectedUser"
+                :size="25"
+                :is-link="true"
+                class="q-mt-sm"
+              />
+              {{ issue.affectedUser.displayName }}
+            </div>
+          </div>
+          <div class="q-ma-sm q-pa-sm bg-white">
+            <span class="text-bold text-secondary uppercase">{{ issue.createdBy.displayName }}</span>
```
tiltec: The user name could be a link to the profile page

src/issues/components/IssueTabsUI.vue
```diff
@@ -0,0 +1,153 @@
...
+      type: Object,
+      default: null,
+    },
+  },
+  computed: {
+    multipleVotings () {
+      return this.issue.votings.length > 1
+    },
+    pastVotings () {
+      return this.issue.votings.filter(v => new Date(v.expiresAt) <= new Date())
```
tiltec: `new Date(...)` can be removed when we extended the `convert` function in the api module

src/issues/components/IssueVote.vue
```diff
@@ -0,0 +1,184 @@
...
+  QProgress,
+  QBtn,
+  QIcon,
+  QTooltip,
+} from 'quasar'
+
+import distanceInWordsStrict from 'date-fns/distance_in_words_strict'
+import differenceInHours from 'date-fns/difference_in_hours'
+import cloneDeep from 'clone-deep'
+// import { objectDiff } from '@/utils/utils'
```
tiltec: Can be removed

src/locales/locale-en.json
```diff
@@ -541,6 +551,48 @@
     "WHAT": "What does it mean to be inactive?",
     "HELP": "Everyone who did not log into the group for {dayCount} days or more counts as inactive. These users will not receive notification emails anymore."
   },
+  "CONFLICT": {
+    "SETUP": "Conflict resolution setup",
+    "STATEMENT": "Statement",
+    "TITLE": "Conflict resolution",
+    "STEPPER1": "It is important to openly address conflicts and you obviously know this. People like you are the ones that keep a community healthy, that is highly appreciated!",
+    "STEPPER2": "You are about to start a group-wide conflict process! This will involve all of your group's editors to try and find the best way to proceed regarding this conflict.",
+    "STEPPER3": "This process will start an open chat and a vote. Every message in the chat will be visible for every editor in the group - including the ppp! The voting will last for 7 days during which all voters will still be able to change their minds and vote differently as many times as they want.",
```
tiltec: "The voting will last for 7 days" -> should be parameterized, I'll add the value to the group endpoint

tiltec:
How about moving the IssueTabs component into the detail sidebar, to resolve the problems with chat and make the PR mergable?

nicksellen:
> How about moving the IssueTabs component into the detail sidebar, to resolve the problems with chat and make the PR mergable?

and the list stays on the main bit of the page?

I think mergable is higher priority than "rethink all of UI" :) so would be in favour of a "shortcut"

tiltec:
> and the list stays on the main bit of the page?

Yes.

Need to see how it will look, might need to make some modifications to the right drawer:

- make it wider, maybe even responsive
- don't occupy too much vertical space with issue tab navigation, at least on mobile
- change the route when opening an issue on desktop

src/issues/components/IssueHistoryItem.vue
```diff
@@ -0,0 +1,97 @@
...
+      </QItemMain>
+      <QItemSide
+        class="desktop-only"
+        stamp
+        right
+      >
+        <DateAsWords :date="issue.createdAt" />
+      </QItemSide>
+    </QItem>
+    <Transition name="slide-toggle">
```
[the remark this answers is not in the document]
djahnie: I'd keep them here because in case one issue has many votings the page would be unecessarily long and cluttered if all details were immediately shown. Also I think the transitions are quite fancy...^^

src/issues/components/IssueHistoryItem.vue
```diff
@@ -0,0 +1,97 @@
+<template>
+  <div>
+    <QItem
```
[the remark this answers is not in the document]
djahnie: You mean to substitute the transition for a collapsible? Hm, could work. But is it important to change this?

src/issues/components/ConflictSetupTitle.vue
djahnie: Yeah, and should be a text we actually want to have in there...^^

src/issues/components/IssueHistoryItem.vue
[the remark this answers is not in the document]
tiltec: It's only about the transitions really. Not the biggest fan of them myself.

The collapse seems fine, could be made into a QCollapsible though.

src/issues/components/IssueTabsUI.vue
djahnie: ...which is not yet, right?

tiltec:
The detail drawer doesn't have the compose field always visible either, the user has to scroll down to type in a message. Insofar it's similar to having the chat on a main page. 

I think the main problem is having a chat _and_ related content on the same page. If related content is always relevant to the user, it should not be in the same scrollable container as the messages, but in the header.
Applications have the initial messages in the same scrollable container, but our assumption was that the user starts reading there and then scrolls down to write their own input. I wonder if the same should be true for issues?

src/issues/components/VotingResults.vue
```diff
@@ -0,0 +1,127 @@
...
+      </div>
+      <div
+        class="q-pt-xs"
+      >
+        {{ $t('ISSUE.VOTING.RESULTS.PARTICIPANTS', { number: voting.participantCount }) }}
+      </div>
+      <QItemSide
+        v-if="!isCancelled"
+        right
+        stamp="Total score"
```
tiltec: Needs translation

test/enrichedFactories.js
```diff
@@ -169,6 +172,81 @@ export const makePickupSeries = data => {
...
+    type: 'conflictResolution',
+    createdBy: makeUser(),
+    affectedUser: makeUser(),
+    group: makeGroup(),
+    status: 'ongoing',
+    isCancelled: false,
+    isOngoing: true,
+    votings: [
+      makeVoting({
+        expiresAt: new Date('2018-01-24T21:22:54.730980Z'),
```
tiltec: date could be generalized

src/issues/datastore/issues.js
```diff
@@ -0,0 +1,109 @@
...
+        dispatch('detail/clear', null, { root: true })
+
+        await dispatch('maybeFetchOne', issueId)
+        dispatch('conversations/fetchForIssue', { issueId }, { root: true })
+        commit('setCurrentIssue', issueId)
+      },
+    }, {
+      findId: ({ issueId }) => issueId,
+    }),
+    async saveScores ({ commit, dispatch, state }, data) {
```
tiltec: wrap this into withMeta and show pending/error states in UI

tiltec:
We had some talks already about allowing conflict resolution in the playground group, so that users can try it out. It's currently not possible because the playground group is an open group and for open groups it didn't make sense to me to allow conflict resolution. 

But it actually might make sense: a group decision might be powerful enough to make someone leave, even if they could just join again.

nicksellen:
Would be handy to have had a _how to try it out_ description, I fired up my local development environment, but then was a bit lost how to proceed. I can work it out though.

I did something like this to get enough editors:
```
g = Group.objects.get(pk=5)
g.groupmembership_set.update(roles=['editor'])
```

src/issues/pages/IssueChat.vue
```diff
@@ -0,0 +1,139 @@
...
+    <div
+      v-if="conversation"
+      class="bg-secondary absolute-top-right round-borders q-pa-xs"
+    >
+      <NotificationToggle
+        :muted="conversation.muted"
+        :is-participant="conversation.isParticipant"
+        :user="currentUser"
+        in-toolbar
+        @set.stop="setNotifications"
```
nicksellen: the notification toggle does not work as using `.stop` on a custom event is does not work (throws an exception that `stopPropagation` is not a function)

src/issues/pages/IssueChat.vue
```diff
@@ -0,0 +1,139 @@
...
+        :size="$q.platform.is.mobile ? 'sm' : 'md'"
+      />
```
nicksellen: Also, setting notifications to off means I leave the conversation, but I cannot rejoin it again :/

I think it would be reasonable to be a member of the conversation but have all notifications off (distinct from not being a member of the conversation - but with same notification behaviour).

src/base/routes/main.js
```diff
@@ -102,6 +110,12 @@ export default [
           name: 'subheader',
         },
       }) },
+      detail: { render: h => h('router-view', {
+        props: {
+          name: 'detail',
+        },
+      }) },
+      footer: { render: h => h('router-view', { props: { name: 'footer' } }) },
```
nicksellen: Now it's used so much, maybe we can encapsulate this pattern into a function:

```js
{ render: h => h('router-view', { props: { name: 'subheader' } }) }
```

could become:

```js
routerView('subheader')
```

(and a `mobileRouterView` function for that case...)

src/issues/components/IssueListUI.vue
```diff
@@ -0,0 +1,85 @@
+<template>
+  <QList
+    no-border
+    class="bg-white"
+  >
+    <KSpinner v-show="isPending" />
+    <IssueItem
+      v-for="i in ongoingIssues"
+      :key="i.id"
+      :issue="i"
```
nicksellen: `i` is normally for a numeric loop index, `issue` would be clearer

src/issues/components/IssueTabs.vue
```diff
@@ -0,0 +1,42 @@
...
+} from 'quasar'
+
+export default {
+  components: {
+    QTabs,
+    QRouteTab,
+  },
+  computed: {
+    ...mapState({
+      groupId: state => state.currentGroup.id,
```
nicksellen: This is available via `currentGroup/id` getter

src/issues/components/IssueVote.vue
```diff
@@ -0,0 +1,201 @@
...
+        <div>
+          <div
+            v-for="o in edit"
+            :key="o.id"
+          >
+            <div class="q-mt-md">
+              {{ getTitle(o.type) }}
+            </div>
+            <QSlider
+              style="width: 85%; margin: 0 auto"
```
nicksellen: Random place to find some style, can go into style section?

src/issues/components/IssueVote.vue
```diff
@@ -0,0 +1,201 @@
...
+  data () {
+    return {
+      edit: null,
+    }
+  },
+  watch: {
+    options: {
+      immediate: true,
+      handler (current, previous) {
+        this.edit = cloneDeep(current)
```
nicksellen: Not really needed, but can do a fancy thing and only get the diff of the changes, so keep local changes.

src/users/pages/Profile.vue
```diff
@@ -21,38 +165,155 @@
...
+      this.clearIssueMeta(['create'])
+    },
   },
   computed: {
     ...mapGetters({
       user: 'users/activeUser',
       currentGroup: 'currentGroup/value',
       history: 'history/byCurrentGroupAndUser',
       historyStatus: 'history/fetchStatus',
+      getIssues: 'issues/ongoing',
```
nicksellen: `issues` or `ongoingIssues` would be a better name, the _get_ is implicit

src/users/pages/Profile.vue
```diff
@@ -21,38 +165,155 @@
...
+      return this.getIssues.find(i => i.affectedUser.id === this.user.id)
+    },
+    isConflictOngoing () {
+      return Boolean(this.ongoingConflict)
+    },
+    profilePictureSize () {
+      if (this.$q.platform.is.mobile) {
+        return 80
+      }
+      return 180
```
nicksellen: A ternary operator would be much shorter

src/locales/locale-en.json
```diff
@@ -563,6 +573,71 @@
...
+    "PAST": "Past issues",
+    "PREVIOUS_VOTINGS": "Previous votings in this issue",
+    "PARTICIPANTS": "{count} members have been added to this discussion.",
+    "VOTING": {
+      "TIME_LEFT": "Time left to vote",
+      "BTN_START": "Vote now",
+      "BTN_DELETE": "Delete vote",
+      "TOAST": "Your vote was successfully submitted",
+      "TOAST_DELETE": "Your vote was deleted",
+      "NO_CHANGE": "{userName} stays in {groupName}",
```
nicksellen: I would wonder about prefixing this with _No change:_ as it's a kind of _special_ status quo option.

src/issues/pages/IssueChat.vue
```diff
@@ -0,0 +1,139 @@
...
+        </span>
+        <Markdown :source="issue.topic" />
+      </div>
+      <div
+        v-if="conversation"
+        class="q-ml-sm q-pt-sm q-pl-sm"
+      >
+        <div>
+          <ProfilePicture
+            v-for="user in conversation.participants"
```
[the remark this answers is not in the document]
tiltec: Yeah, could be potentially large, even though it's only all group _editors_.

src/base/pages/MainLayout.vue
```diff
@@ -239,6 +255,11 @@ export default {
       }
       return this.windowWidth > 1000 ? 380 : 280
     },
+    detailWidth () {
+      const contentWidth = this.windowWidth - this.sidenavWidth
+      const columnWidth = Math.floor(contentWidth / 2)
```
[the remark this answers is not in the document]
tiltec: It's not really a sidebar anymore, more like columns. But yeah, a max width could make sense. On the other hand, why should the list of issues be extremely wide then?

src/base/routes/main.js
tiltec: I tried that, but it didn't work, some parts of the page didn't show in some circumstances. Would leave that for another refactor.

src/issues/components/IssueTabs.vue
tiltec: It's not the same:
- `state.currentGroup.id` is set when the group _started_ loading
- `currentGroup/id` is set _after_ the group finished loading

I used this here to prevent vue-router warnings during page load. Another option would be to set `:to="null"` when the group and issue are not loaded yet.

src/base/pages/MainLayout.vue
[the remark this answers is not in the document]
nicksellen: Max width makes sense for columns too, and the list of issues has it's inner container with a constrained width already (just centred).

src/issues/components/IssueTabs.vue
nicksellen: Perhaps the solution it to make the `currentGroup/id` getter return the same as `state.currentGroup.id` - nothing about either of those ways of getting the id would inform me of that difference currently, whereas they _seem_ like they would be the same.

In general would like to avoid depending on nuances of load ordering.

src/issues/components/VotingResults.vue
```diff
@@ -0,0 +1,127 @@
...
+        <QItemMain>
+          {{ label }}
+        </QItemMain>
+        <QItemSide stamp>
+          {{ sumScore }}
+        </QItemSide>
+      </QItem>
+    </template>
+    <QItem
+      v-else
```
djahnie: A tie seems to also count as isCancelled and thus triggers the undecided messages:
image

I think it would be cooler to have a separate message for ties as well. Maybe we compare the sumScores and make another computed property to be able to have one more special case in the template?

src/issues/pages/IssueChat.vue
tiltec: As a short-term fix, I'll remove the "off" option for issue chats. 

I'm less motivated right now to implement yet another notification state, but maybe it's necessary at some point.

tiltec: Ah, I guess I should really fix this in the backend... let's see what the best option is.

src/locales/locale-en.json
```diff
@@ -452,7 +454,15 @@
     "PICKUP_UPCOMING": "Reminder: You have a pickup at {time}",
     "PICKUP_DISABLED": "Pickup at {dateTime} was disabled",
     "PICKUP_ENABLED": "Pickup at {dateTime} was enabled again",
-    "PICKUP_MOVED": "Pickup was moved to {dateTime}"
+    "PICKUP_MOVED": "Pickup was moved to {dateTime}",
+    "CONFLICT_RESOLUTION_CREATED": "Join the discussion about a conflict with {userName}",
+    "CONFLICT_RESOLUTION_CREATED_ABOUT_YOU": "A conflict resolution process was started with you",
```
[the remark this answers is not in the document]
tiltec: I'll adjust it in the email template.

src/issues/components/VotingResults.vue
tiltec: Found it! Historic votings received the status of the issue, but they are not affected by it. `isCancelled` should always be `false` for historic votings.

PR #1292, opened 2019-02-12
Place wall & subscriptions

### LINKED ISSUE #986: Place conversations & subscriptions
It was requested to have the possibility to subscribe to stores. Store subscribers would then be the only ones who–
- receive pickup notifications for that specific store
- are part of the store conversation (#649 )

The default could be that everyone is subscribed to every store to not have users be confused that they don't receive info anymore.

tiltec:
Closes #986 
Needs backend PR https://github.com/yunity/karrot-backend/pull/662

## What does this PR do?

- add place wall
  - add "place" type to latest messages entries
  - add notification chips to sidenav
- add place subscription toggle
  - show subscribed places first in sidenav (even if not "active")
- add "unsubscribe" option to notification toggle (and remake it completely)
  - cannot unsubscribe from private conversations and thread replies
  - don't show message unread status if user is not subscribed
- filter available pickups by subscribed stores and add a info for the user

Pictures are in https://github.com/yunity/karrot-frontend/issues/986#issuecomment-462466800

To do:
- walls of archived places are hidden right now, maybe need to show them in a special mode
- store wall looks very similar to group wall, how can it be less confusing?
- new users won't be subscribed to any store and don't see pickups on the wall - maybe that's fine? but it's not easy to discover...

## Checklist

- [ ] no unrelated changes


src/messages/components/DetailHeaderUI.vue
```diff
@@ -204,17 +215,17 @@ export default {
     },
   },
   methods: {
-    toggleNotifications () {
+    toggleNotifications (value) {
```
nicksellen: `setNotifications`

src/places/components/PlaceList.vue
```diff
@@ -97,6 +119,11 @@ export default {
 
   },
   computed: {
+    sortedPlaces () {
+      const subscribed = this.places.filter(e => e.isSubscribed)
+      const notSubscribed = this.places.filter(e => !e.isSubscribed)
+      return subscribed.concat(notSubscribed)
```
nicksellen: Can use a sort.

tiltec: I had a try and felt that using `sort` is actually less readable, at least how I wrote it. Also `sort` is not necessary stable in JS, so we would take care not to lose the previous alphabetical sorting. Would leave as is.

PR #1293, opened 2019-02-12
Move footer information to a modal

nicksellen:
Infinite scroll and footers doesn't make much sense together and now we want a chat on the main page with the input box at the bottom, so seems to make sense to dump all this about karrot info elsewhere :)

Also included in https://github.com/yunity/karrot-frontend/pull/1291


nicksellen:
Ah, I removed the `<UnsupportedBrowserWarning/>` but didn't add it back anywhere...

tiltec:
Looks good!

I think you can remove `KFooter.vue` as it's not used.

What about `GLOBAL.FSWW_NOTE` ("Read our info page"), should we remove it from `locale-en.json`?

src/base/pages/MainLayout.vue
```diff
@@ -60,6 +60,20 @@
           <SidenavTitle @click="toggleSidenav" />
           <RouterView name="sidenav" />
           <MobileSidenav/>
+          <QItem
+            link
+            @click.native="toggleAbout()"
+          >
+            <QItemSide class="text-center">
+              <KarrotLogo
+                disable-loading="true"
```
tiltec: > type check failed for prop "disableLoading". Expected Boolean, got String with value "true"

src/topbar/components/KTopbarUI.vue
```diff
@@ -11,7 +11,10 @@
         :src="currentGroup.photoUrls.thumbnail"
         style="height: 95%"
       >
-      <KarrotLogo v-else />
+      <KarrotLogo
+        v-else
+        :show-loading="true"
```
tiltec: Just `show-loading` should be enough!

nicksellen: Ah yeah, I wasn't sure if that suddenly made it a directive... will update :)

PR #1300, opened 2019-02-15
Rejig page loading

tiltec:
## What does this PR do?

Make page loading more responsive and more robust, especially with rapid user interactions.

- withMeta: abort outdated vuex actions (instead of preventing new ones with "already pending")
- load page data _after_ confirming the route, add "loading" indicator on some pages

The idea is to allow another request to the same _kind_ of resource if another one is already pending (e.g. two places) and try to abort the first one. I implemented this by wrapping the vuex `commit` and `dispatch` actions that are being passed into withMeta actions. Inspired by https://github.com/sebdiem/vuex-cancellable-actions

withMeta will work like this:
1. if a request to the same resource (e.g. place 1) is still pending, prevent the new one
   - but it allows _throttled_ retries: if the previous request was started more than 1 second ago and is still pending, allow the new request
2. set the _id_ of the resource as _current id_, e.g. `activePlaceId = 1`
3. for every commit/dispatch the action does, check if `activePlaceId` is `1`. If not, throw `ActionAbortedError` and catch it in the action wrapper - we have aborted the action!

This has some caveats: any other action side effect than `commit` or `dispatch` will still execute and downstream actions can't get aborted because they receive an unwrapped `dispatch`.

Some pages might not handle it well yet to be shown without data. I should add some kind of loading indicator to them.

## Checklist


tiltec:
I did a little benchmark - loading the group wall page 3 times with Chromium performance tools

With the rejig it seems a little slower: 4.5%, or 3.29s vs 3.44s. That seems reasonable, as it needs to repaint more often. I guess the effect might be bigger on slower devices.

On the upside with this change the page should feel _faster_ to the user, as the Karrot logo is shown for much shorter time.

I'm not entirely convinced about my implementation, but I would be in favor of merging it nevertheless to try it out with a bigger developer/user base and to progress this approach further.
I think the general direction taken is good: 
- load data after route confirmation
- let the components handle loading states
- extend `withMeta` to handle data loading logic

PR #1375, opened 2019-03-30
Adds an edit field for the group welcome message that is sent to newcomers.

### LINKED ISSUE #1038: Group-specific welcome emails
Teddy wrote in Slack:

> I send out a seperate welcome mail to the new members with explanations (because i know not everybody reads the describtions). Is there a way to develop a welcom message that every member gets automatically?

Bruno added:

> The welcome message could be basically the same feature as the terms and conditions (but without the "I accept" tickbox) that was discussed earlier? That would also be cool for us

cstefanj:
This implements the frontend part of #1038.

Closes #1038

### DIFF
--- a/src/__snapshots__/storyshots.spec.js.snap
+++ b/src/__snapshots__/storyshots.spec.js.snap
@@ -1335,6 +1335,55 @@ It would make sense if it was markdown
             </div>
           </div>
         </div>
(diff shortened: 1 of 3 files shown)


tiltec:
Merged in https://github.com/yunity/karrot-frontend/commit/fbcbf6db34bc36bf268803e2aeccb9b834057735

PR #1421, opened 2019-04-15
Emoji picker search bar

### LINKED ISSUE #1072: Implement extended emoji picker for reactions
Right now, we hardcoded a selection of 5 in the frontend:

image

It would be nice if people could choose from more emoji when giving reactions. 

Our implementation already supports all of [twemoji](https://twitter.github.io/twemoji/2/test/preview-svg.html), so the only thing that's left is a nice way to select emoji. A few commonly used emoji should still be easy to access, and a full text search (in emoji names) could be added to aid emoji power users :)

https://github.com/missive/emoji-mart/ seems the best open-source implementation out there, but it might be a bit too heavy - and it's only available for React.
https://github.com/jm-david/emoji-mart-vue exists for Vue, but the last time @mrkvon tried it there were some problems (I remember that it used non-standard emoji names).

Some time ago, I started an implementation based on simple text input: https://github.com/yunity/karrot-frontend/compare/emoji-autocomplete
Maybe it could be used as a starting point.

andreseg9726:
Closes #1072 

## What does this PR do?

Adds a search bar to the emoji picker so users can choose a wider range of emoji. 5 additional emoji were hardcoded on the default options. 

The emoji names are looked up in the following [library](https://github.com/markdown-it/markdown-it-emoji/blob/master/lib/data/full.json) and later parsed with [twemoji](https://twitter.github.io/twemoji/) to get the image. If the emoji exists in the library, but twemoji can't parse it, the carrot emoji is shown.

## Checklist

- [x] no unrelated changes

### DIFF
--- a/src/messages/components/ConversationAddReaction.vue
+++ b/src/messages/components/ConversationAddReaction.vue
@@ -9,11 +9,13 @@
       @show="open = true"
       @hide="open = false"
     >
-      <ConversationAddReactionInner
-        v-if="open"
-        :reacted="reacted"
-        @toggle="$emit('toggle', arguments[0])"
-      />
+      <div style="width: 195px">
+        <ConversationAddReactionInner
+          v-if="open"
+          :reacted="reacted"
+          @toggle="$emit('toggle', arguments[0])"
+        />
+      </div>
     </QPopover>
   </QBtn>
 </template>
--- a/src/messages/components/ConversationAddReaction.vue
+++ b/src/messages/components/ConversationAddReaction.vue
@@ -7,15 +7,11 @@
       anchor="top left"
       self="top left"
       @show="open = true"
-      @hide="open = false, search = ''"
+      @hide="open = false"
     >
-      <QSearch
-        v-model="search"
-      />
       <ConversationAddReactionInner
         v-if="open"
         :reacted="reacted"
-        :search="search"
         @toggle="$emit('toggle', arguments[0])"
       />
     </QPopover>
@@ -23,12 +19,12 @@
 </template>
 
 <script>
-import { QBtn, QPopover, QSearch } from 'quasar'
+import { QBtn, QPopover } from 'quasar'
 
 const ConversationAddReactionInner = () => import('./ConversationAddReactionInner')
 
 export default {
-  components: { QBtn, QPopover, QSearch, ConversationAddReactionInner },
+  components: { QBtn, QPopover, ConversationAddReactionInner },
   props: {
     opacity: {
       type: Number,
@@ -42,7 +38,6 @@ export default {
   data () {
     return {
       open: false,
-      search: '',
     }
   },
 }
--- a/src/messages/components/ConversationAddReaction.vue
+++ b/src/messages/components/ConversationAddReaction.vue
@@ -9,10 +9,14 @@
       @show="open = true"
       @hide="open = false"
     >
-      <QSearch v-model="search" />
+      <p>{{ search }}</p>
+      <QSearch
+        v-model="search"
+      />
       <ConversationAddReactionInner
         v-if="open"
         :reacted="reacted"
+        :search="search"
         @toggle="$emit('toggle', arguments[0])"
       />
     </QPopover>
@@ -39,6 +43,7 @@ export default {
   data () {
     return {
       open: false,
+      search: '',
     }
   },
 }
--- a/src/messages/components/ConversationAddReaction.vue
+++ b/src/messages/components/ConversationAddReaction.vue
@@ -9,6 +9,7 @@
       @show="open = true"
       @hide="open = false"
     >
+      <QSearch v-model="search" />
       <ConversationAddReactionInner
         v-if="open"
         :reacted="reacted"
@@ -19,12 +20,12 @@
 </template>
 
 <script>
-import { QBtn, QPopover } from 'quasar'
+import { QBtn, QPopover, QSearch } from 'quasar'
 
 const ConversationAddReactionInner = () => import('./ConversationAddReactionInner')
 
 export default {
-  components: { QBtn, QPopover, ConversationAddReactionInner },
+  components: { QBtn, QPopover, QSearch, ConversationAddReactionInner },
   props: {
     opacity: {
       type: Number,
--- a/src/messages/components/ConversationAddReaction.vue
+++ b/src/messages/components/ConversationAddReaction.vue
@@ -7,9 +7,8 @@
       anchor="top left"
       self="top left"
       @show="open = true"
-      @hide="open = false"
+      @hide="open = false, search = ''"
     >
-      <p>{{ search }}</p>
       <QSearch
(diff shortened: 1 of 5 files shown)


src/messages/components/ConversationAddReaction.vue
```diff
@@ -7,24 +7,28 @@
       anchor="top left"
       self="top left"
       @show="open = true"
-      @hide="open = false"
+      @hide="open = false, search = ''"
     >
+      <QSearch
```
tiltec: Can you move this inside `ConversationAddReactionInner`? Then the search term should be cleared implicitly when the popover gets closed (because of `v-if="open"`).

src/messages/components/ConversationAddReactionInner.vue
```diff
@@ -17,6 +17,19 @@
 
 <script>
 import EmojiButton from './EmojiButton'
+import emojiList from 'markdown-it-emoji/lib/data/full.json'
+
+function searchEmoji (search) {
+  // clean search by removing colons and setting lowercase
+  let cleanedSearch = search.replace(/:/g, '').toLowerCase()
+  // only considers exact match searches
+  if (Object.keys(emojiList).includes(cleanedSearch)) {
```
tiltec: I would also show partial matches, it makes it a bit more complex but could be a improvement for users.

Some ideas:
- show partial matches exact ones
- only start the search with more than 2 characters (or if there's an exact match, thinking of `+1`)
- limit number of results (20?)
- show the emoji names to users, then they have an idea why the emoji matched the search

What do you think about it?

src/messages/components/EmojiButton.vue
```diff
@@ -22,7 +22,13 @@ function getEmojiElement (name) {
   const cached = EMOJI_CACHE[name]
   if (cached) return EMOJI_CACHE[name].cloneNode(true)
   const container = document.createElement('div')
-  container.innerHTML = twemoji.parse(emojiList[name])
+  if (Object.keys(emojiList).includes(name)) {
+    container.innerHTML = twemoji.parse(emojiList[name])
+  }
+  // On error, we display the 'carrot' emoji
```
tiltec: You can remove this exception again, I'll fix it in the backend before merging this PR.

src/messages/components/ConversationAddReaction.vue
andreseg9726: Sure, I can do that.

src/messages/components/ConversationAddReactionInner.vue
andreseg9726: I agree with the first three bullets and will start working on it! 
About showing the names, I think maybe since the names already pop up once the user hovers over the emoji, it might be overkill to display all the names. If you want I could try to explicitly display the name of the emoji that is currently hovered at the bottom of the popover.

src/messages/components/EmojiButton.vue
andreseg9726: Will do!

src/messages/components/ConversationAddReactionInner.vue
tiltec: > it might be overkill to display all the names

Agree, let's try without for now!

I found another challenge: the list from `markdown-it-emoji` contains duplicates that _could_ show up next to each other in the results. For example, searching for `boat` matches `:boat:` and `:sailboat:`, which both show ⛵️. Ideally we would only show one of them.

andreseg9726: Added some changes to address that!

src/messages/components/ConversationAddReactionInner.vue
```diff
@@ -17,24 +20,73 @@
...
+  // First add the exact search in case there is an emoji shortcode with less
+  // than three characters
+  if (Object.keys(emojiList).includes(cleanedSearch)) {
+    matchingEmojiUnicodes.add(emojiList[cleanedSearch])
+    matchingEmojiNames.push(cleanedSearch)
+  }
+
+  // Checks the length because we don't want to have too many results
+  if (cleanedSearch.length > 2) {
+    for (var index in Object.keys(emojiList)) {
```
tiltec: Since you don't need the index, what about using `emojiList.forEach()`?

src/messages/components/ConversationAddReactionInner.vue
```diff
@@ -1,8 +1,11 @@
 <template>
   <div
     class="row"
-    style="margin: 2px 4px"
+    style="margin: 2px 4px; width: 185px"
```
tiltec: As we are trying to minimize inline styles: can you move this to the style section?

src/messages/components/ConversationAddReactionInner.vue
```diff
@@ -17,24 +20,73 @@
 
 <script>
 import EmojiButton from './EmojiButton'
+import emojiList from 'markdown-it-emoji/lib/data/full.json'
+import { QSearch } from 'quasar'
+
+function searchEmoji (search) {
+  // Remove colons and set to lowercase to normalize search
+  let cleanedSearch = search.replace(/:/g, '').toLowerCase()
```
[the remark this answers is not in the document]
andreseg9726: Done. Thanks for letting me know.

src/messages/components/ConversationAddReactionInner.vue
```diff
@@ -17,24 +20,73 @@
...
+  }
+
+  // Checks the length because we don't want to have too many results
+  if (cleanedSearch.length > 2) {
+    for (var index in Object.keys(emojiList)) {
+      // This regular expression does a substring search of emoji names to
+      // find more interesting results
+      let regexp = '.*' + cleanedSearch + '.*'
+      let emojiShortcode = Object.keys(emojiList)[index]
+      if (emojiShortcode.match(regexp) &&
```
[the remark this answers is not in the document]
andreseg9726: You are right! Just did that.

src/messages/components/ConversationAddReactionInner.vue
andreseg9726: `forEach()` didn't let us return when the emoji matches reached 20, but I scrapped using the index anyways by looping over `emojiList` instead of over `Object.keys(emojiList)`

src/messages/components/ConversationAddReactionInner.vue
andreseg9726: Thanks, I moved it. Changed some numbers to allow 5 emoji in each row.

andreseg9726:
Okay just pushed new commits! 
About the scrollbars, I've been having some trouble controlling the popovers. They don't always behave the same way even if I change the `anchor` and `self` properties and I'm not sure why. Maybe the scrollbars appeared since I had hardcoded some width values? I couldn't replicate that but it might be fixed now that I set the `max width` instead of `width`?

andreseg9726:
I tried adding two simple tests. Let me know if the format is correct!

Issue #1297, opened 2019-02-14
Place subscriptions & conversations - follow-up


tiltec:
From the PRs that resolved https://github.com/yunity/karrot-frontend/issues/986:

* walls of archived places are hidden right now, maybe need to show them in a special mode. backend could "close" conversations of archived stores
* store wall looks very similar to group wall, how can it be less confusing?
* new users won't be subscribed to any store and don't see pickups on the wall - maybe that's fine? but it's not easy to discover...
* "star" icon in sidenav is not very nicely aligned, could replace the status dot?

tiltec:
Frontend bug: if the place has no wall messages and the user is not subscribed yet, it doesn't show the wall at all.

brnsolikyl:
It looks good! I'm more concerned about the navigation issues you mentioned: 

- finding store-specific pickups comes new with an extra step; and 
- it might not be clear which wall they're at...

The first point might not be a big deal, given the subscription option. On the other hand, if pickups from  unsubscribed stores are not easily discoverable, that's a major drawback for non-popular stores. This something to be tested after a while in production.

Could the second point be solved by making it possible for users to add a banner to each store?

tiltec:
> Could the second point be solved by making it possible for users to add a banner to each store?

Possibly yes, although it would be cool to have something that doesn't require user interaction. For example, we could add shopping cart to the top of the page.

That makes me thinking if the breadcrumbs in the topbar are actually useful.. isn't their purpose to give the user orientation on which page they are on?

tiltec:
From https://community.foodsaving.world/t/introducing-store-wall-and-store-subscriptions/241/12

- [ ] show people who starred a place
- [ ] try out "star" button in sidenav, as shortcut
- [ ] try out tab navigation in place header (to replace sidenav section)

tiltec:
We had discussions how we can make group members more aware of place messages when they are not subscribed. Ideas I can remember:
- show a dot if there were recent messages - but it can't be marked as read
- as before, but track read status even for unsubscribed users
- as before, but allow users to disable the "dot" as well (per place or per group?)

Seems a lot of effort to me to implement this right, but could also be very useful!

tiltec:
[...]

PR #1432, opened 2019-04-16
Tab navigation for places

tiltec:
## What does this PR do?

- add tab navigation header to all place pages
- always show place name in topbar breadcrumbs
- move place description into collapsible
- remove all sidenav toggles

## Links to related issues

Related to #1297 

## Checklist

- [ ] no unrelated changes


tiltec:
I worked on it a bit more and now I'm quite happy with it!

Screenshot_2019-04-23_18-25-07
Screenshot_2019-04-23_18-26-09
Screenshot_2019-04-23_18-25-41

tiltec:
From Bruno

> Nice! I like the expandable description, for descriptions that are too long. Although at a first glimpse it took me some time to notice the tabs. Maybe we could highlight them in a different color?

Although it could be useful to reduce the persistent colors on the website a bit more (e.g. make it less intense), to have some reserves if we actually want to notify people.

It wasn't visible in the previous screenshots, but the wall tab shows the unread messages:

Screenshot_2019-04-24_12-38-38

This would need to be different if the whole bar had a color, making it a bit less obvious:

Screenshot_2019-04-24_12-39-30

We could also choose a different color for the text only, but I'm not sure if this is any better than the original solution.

Screenshot_2019-04-24_12-42-34

brnsolikyl:
Great with the little notification of messages there! I don't have any strong opinion about the colour actually. Your reasoning makes sense too. Karrot is already quite colourful :smile:

PR #1449, opened 2019-04-24
Experiment: public places

tiltec:
Experiment: would it be useful if places could be set to "public", showing their name, location and public description to non-members?

- added a markdown field `public_description`
- if the markdown field is not blank, return public place info via `/api/places-info/`
- show public places in group gallery and group preview page

Screenshot_2019-04-24_11-23-30


tiltec:
From Bruno:

> Public places: Use case - one (or many) channel(s) for a group to make public announcements (changing opening hours, invitation to events, etc.) and drawing more attention to Karrot for potential volunteers.

This is lower priority.

PR #1724, opened 2019-09-05
Add vision text to repository

djahnie:
## What does this PR do?

Adds in Bruno's vision text [as found on the pad](https://pad.disroot.org/p/karrotvision).


djahnie:
Now that I read the whole text in one go again I'm unsure of the roadmap part for two reasons:
- the word 'roadmap' is already used for something else
- the plan outlined there makes the vision text time sensitive

I wonder if it would be better to employ a different term and/or remove the whole practical part from VISION.md and move it somewhere else (maybe MISSION.md? :wink:).

brnsolikyl:
I think you're right about not having the second part on the Vision text, @djahnie . But is it a mission? (maybe it is, I'm always confused by this word). Anyhow there are some practical steps there that could be eventually added to the official roadmap

PR #1790, opened 2019-10-09
General refactoring

raymonshansen:
Refactor into smaller functions for easier debugging. Flake 8 compliant.
Haven't tested, but should be functionally equivalent to previous version.
* Moved static strings into global variables in case they change
* Smaller functions
* Fixed shadowing of 'config' variable
* Separate global from local imports
* Imports are specific

### DIFF
--- a/updateLocalesHelper/updateLocaleStatus.py
+++ b/updateLocalesHelper/updateLocaleStatus.py
@@ -1,59 +1,92 @@
 #! /bin/env python
-import json
+from configparser import ConfigParser
+from http import HTTPStatus
+from json import dump
+from os import path, environ
+from requests import get as GET
+import sys
+
 from txclib import utils
 from txclib import config
-import requests
-import os
-import sys
-import configparser
 
-script_path = os.path.dirname(os.path.realpath(sys.argv[0]))
 
-status_file = os.path.join(script_path, '../src/locales/translationStatus.json')
-config_file = os.path.join(script_path, '../.tx/config')
+STATUS_FILE_PATH = '../src/locales/translationStatus.json'
+CONFIG_FILE_PATH = '../.tx/config'
+SCRIPT_PATH = path.dirname(path.realpath(sys.argv[0]))
+TRANSLATION_URL = 'https://api.transifex.com/organizations/yunity-1/projects/karrot/'
+DEFAULT_TOKEN_URL = 'https://www.transifex.com'
+
+STATUS_FILE = path.join(SCRIPT_PATH, STATUS_FILE_PATH)
+CONFIG_FILE = path.join(SCRIPT_PATH, CONFIG_FILE_PATH)
+
 
 def get_token():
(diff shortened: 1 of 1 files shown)


updateLocalesHelper/updateLocaleStatus.py
```diff
@@ -1,59 +1,92 @@
 #! /bin/env python
-import json
+from configparser import ConfigParser
+from http import HTTPStatus
+from json import dump
```
tiltec: I prefer to `import json`, because it's clearer to read `json.dump` than just `dump`. Also, that's how they do it in the examples: https://docs.python.org/3.7/library/json.html

updateLocalesHelper/updateLocaleStatus.py
```diff
@@ -1,59 +1,92 @@
...
+from os import path, environ
+from requests import get as GET
```
tiltec: same here, I prefer to `import requests` and then `requests.get` - much more similar to their examples: https://requests.kennethreitz.org/en/master/user/quickstart/#make-a-request

updateLocalesHelper/updateLocaleStatus.py
```diff
@@ -1,59 +1,92 @@
 #! /bin/env python
-import json
+from configparser import ConfigParser
+from http import HTTPStatus
+from json import dump
+from os import path, environ
```
tiltec: Same here, keeping the context by using `os.environ' seems widespread to me: https://stackoverflow.com/questions/4906977/how-to-access-environment-variable-values

updateLocalesHelper/updateLocaleStatus.py
```diff
@@ -1,59 +1,92 @@
...
-lang_map = {}
-for arg in lang_map_raw.replace(' ', '').split(','):
-    k, v = arg.split(":")
-    lang_map.update({k: v})
-
-# translate to HTTP Accept-Language format
-def translate_key(key):
+def get_translation_status(token):
+    response = GET(TRANSLATION_URL, auth=('api', token))
+    if response.status_code != HTTPStatus.OK:
```
tiltec: As we are using `requests` here, might use their abstraction: https://requests.kennethreitz.org/en/master/user/quickstart/#response-status-codes

```suggestion
    if response.status_code != requests.codes.ok:
```

raymonshansen:
Totally agree with all of the above! 
I think I fixed all of them.
(I might have been in the mood you get when people do import numpy and then use numpy arrays to store 10 ints...)

tiltec:
@raymonshansen the script fails with: 
```
Traceback (most recent call last):
  File "updateLocalesHelper/updateLocaleStatus.py", line 83, in <module>
    translation_status = get_translation_status(token)
  File "updateLocalesHelper/updateLocaleStatus.py", line 34, in get_translation_status
    response = request.get(TRANSLATION_URL, auth=('api', token))
NameError: name 'request' is not defined
```

PR #1811, opened 2019-10-16
Use cross-env for script env variables

nicksellen:
Should allow better windows support

## What does this PR do?

Uses [cross-env](https://www.npmjs.com/package/cross-env) to set script environment variables in a cross platform way.

## Checklist

- [ ] no unrelated changes


tiltec:
Cool!

But there might still be a problem in the `jest` case, as we need to tell `node` to use UTC as time zone. That seems impossible on Windows without changing the global system time zone: https://github.com/nodejs/node/issues/4230#issuecomment-224093946
There's https://github.com/capaj/set-tz, but:
> On windows, when you call setTZ() it effectively changes your whole OS timezone so beware-if you kill the process, you'll need to manually revert back after you're done-I am assuming you'r process is short-lived. If you don't kill your process, it will revert your original timezone back for you.

nicksellen:
Yup, I guess that's something in addition to this.

That is quite a caveat for `set-tz` I think. Maybe merge this one as this needs fixing anyway as windows is interpreting `TZ` as a command...

PR #1867, opened 2019-11-09
Remove agreement frontend

tiltec:
It has never been used because we didn't figure out the permission handling.
The frontend implementation is most likely broken due to the Quasar upgrade.

Related to https://github.com/yunity/karrot-frontend/issues/324


tiltec:
I talked with @nicksellen, he wants to find out if users would still find this feature useful. Closing for now.

PR #1878, opened 2019-11-12
Use address autocomplete

nicksellen:
## What does this PR do?

Use autocomplete for address picker again as it used to. I made it so it gives you an explicit option to use the search term as the address without adding/changing any existing map marker.

Also defaults place edit map to be group centre if available.

I didn't add everything into translation strings, just as maybe there are some language changes in review (plus, I'm tired and about to go to bed ;) )

## Checklist

- [ ] no unrelated changes

### DIFF
--- a/src/maps/components/AddressPicker.vue
+++ b/src/maps/components/AddressPicker.vue
@@ -1,61 +1,59 @@
 <template>
   <div>
-    <div>
-      <QSelect
-        use-input
-        fill-input
-        clearable
-        hide-dropdown-icon
-        :label="label"
-        :error="error"
-        :error-message="errorMessage"
-        :options="options"
-        :placeholder="$t('BUTTON.SEARCH')"
-        :value="value.address"
-        :input-debounce="1000"
-        @input="input"
-        @filter="search"
-      >
-        <template #before>
-          <QIcon :name="icon" />
-        </template>
-        <template #selected-item />
-        <template #option="{ index, itemProps, itemEvents, opt: { label: itemLabel, useSearchTerm } }">
-          <QItem
-            :key="index"
-            v-bind="itemProps"
-            v-on="itemEvents"
-          >
-            <QItemSection>
-              <QItemLabel v-if="useSearchTerm">
-                Set address as
-                <strong>{{ itemLabel }}</strong>
-                <span v-if="hasLocation">(keep existing location on map)</span>
-                <span v-else>(choose location manually on map)</span>
-              </QItemLabel>
-              <QItemLabel v-else>
-                {{ itemLabel }}
-              </QItemLabel>
-            </QItemSection>
-          </QItem>
-          <QSeparator v-if="useSearchTerm" />
-          <QItemLabel
-            v-if="useSearchTerm"
-            v-t="'GLOBAL.SEARCH_RESULTS'"
-            header
-          />
-          <QItem
-            v-if="useSearchTerm && options.length === 1"
-          >
-            <QItemSection>
-              <QItemLabel>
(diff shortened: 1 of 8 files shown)


src/maps/components/AddressPicker.vue
```diff
@@ -192,7 +197,7 @@ export default {
 <style scoped lang="stylus">
 .map
   height 260px
-  margin-left 42px
+  margin-left 37px
```
nicksellen: This makes it align nicely for place edit and user profile edit, but not group edit :/

src/maps/components/AddressPicker.vue
```diff
@@ -1,66 +1,64 @@
...
-            <QItemSection>
-              <QItemLabel>
-                {{ $t('GLOBAL.SEARCH_NOT_FOUND') }}
-              </QItemLabel>
-            </QItemSection>
-          </QItem>
-        </QList>
-      </QMenu>
-    </QInput>
+      <template #selected-item />
```
nicksellen: It's a bit odd that it shows the selected item, as well as the text, maybe there is a better way to fix it up, but did this for now - basically an empty slot template thing.

src/maps/components/AddressPicker.vue
```diff
@@ -1,66 +1,64 @@
...
+          :key="index"
+          v-bind="itemProps"
+          v-on="itemEvents"
+        >
+          <QItemSection>
+            <QItemLabel v-if="useSearchTerm">
+              Set address as
+              <strong>{{ itemLabel }}</strong>
+              <span v-if="hasLocation">(keep existing location on map)</span>
+              <span v-else>(choose location manually on map)</span>
```
nicksellen: These are the strings I didn't add into translations yet, not sure about the wording at this moment.

nicksellen:
Test failures were a bit weird, unrelated to my changes, but fixed now.

tiltec:
I had a quick try, UI seems good; definitely better than before. Will have a detailed look later into code and wording!

src/maps/components/AddressPicker.vue
tiltec: Seems fine to me...

src/maps/components/AddressPicker.vue
```diff
@@ -1,66 +1,64 @@
...
+            </QItemLabel>
+          </QItemSection>
+        </QItem>
+      </template>
+    </QSelect>
     <StandardMap
       class="map"
       :markers="marker ? [marker] : []"
       :prevent-zoom="preventZoom"
+      :default-center="defaultMapCentre"
```
tiltec: `defaultMapCenter` seems more consistent

src/maps/components/AddressPicker.vue
```diff
@@ -1,66 +1,64 @@
...
+              <strong>{{ itemLabel }}</strong>
+              <span v-if="hasLocation">({{ $t('ADDRESS_PICKER.KEEP_EXISTING_LOCATION') }})</span>
+              <span v-else>({{ $t('ADDRESS_PICKER.CHOOSE_LOCATION') }})</span>
+            </QItemLabel>
+            <QItemLabel v-else>
+              {{ itemLabel }}
+            </QItemLabel>
+          </QItemSection>
+        </QItem>
+        <QSeparator v-if="useSearchTerm" />
```
tiltec: Maybe move everything below inside a `<template v-if="useSearchTearm">` to make the meaning more obvious/less repetitive.

nicksellen:
> Aside, I wonder if we should rewrite v-slot: to # to keep it uniform?

Yes, I would like that, seems neater and consistent with generally using short versions where possible.

Issue #1154, opened 2018-12-15
Add stylus linter


tiltec:
It seems rather easy to add a stylus linter for vue files: https://vue-loader.vuejs.org/guide/linting.html#stylelint
Should be helpful to maintain a consistent style and prevent errors like this: https://github.com/yunity/karrot-frontend/pull/1153

EDIT: the linter above is just a CSS linter, need to check if it makes sense to apply it.

tiltec:
stylelint is no good solution for us and stylint (not to be confused!) is not actively developed, so it seems we better continue without.

PR #1888, opened 2019-11-18
Stylus linting

nicksellen:
## What does this PR do?

For some reason I thought to try adding a stylus linter again. I should be doing something else though.

It sort of worked!

I would have liked an eslint-based one, but that doesn't seem to exist. This uses a postcss approach... and this experimental stylus parser [postcss-styl](https://www.npmjs.com/package/postcss-styl).

It has:
- `yarn lint:style`
- `yarn fix:style`
- webpack dev linting with the usual nice overlay thing

## Links to related issues

Also see:
#1154 
#1156 

## Checklist

- [ ] no unrelated changes


src/base/pages/Root.vue
```diff
@@ -34,11 +34,12 @@ export default {
 
 <style lang="stylus" scoped>
 .root-splash-second
+  opacity 1;
```
nicksellen: it adds a `;` when it reorders them :-1:

nicksellen: hmmm... I hoped the "no-trailing-semicolons" rule had fixed it, but that just means no semicolons at the end of a block. The problem is the parser is stylus, but the rules are not designed for stylus....

src/pickups/components/PickupItem.vue
```diff
@@ -116,25 +116,23 @@ export default {
...
+
   &.isEmpty:not(.isDisabled)
-    background repeating-linear-gradient(
-      135deg,
-      white,
-      white 15px,
-      $lightRed 15px,
-      $lightRed 30px
-    )
+    background repeating-linear-gradient(     135deg,     white,     white 15px,     $lightRed 15px,     $lightRed 30px   )
```
nicksellen: Oo-er that's ugly

nicksellen:
Whoops I wrote two custom plugins to get it to look right. Should definitely have not been working on that.

src/pickups/components/PickupItem.vue
tiltec: Why is there so much whitespace?

.circleci/config.yml
```diff
@@ -70,7 +70,7 @@ jobs:
           keys:
           - v1-dependencies-{{ checksum "package.json" }}
           - v1-dependencies-
-      - run: yarn
+      - run: yarn --ignore-engines
```
tiltec: Why would the engine check fail?

nicksellen: The postcss-styl library has an oddly specific engines configuration, probably not intended... but we had something like 10.15...

https://github.com/ota-meshi/postcss-styl/blob/647780fb3d40e497c53d518bf60145c5ed69ad2a/package.json#L37

nicksellen:
Yes, ready to merge! Except for the yarn.lock conflict...

Issue #1459, opened 2019-04-28
Keep message drafts


tiltec:
It's quite easy to lose message drafts on Karrot. Typical causes might be page reloads and components getting temporarily destroyed  (`v-if`). It would be useful to have a "message drafts" concept.

It could be saved in local storage or synchronized to the server. Drafts should be saved for each user and conversation, maybe also when a message gets edited.

Gerald1614:
Hi, i am not sure it correspond 100% to the initial ask, but I could use localStorage to keep the content of the message.  I delete it only if the user press escape, submit the form or leave voluntary the page. but on a page refresh, the content of the draft will be included in the input of the message. I use local Storage because my feeling is that this feature is mainly useful to maintain continuity of user experience when they are inside a specific session. There are not many reasons to maintain the state of a draft across different devices. I guess that at this stage a use would understand he would lose his draft.  it could be done but I personally it would generate backend logic unnecessarily. let me know if you want to exchange on this capability and see the 6 lines of code I had to produce to address that to see if he code does the job or if you would prefer to get different behavior. thanks

tiltec:
Hi, thanks for looking into this! Sure, I'm happy about a localStorage solution too. Can you open the Pull Request?

Gerald1614:
HI, I will. But i have a question. in fact when i go to pickups, we can comment on all pickup dates. when refreshing the web browser the detailsliderbar that contains the conversationCompose disappears. The n the local storage value gets used even if I click on another pickup date. I could instead of saving a single key value pair save an object with the draft and the groupId, placeId and pickup date. in such case I could make it clean but I think it would not be the right thing to do. we should just repvent the detailsliderbar to disappear on page refresh. in such case, the thnigs I build would work as only volunteer exit of this forms would delete the draft. (escape, submit or move to another page). what is exactly the behavior you expect. I could look at why the detailsliderbar disappears on page refresh and try to correct it at the same time. let me know

tiltec:
I think it should keep the draft even when the sidebar is closed or another one gets opened.
Every conversation has a unique id, you could use that to store message drafts (instead of groupId, placeId, etc).

You might need a special handling for editing message that already have been sent, maybe it's ok to not keep drafts of the edited text.

tiltec:
[...]

PR #1891, opened 2019-11-19
Fix Keep message drafts #1459

Gerald1614:
Closes ___ (e.g. #230)

## What does this PR do?
use localStorage to keep draft of messages (wall and pickups messages) even if page is refreshed. The Key refers to the ConversationId and the value to the messsage. The key-value pair is destroyed only when the draft is submited, or the user voluntary quit the page or escape the form.

(briefly describe what this PR is about)

## Links to related issues
[https://github.com/yunity/karrot-frontend/issues/1459]
)## Checklist

- [ ] no unrelated changes

### DIFF
--- a/src/messages/components/ConversationCompose.vue
+++ b/src/messages/components/ConversationCompose.vue
@@ -93,8 +93,8 @@ export default {
       type: String,
       default: null,
     },
-    conversation: {
-      type: Number,
+    conversationId: {
+      type: String,
     },
     slim: {
       type: Boolean,
@@ -107,10 +107,14 @@ export default {
   },
   data () {
     return {
-      message: (this.value) || localStorage[this.conversation] || '',
+      message: localStorage[this.conversationId] || (this.value) || '',
       hasFocus: false,
-      key: this.conversation,
     }
+  },
+  computed: {
+    key () {
+      return this.conversationId
+    },
   },
   watch: {
     value (val) {
@@ -120,7 +124,7 @@ export default {
     },
     message (val) {
       if (val) {
-        localStorage.setItem(this.getKey(), val)
+        localStorage.setItem(this.key, val)
       }
     },
     isPending (val) {
@@ -136,26 +140,20 @@ export default {
     },
   },
   methods: {
-    getKey () {
-      if (this.key) return this.key
-      else return 'wallConv'
-    },
     submit () {
       this.$emit('submit', this.message)
-      localStorage.removeItem(this.getKey())
+      localStorage.removeItem(this.key)
     },
     leaveEdit () {
       this.$emit('leaveEdit')
       this.message = ''
-      localStorage.removeItem(this.getkey())
+      localStorage.removeItem(this.key)
     },
     onFocus (event) {
       this.hasFocus = true
     },
     onBlur () {
       this.hasFocus = false
-      this.message = ''
-      localStorage.removeItem(this.getKey())
     },
   },
 }
--- a/src/messages/components/ConversationCompose.vue
+++ b/src/messages/components/ConversationCompose.vue
@@ -93,6 +93,9 @@ export default {
       type: String,
       default: null,
     },
+    conversation: {
+      type: Number,
+    },
     slim: {
       type: Boolean,
       default: false,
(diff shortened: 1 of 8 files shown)


src/messages/components/ConversationCompose.vue
```diff
@@ -104,13 +107,21 @@ export default {
   },
   data () {
     return {
-      message: (this.value) || '',
+      message: (this.value) || localStorage[this.conversation] || '',
       hasFocus: false,
+      key: this.conversation,
```
tiltec: This is not reactive: if conversation ID gets loaded lazily (i.e. after this component initializes) or changes, `key` will be invalid.

src/messages/components/ConversationCompose.vue
```diff
@@ -125,17 +136,28 @@ export default {
...
+      else return 'wallConv'
+    },
     submit () {
       this.$emit('submit', this.message)
+      localStorage.removeItem(this.getKey())
     },
     leaveEdit () {
       this.$emit('leaveEdit')
+      this.message = ''
+      localStorage.removeItem(this.getkey())
```
tiltec: There's a small bug in this: it would delete a message draft from localStorage when editing a previous message in the same conversation. 

For example, I start writing a new message to you, then edit an old message in our conversation but abort editing and close my browser. I would expect the new message to still exist when I open the browser again.

src/messages/components/ConversationCompose.vue
```diff
@@ -125,17 +136,28 @@ export default {
     },
   },
   methods: {
+    getKey () {
+      if (this.key) {
+        return this.key
+      }
+      else return 'wallConv'
```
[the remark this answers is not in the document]
Gerald1614: ok i will modify that.

src/messages/components/ConversationCompose.vue
```diff
@@ -93,6 +93,9 @@ export default {
       type: String,
       default: null,
     },
+    conversation: {
```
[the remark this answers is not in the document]
Gerald1614: ok

src/messages/components/ConversationCompose.vue
Gerald1614: on the wall there is a cancel button that user can use to cancel the editing and then accept the draft to be deleted but on pickup messages the only way to cancel is to press the escape button. I think it would make sense to add a button here and in such case, I agree this deletion is not necessary. but without this button, i guess most users will just leave the draft thinking that it would cancel the draft. It is more a UX question

src/messages/components/ConversationCompose.vue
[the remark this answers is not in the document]
Gerald1614: in fact, i could get conversationId from editing an existing message but for the one on the top which enables to create a new message, we need an identifier. but what I did was seting the key in the parent component so in my new version it is consistent to just return the key.

src/messages/components/ConversationCompose.vue
Gerald1614: agreed I switched that to a computed prop

src/messages/components/ConversationCompose.vue
tiltec: I'm not sure what you are referring to: as far as I know, the "cancel edit" button shown in both WallConversation and ChatConversation.

Gerald1614:
HI @tiltec  I just pushed a new commit that addresses most of the points mentioned. I unfied the way we handle the conversationId, even for wall conversation. I kept a 'wallConv, key to keep the draft of new wall conversation( no iD as long as message is not submitted). I kept the deletion of the draft on Leave, because it is the only way to delete the draft. As explained earlier it is a UI decision, you can remove if your prefer, but that would involve the localstorage will never get cleaned. I will want on an alterbative leveraging vuex and vuex-persistant as you mentioned, but I not sure it is a good idea as the data is really used only in a single component (ConversationCompose). I wanted to finalise his PR so at leats we have something that works.  as of now people can also edit existing messages. the message showing on the page is the one that is coming from backend, but if they come back to finalise an edit they started earlier, this draft will be used in the edit form. it is only if they submit that the draft will replace previous message. Please test it and let me know if it does the job.

tiltec:
>  I kept a 'wallConv, key to keep the draft of new wall conversation( no iD as long as message is not submitted).

There's more than one wall: each group and each place has one. You implemented it in a way that would remember _one_ message draft for walls, no matter where the user originally typed it in.
I think that's sligthly confusing to users and rather unexpected to both developers and users - what do you think? It should be possible to use conversation ID as well, it exists already for empty walls.

Gerald1614:
Thanks for your feedback, In fact I am using conversation in all the places. as of today, I saw usage of ConversationCompose on the wall and in pickups. for both of them I am using the conversationId so it is universal and it wil store all drafts based on this conversationId. but if you look at the top of the wall page, users can enter directly in Converation compose to edit a new conversation. This conversation can not have an idea because it is not yet part of any conversation. this is only for this one that i created the wallConv. if people edit an existing conversation or add new comment on an existing conversation, the feature will save the draft based on its conversationId. because local storage is used, the draft will be saved only for this specific user (and specific device in fact) so there is no conflict with other groups or places. But if it happens that users are part of different groups or places then I would have to use the groupId and PlaceId in my key used in local storage. that would be easy to implement

PR #1912, opened 2019-11-27
Possible performance improvements

nicksellen:
Two things:
1. throttling the vuex-persisted state saving to localStorage - state changes a lot....
2. minimise calls to reactive methods (Vue.set and watched array methods)

I did look at the Chrome performance tab, although did not make a super rationale approach, but did notice some long calls go away after some of the changes... let's see how the netlify deploy feels.

### DIFF
--- a/src/base/datastore/persistedState.js
+++ b/src/base/datastore/persistedState.js
@@ -1,5 +1,5 @@
 import createPersistedState from 'vuex-persistedstate'
-import { throttle } from 'quasar'
+import { debounceAndFlushBeforeUnload } from '@/utils/utils'
 
 export default createPersistedState({
   paths: [
@@ -7,5 +7,5 @@ export default createPersistedState({
     'auth.push.intention',
     'auth.push.token',
   ],
-  subscriber: store => handler => store.subscribe(throttle(handler, 5000)),
+  subscriber: store => handler => store.subscribe(debounceAndFlushBeforeUnload(handler, 5000)),
 })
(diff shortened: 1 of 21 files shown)


src/base/main.js
```diff
@@ -4,6 +4,7 @@ import 'base/helloDeveloper'
 
 Vue.config.productionTip = false
 Vue.config.devtools = true
+Vue.config.performance = true
```
tiltec: Should we keep this enabled?

src/base/datastore/persistedState.js
```diff
@@ -1,9 +1,11 @@
 import createPersistedState from 'vuex-persistedstate'
+import { throttle } from 'quasar'
 
 export default createPersistedState({
   paths: [
     'i18n.locale',
     'auth.push.intention',
     'auth.push.token',
   ],
+  subscriber: store => handler => store.subscribe(throttle(handler, 5000)),
```
tiltec: I think we might forget some values this way. For example when the locale changes twice within 5 seconds, it would only persist the first value.
`debounce` might be a bit better, as it would always persist the last value. It fails when the user quits the app before the debounce time is over.

nicksellen: Yes, good point. I originally used `debounce`, but switched to `throttle` to catch the case of leaving the page. In freegle I recently added handling to use the `beforeunload` event to persist changes before leaving the page... but would need to do a kind of custom `debounce` for that. Maybe I do that...

src/base/main.js
nicksellen: No, that was an accidental commit.

src/base/datastore/persistedState.js
nicksellen: Did it! `lodash.debounce` has a nice `.flush()` method that I can call in the unload handler to flush any calls that are waiting... fits in nicely.

PR #1915, opened 2019-11-29
Prefetch webfonts

nicksellen:
... and removes the cabin sketch font (used on profile page and error page, but probably not worth downloading it all just for that...)

### DIFF
--- a/src/base/style/app.styl
+++ b/src/base/style/app.styl
@@ -46,9 +46,6 @@ body.mobile .no-mobile-margin
   border 1px solid lightgrey
   border-radius 0
 
-.accent-font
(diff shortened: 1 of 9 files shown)


src/base/style/app.styl
```diff
@@ -46,9 +46,6 @@ body.mobile .no-mobile-margin
   border 1px solid lightgrey
   border-radius 0
 
-.accent-font
-  font-family $accentFont
-
```
nicksellen: Ah yeah, I had somehow made the assumption that quasar had defined an accent font and we were only just overriding it before. Thanks!

Issue #1914, opened 2019-11-29
[Meta] Performance improvements


nicksellen:
A meta issue to collect ideas to work on:
- [x] collect metrics on live sites about loading/rendering/etc times https://github.com/yunity/karrot-frontend/pull/1916
  - chrome (but not firefox) will give interesting things on `performance.getEntriesByType("paint")`
  - ... but we also want to define our own version of "First Meaningful Paint" (e.g. a wall post on the screen), also maybe various intermediate steps (e.g. `hasView` on `Root.vue`)
  - other places to consider:
    - when Root.vue is created/rendered?
    - ..?
- [ ] big one: javascript execution time...
- [x] don't serve fullsize images on group preview page https://github.com/yunity/karrot-frontend/commit/db903ccc5445303239f5dcb62ee842ade3cc249a
- [ ] preload webfonts/css/whatever? (somehow get webpack to know what to do load...)
  - [x] webfonts yunity/karrot-frontend#1915
- [x] `Object.freeze` stuff we don't need reactive (e.g. each individual pickup, in the mutation), to reduce time spent in `_traverse` function #1945
- [ ] use better image formats, google says: _JPEG 2000, JPEG XR, and WebP often provide better compression than PNG or JPEG_
- [ ] investigate loading core js libraries in a main (preloaded?) entry rather than a common chunk (e.g. vuex, quasar...)
- [ ] reduce amount of "all group" data loaded always (group infos, places, ... maybe users?) - this is maybe not so much performance related though
- [ ] optimize loading time of one page, e.g. group wall: reduce number of backend requests, javascript files etc...

tiltec:
Feedback from Bruno, things that feel slow:
- opening notifications takes some time, delayed loading feels "slow"
- in app, "trying to connect to server" can constantly appear and makes it feels slow

tiltec:
I noticed most of the chunks in Karrot change between deploys, even when there are not many changes. It seems due to scoped css id - they get calculated by hashing the whole SFC source file (not just the style section). The parent component apparently needs to know about the scope id. If component x is used 10 times and x changes, this will invalidate about 11 chunks.
Maybe not super important to fix, but still quite unexpected.

tiltec:
[...]

PR #1916, opened 2019-11-30
Record timing stats

nicksellen:
From #1914 

My initial work for recording timing stats.

It works like this:
- we record various points using performance.mark/measure
- when we hit the first component that is mounted with the `v-measure` directive on it, we stop
- at that point we can collect up the points so far and do something with them (print them out in this case)
- we then clear again ready for the next route...

so there are two types of recordings:
1. initial page loading (this is the most interesting one)
2. subsequent page views within the SPA app (less interesting, but still interesting I think)

I called the point where we first render a `v-measure` component "MM" for "Meaningful Mount" (I didn't see a way to detect when it is actually painted, and a brief attempt to only trigger when it was visible showed that it was always visible by the time it was mounted, so I removed that bit). MM triggers earlier than Chromes FMP (First Meaningful Paint), looking at performance tools a bunch of stuff happens after our MM still, looks like re-rendering of stuff because dependent state changed (but not really... maybe just pointless re-renderings...). Still, I think the MM defines an interesting point we can try and reduce.

We could _just_ save MM, and not the intermediate stages, but maybe they are interesting too.

performance

TODO:
- [ ] make it work when `window.performance` is not available
- [ ] include whether it is initial request or not in payload
- [ ] send to backend
- [ ] tidy up a bit
- [ ] define `v-measure` in more places for other routes

### DIFF
--- a/src/utils/performance.js
+++ b/src/utils/performance.js
@@ -101,7 +101,8 @@ function finish () {
     mobile: Boolean(Platform.is.mobile),
     browser: Platform.is.name,
     os: Platform.is.platform,
-    dev: __ENV.DEV,
+    dev: Boolean(__ENV.DEV),
+    app: Boolean(__ENV.CORDOVA),
   })
   debouncedSave()
 }
--- a/src/utils/performance.js
+++ b/src/utils/performance.js
@@ -1,4 +1,3 @@
-import 'whatwg-fetch'
 import Vue from 'vue'
 import { Platform } from 'quasar'
 import router from '@/base/router'
@@ -10,10 +9,13 @@ import { underscorizeKeys } from '@/utils/utils'
 const SAVE_INTERVAL_MS = 5000 // batch saves to the backend
 
 const performance = window.performance
+const fetch = window.fetch
 
 // Make sure we have all the required performance methods
+// and fetch to avoid needing the polyfill (we don't need all browser support)
 // and that vue's performance mode is not enabled (don't want to fight it)
-const ENABLED = performance &&
+const ENABLED = fetch &&
+  performance &&
   performance.clearMeasures &&
   performance.clearMarks &&
   performance.measure &&
--- a/src/utils/performance.js
+++ b/src/utils/performance.js
@@ -1,0 +1,84 @@
+import Vue from 'vue'
+import { Platform } from 'quasar'
+import router from '@/base/router'
+
+// Don't want to fight with vue about clearing marks, etc...
+const ENABLED = !Vue.config.performance
+
+let done = false
+let firstLoad = true
+const startMark = 'karrot-start'
+
+function reset () {
+  done = false
+  clearPerformanceData()
+  performance.mark(startMark)
+}
+
+function clearPerformanceData () {
+  // Don't mess with clearing things if vue performance mode is on!
+  if (!Vue.config.performance) {
+    performance.clearMarks()
+    performance.clearMeasures()
+  }
+}
+
+if (ENABLED) {
+  router.beforeEach((to, from, next) => {
+    if (firstLoad) {
+      firstLoad = false
+    }
+    else {
+      reset()
+    }
+    measure('route', 'before')
+    next()
+  })
+
+  router.afterEach(() => {
+    measure('route', 'after')
+  })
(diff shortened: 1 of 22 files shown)


nicksellen:
I decided to simplify it, so _only_ record the "meaningful mount" (the bit where the first `v-measure` is triggered). Will make it much simpler to graph and look at, and less data to store...

It's almost ready now... just need a backend, then to wire them up.

nicksellen:
So it records stuff like:

simple

tiltec:
Nice! Maybe add the route params or the full path, that could make it easier to find out if some pages are slow because they have many variable elements on them.

nicksellen:
Yup, added that now too :)
So the stats backend will convert it into this format:

```js
[
  {
    "measurement": "karrot.stats.frontend",
    "fields": {
      "ms": 14661
    },
    "tags": {
      "first_load": true,
      "logged_in": true,
      "mobile": false,
      "group": 3,
      "route_name": "placePickups",
      "route_path": "/group/3/place/12/pickups",
      "route_params__group_id": "3",
      "route_params__place_id": "12"
    }
  }
]
```

nicksellen:
One additional thing I'm wondering whether to add in is the duration of various HTTP requests - as the current timing would not make it clear whether it's just a slow connection, or slow rendering.

... the performance API can tell me the duration of each of the resource requests (img, script, xhr, etc....), and maybe we could sum them per type. Many of the requests would happen in parallel but it would give an indication.

Or, we leave it till next time...

nicksellen:
Additionally, I think the beforeunload will not work in all browsers, but this sendBeacon thing is kind of designed for it and has more support: https://developer.mozilla.org/en-US/docs/Web/API/Navigator/sendBeacon

nicksellen:
Hmm, send beacon doesn't allow to set custom headers (so server fails it with XSRF error), it could be disabled, but it looks like there is another approach using `keepalive` flag (https://fetch.spec.whatwg.org/#request-keepalive-flag), but can't see how to set that in axios. It has a `keepAlive` thing when using an agent, but that's a totally different thing iiuc.

Maybe this is enough as it is for now, and can do those bits later...

nicksellen:
I will make it work! https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequest/Synchronous_and_Asynchronous_Requests says:

> There are some cases in which the synchronous usage of XMLHttpRequest is not replaceable, like during the unload, beforeunload, and pagehide events. You should consider using the fetch() API with the keepalive flag. When fetch with keepalive isn't available, you can consider using the navigator.sendBeacon() API, which can support these use cases while typically delivering a good UX.

I've remade it with `fetch()` .. but will finish that up later...

nicksellen:
Ok, I added the resource timing too... just a total of everything. Should hopefully give some indication of connection speed.

tiltec:
I wonder if all browsers that support  the `performance` API also have `fetch` built in, then we wouldn't need the extra dependency!

nicksellen:
Could be! Also maybe I should load it conditionally depending on if fetch is available.

nicksellen:
Ok, I removed whatwg-fetch again, and we just only do the stats if we have a native fetch available. Should be enough browsers for that. fetch and performance browser support is not the same, some that support one but not the other.

I also added a few more stat fields whilst I was there, namely: `browser`, `os`, `dev`, `app`.

I think I'll stop now.

src/utils/performance.js
```diff
@@ -0,0 +1,145 @@
...
+    keepalive: true,
+    headers: {
+      'Content-Type': 'application/json',
+      [axios.defaults.xsrfHeaderName]: readCookie(axios.defaults.xsrfCookieName),
+    },
+    body: JSON.stringify(underscorizeKeys(data)),
+  })
+}
+
+const debouncedSave = debounce(save, SAVE_INTERVAL_MS, { maxWait: SAVE_INTERVAL_MS * 2 })
```
tiltec: Or rather use `debounceAndFlushBeforeUnload` and remove the `unload` listener at the end of the file?

nicksellen: Ah yeah, I went in a few cycles with it, and back to the starting point again :)

Issue #1940, opened 2019-12-09
Add create group button


nicksellen:
I think the time is right for making the group create button :)

I think it could go on the groupPreview page (https://karrot.world/#/groupPreview) on the same line as the "Which group do you want to join?". Something like this:

![createanewgroup](https://user-images.githubusercontent.com/31616/70437123-2f073800-1a8b-11ea-9e5e-5058c71b521e.png)

I think ideally the create a group form would have more context, maybe a multi-page stepper with some explanatory text to start with, and suggestion to join the community forum, etc... but also OK to leave that off for now.

tiltec:
Done in #1942

PR #1942, opened 2019-12-10
Add create new group link

nicksellen:
#1940 

## What does this PR do?

This adds a "create new group" link on the group gallery page.

I did not handle the case that there are no groups the user can join. In that case the link won't show. But I don't _think_ that should be too common.... I just wanted to make the simplest possible thing. I mean, it might actually be nicer to make another group gallery card with a "create your group" kind of thing (like the new offer thing)... but maybe that is just an optimization.

createnewgroup

## Checklist

- [ ] no unrelated changes

### DIFF
--- a/src/groupInfo/components/GroupGalleryUI.vue
+++ b/src/groupInfo/components/GroupGalleryUI.vue
@@ -40,6 +40,13 @@
         class="text-primary header"
       >
         {{ $t('JOINGROUP.WHICHGROUP') }}
+        or
+        <router-link
+          :to="{ name: 'groupCreate' }"
+          class="underline text-lowercase"
+        >
+          {{ $t('GROUP.CREATE_TITLE') }}
+        </router-link>
       </p>
       <div class="row items-start no-wrap q-mt-md">
         <div class="col">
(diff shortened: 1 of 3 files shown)


src/groupInfo/components/GroupGalleryUI.vue
```diff
@@ -40,6 +40,13 @@
         class="text-primary header"
       >
         {{ $t('JOINGROUP.WHICHGROUP') }}
+        or
```
tiltec: I guess that should be translatable. Maybe use the i18n functional component to provide more context/give more freedom to translators? e.g. `{joinGroup} or {createGroup}`

nicksellen: Oh yeah, you mean not everyone says _or_? So complicated!

nicksellen:
Done!

> I think you forgot to include locale-en.json in this PR.

Originally, I hadn't actually added anything. I secretly reused the create group title.... so it would already be translated...

tiltec:
Clever with the `text-lowercase`! I didn't notice that yesterday, but seems good.

PR #1945, opened 2019-12-17
Freeze some objects to improve performance

tiltec:
Related to https://github.com/yunity/karrot-frontend/issues/1914

- freeze most objects and arrays in Vuex
- replace objects instead of modifying them (no Vue.set/delete on frozen objects)

I also got rid of group and pickup join/leave mutations because they are also done vie websockets. I know that @nicksellen prefers to have basic functionality working without websockets, so I would consider adding it back later after deploy when there's a performance improvement.

### DIFF
--- a/src/authuser/datastore/auth.js
+++ b/src/authuser/datastore/auth.js
@@ -202,7 +202,7 @@ export default {
   },
   mutations: {
     setUser (state, user) {
-      state.user = user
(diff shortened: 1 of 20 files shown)


tiltec:
I did some small tests in Chrome, maybe 4% performance improvement on my computer.

PR #1982, opened 2020-02-01
Add pagination for pickups

### LINKED ISSUE #952: Add proper pickups pagination
Update: pagination might be the solution, see https://github.com/yunity/karrot-frontend/issues/952#issuecomment-434743691

---

Sometimes rendering the pickups page seems a bit slow. I used the new [vue performance tool](https://github.com/vue-perf-devtool/vue-perf-devtool) to record when loading that page:

pickups-performance

It shows it taking 745ms to render the `<Pickups>` component with the data I have in my localhost database, and the breakdown of which subcomponents make up that time.

I don't think it's a high priority issue, as maybe groups don't have that many pickups on the page, but something to be aware of, or to investigate if you feel particularly motivated :)

ang-zeyu:
Closes #952 

## What does this PR do?

Adds pagination for `GroupPickups.vue`

pagination

## Links to related issues
https://github.com/yunity/karrot-frontend/issues/952

## Checklist

I can't find an existing test for the `GroupPickups.vue` page, so I'm not sure if it was intentionally omitted. ( I'll be glad to add one if not though 😄 )

- [x] no unrelated changes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,8 @@ Please document your changes in this format:
 ```
 
 ## [Unreleased]
+### Added
+- Pagination for pickups [#1982] @ang-zeyu
 
 ## [8.3.0] - 2019-12-02
 ### Added
--- a/src/__snapshots__/storyshots.spec.js.snap
+++ b/src/__snapshots__/storyshots.spec.js.snap
@@ -8789,167 +8789,169 @@ exports[`Storyshots PickupList Default 1`] = `
         <animate attributeName="fill-opacity" from="1" to="1" begin="0s" dur="0.8s" values="1;.5;1" calcMode="linear" repeatCount="indefinite"></animate>
       </circle>
     </svg></div>
-  <div class="q-card">
-    <div class="no-padding content q-card__section q-card__section--vert">
-      <div class="q-pa-sm full-width">
-        <div><span class="featured-text">
+  <div class="q-infinite-scroll">
+    <div class="q-card">
+      <div class="no-padding content q-card__section q-card__section--vert">
+        <div class="q-pa-sm full-width">
+          <div><span class="featured-text">
           8:00 AM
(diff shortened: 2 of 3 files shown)


tiltec:
Hi @ang-zeyu, welcome! Nice work :)

One thing I noticed is that the filter count becomes misleading: it shows "10/10", whereas it's rather 30 entries on 3 pages.

image

I wonder if we should move pagination into `PickupList`?

ang-zeyu:
Hi @tiltec , thanks for looking through this!

> I wonder if we should move pagination into `PickupList`?

I realised less than 10 entries may show too actually, in addition to it being misleading, which isn't ideal.

Should we add an additional label of some sort beside the filter count to clarify what it shows, in addition to moving pagination into `PickupList`?

E.g. 
Perhaps something like `1-10 out of 23 results for <filter name>` ( from [amazon](https://www.amazon.com/b?node=18332383011&pf_rd_p=fed09b50-0584-490e-a698-ff8a751dc738&pf_rd_r=APSESKRX1ZSJ0AFAZWMT) )

Issue being internationalisation is needed, which's workflow I'm not too familiar with. ( is full internationalization usually pushed together with a feature here? )

nicksellen:
> Issue being internationalisation is needed, which's workflow I'm not too familiar with. ( is full internationalization usually pushed together with a feature here? )

No, we use https://www.transifex.com/ for that so you don't need to worry about that - just make sure text is in English.

I actually wondered if this would be better and more consistent as an infinite scroll (as we use that pattern elsewhere).

I would also be happy with it as seperate pages though as it is here, just with the "1-10 out of 23 results" bit implemented.

ang-zeyu:
Sorry for the late update!

> I actually wondered if this would be better and more consistent as an infinite scroll (as we use that pattern elsewhere).

Went with this, think its more consistent as well :+1:

Issue #1535, opened 2019-05-23
Rework group sidenav


tiltec:
From Bruno.

To remove/move elsewhere:
- issues
- description

To add:
- [x] applications (then the banner on the group wall could go away)

Although people seem to like the group description: https://community.foodsaving.world/t/remove-or-move-internal-group-description-aka-group-description-visible-for-members/259/3
We could create a "group info" page, that also contains the "leave group" button as well as the public group description.

tiltec:
To keep in mind: potential confusion between group wall and place wall, as both are just called "Wall" in Karrot. Maybe rename group wall to something like "everyone" or the group name? 

(From https://community.foodsaving.world/t/karrot-ux-design-suggestion-a-little-mockup/273/3)

PR #1987, opened 2020-02-29
changed sidenav, notification icons and added settings to topbar notifications

brnsolikyl:
Closes ___ (e.g. #230)

## What does this PR do?
A work in progress... Probably a lot of details on the same request that should be tackled separately. Here are the things that have been changed:
- Some options on sidenav were hidden or changed place
- Icons on notification toggle were changed (trying to bring more coherence)
- Added a shortcut to notification settings when clicking at message notifications on the topbar (still not very pretty and need to make it not selected by default after you click on it)

## Links to related issues
Rework Group Sidenav - https://github.com/yunity/karrot-frontend/issues/1535

## Checklist

- [ ] no unrelated changes


src/sidenav/components/SidenavMenu.vue
```diff
@@ -32,6 +32,47 @@
         </QBadge>
       </QItemSection>
     </QItem>
+    <QExpansionItem
+      switch-toggle-side
+      :label="$t('BUTTON.SHOW_MORE')"
+      dense
+      dense-toggle
+      expand-separator
+    >
```
nicksellen: The "Show more" is not aligned with the other icons/labels which looks odd to me:

misaligned

src/messages/components/LatestMessages.vue
```diff
@@ -19,6 +19,11 @@
         :label="$t('CONVERSATION.REPLIES')"
         :count="unseenThreadsCount > 9 ? '9+' : unseenThreadsCount"
       />
+      <QRouteTab
+        :title="$t('GROUP.SETTINGS')"
+        icon="fas fa-cog"
+        :to="{ name: 'settings', hash: '#notifications' }"
+      />
```
nicksellen: I got a bit confused here, I think because I read "GROUP.SETTINGS" but this is not group settings page it goes to, it's user settings, but also the specific part is notifications.

Hmmm, I have a stronger feeling to actually seperate out this user page into a range of subpages (profile, account, notifications).

But maybe this is a reasonable intermediate solution.

nicksellen: ... but perhaps a different icon/label to make it clearer that it's the notification settings for the messages?

src/sidenav/components/SidenavGroupUI.vue
```diff
@@ -35,7 +35,10 @@
         </QBtn>
       </div>
     </template>
-    <SidenavMenu :entries="entries" />
+    <SidenavMenu
+      :entries="entries"
+      :entriesmore="entriesmore"
```
nicksellen: Could this be `entriesMore` to keep to lowerCamelCase convention?

src/sidenav/components/SidenavMenu.vue
nicksellen: I think we also need to include a badge on the Show more, if it's closed and one of the items inside has a badge.

src/sidenav/components/SidenavGroupUI.vue
brnsolikyl: Changing everything to `entriesMore` gives me a warning about the following rule: https://eslint.vuejs.org/rules/attribute-hyphenation.html

So I changed it to `:entries-more=entriesMore"`
Or should I just ignore the warning and have all lowerCamelCase without hyphenation?

src/messages/components/LatestMessages.vue
brnsolikyl: Yes, it can be confusing looking at the code. It's the string I found that would display "Settings" for the user. I found another string named SETTINGS one level up but could not make it work.

As for the icon solution, there is `fas fa-user-cog`, but from the context I prefer having the cog icon only. Another solution would be to stack icons (like envelope and cog). I'll see how it looks

src/sidenav/components/SidenavGroupUI.vue
nicksellen: Yes, you got it! lowerCamelCase when in javascript, kebab-case when in a template :+1:

brnsolikyl: Camels and kebabs... these things have funny names

src/sidenav/components/SidenavMenu.vue
brnsolikyl: I found a half solution for the "Show More". I could align the toggle but not really the text. But I'm too lazy to be bothered by it now. hehe

Badge included! Not the best solution if we add other items with badges in the future, but works for now

src/messages/components/LatestMessages.vue
```diff
@@ -19,6 +19,11 @@
         :label="$t('CONVERSATION.REPLIES')"
         :count="unseenThreadsCount > 9 ? '9+' : unseenThreadsCount"
       />
+      <QRouteTab
+        :title="$t('SETTINGS.TITLE')"
+        icon="fas fa-cog"
+        :to="{ name: 'settings', hash: '#notifications' }"
+      />
     </QTabs>
```
nicksellen, line 27: The point came up (on our last coding session call) that it would be better to have the cog as a small button in the corner, at the moment it's a implemented as a third tab but with no content, so after first clicking it you get sent to the right place, but then when you re-open the menu you can see the "cog" tab selected, with no content:

emptytab

I had a play around and found a solution I think is nice:

flex

I'll talk through the changes as a learning process, so maybe you can implement them.

Currently the component (when displayed as a menu) is a vertical stack of two components:
- `<QTabs>`
- `<LatestConversations>` or `<LatestThreads>` depending on which tab is selected

(the `<template>` component does not result in anything in the document)

But to implement it not as a tab, we need a more involved layout like:

```
<QTabs> | <the cog>
<LatestConversations or LatestThreads>
```
To get the `<QTabs>` and the cog to appear horizontally side-by-side, we can wrap them in a div and use _flex_ layout (see [quasar docs](https://quasar.dev/layout/grid/introduction-to-flexbox) on flexbox).

Quasar lets you use flex layout using the class `flex`, so we get...

```html
<div class="flex">
  <QTabs>... stuff ...</QTabs>
  <QBtn
    flat
    round
    icon="fas fa-fw fa-cog"
    :title="$t('SETTINGS.TITLE')"
    :to="{ name: 'settings', hash: '#notifications' }"
  />
</div>
<... the rest>
```

Initially this results in the tabs and the cog pushed all the way to the left, but we can use more flex stuff to say that the tabs should fill all the space (this is called "grow" in flexbox terminology) using the quasar class `col-grow` on the thing that should grow, so the `<QTabs>` in our case, i.e.:

```html
<QTabs
  ... all the existing stuff
  class="k-message-tabs col-grow"
>
```

That's it!

src/sidenav/components/SidenavMenu.vue
nicksellen: Ok! I had a dig into the alignment issue of the "Show more", it's because in the rest of the list we are not using the "standard" layout for "list with icons" but our own combination of quasar components.

We use `side` mode for the `<QItemSection>` that contains the menu icon, but the standard approach uses the `avatar` mode. If we used `avatar` mode then all the other menu entry texts would move further to the right and appear inline:

showmore-aligned

However, at some point we/somebody decided that doesn't look great with our more condensed style. So, I wouldn't suggest we switch to `avatar` spacing... but rather find a way to switch the "Read more" to use _our_ way.

... however, `QExpansionItem` doesn't give an option to control the display of the expansion arrow when using `switch-toggle-side` - it always uses `avatar` spacing.

What we _can_ do though, is:
- not use `switch-toggle-side` mode
- specify a custom #header slot (we do this already...)
- define the header bit the same way as we do for the other menu items
- include an icon

Using a horizontal ellipsis icon it would look like this:

showmore-closed

... and when open:

showmore-open

I quite like this approach as I can't see a way to align the arrow when using `switch-toggle-side` without using some custom override CSS, which we try and avoid.

Whaddya think?

src/sidenav/components/SidenavMenu.vue
```diff
@@ -57,6 +90,10 @@ export default {
       type: Array,
       default: () => [],
     },
+    entriesMore: {
```
nicksellen, line 93: You don't actually need to implement `entries-more` in this component, it's only the `SidenavGroupUI` that needs to know about the "more" bit.

Where it renders the the "more" menu it can simply use: `<SidenavMenu :entries="entriesMore" />` instead of `<SidenavMenu :entries-more="entriesMore" />`.

src/sidenav/components/SidenavMenu.vue
nicksellen: (this comment isn't really on the correct file, it should be on `SidenavGroupUI`)

brnsolikyl: Ah that was the case! I think the solution is good. I was even wondering if instead of using the horizontal ellipsis we could use a plus icon (that turns into a minus when open, if it's worth the trouble).

src/messages/components/LatestMessages.vue
brnsolikyl: Nice! Done now, thanks for fixing this!

tiltec:
I finally had a look today, sorry for being late! Great work you did there! Some comments:

Latest messages menu: the grey bell icon now looks like it indicates a special kind of notification, whereas it actually says that you _don't_ get emails for it. I think it's a bit out of context here, maybe we just get rid of it altogether?
Screenshot_2020-05-06_16-22-19

Sidenav: I think it would be nice if the "Show more" button would hide when clicking it, then it doesn't look like it's yet another sidenav button. It wouldn't be possible to hide the buttons without reloading the website, but I think that might be ok.
image

It would also avoid this badge duplication (by the way, great that you added this):
image

In case you like these changes but feel it would be too much hassle for you, I could do them.

brnsolikyl:
@tiltec I think the suggestions make sense!
1 - Grey bell notifications on latest messages: yes we could get rid of it altogether. We could also consider changing the icon to a slashed envelope, but I could not find something like this at FA's icons. Maybe if we stack it. Another detail is that hovering on it says "Notifications are disabled", but the right thing would be "E-mail notifications are disabled". I wonder if this string is used elsewhere in other contexts...

2 - Sounds good to me. I'm not sure the Expansion Item component has this, I could not find it. How would you do it?

3 - Do you mean that the badge on Show More should disappear when clicked?

I will probably not work on this anytime soon (except for #1 which looks the easiest). If you want to, feel free to do them! :)

tiltec:
> I wonder if this string is used elsewhere in other contexts...

I checked, it's only used there. So we are free to change or remove it.

> I'm not sure the Expansion Item component has this, I could not find it. How would you do it?

That requires a bit of trickery, since it's technically not just expansion anymore. Maybe to not use the ExpansionItem... needs to experimentation.
@nicksellen as you already tried out some variations of this, did you also consider the option to hide the "show more" button after expansion?

> Do you mean that the badge on Show More should disappear when clicked?

Exactly, or rather if the whole "Show more" entry disappears it doesn't matter anymore ;)

nicksellen:
>  @nicksellen as you already tried out some variations of this, did you also consider the option to hide the "show more" button after expansion?

nope, I can't quite imagine how that would work, as there would be no way to close it back up again after...?

relatedly, I recently was enjoying the protonmail settings menu with expanding sections:

protonmenu

so that's another possible direction, somehow categorizing the menu items rather than hiding the less used ones behind the "show more", which, with a quick think could be like:

- wall
- pickups
  - pickups
  - feedback
- offers
- community
  - members
  - applications
  - issues

(where clicking on the category expands the menu _and_ takes you to the first sub-item)

then remove "Show public group information" (who wants to see that?) and "Map" (is also available from the expand icon on the map itself).

also, I guess we shouldn't be discussing this on a merged PR ;)

brnsolikyl:
I like the idea above. Kind of solves the problem of excess of buttons in the future when more features are developed.

Btw the "Map" button was not supposed to show on desktop, only on mobile. There's a conditional there on the code that doesn't work. Tilmann said one if I remembe correctly that it probably had to do with the Quasar update.

nicksellen:
Another implementation of show more/less, from slack:

slackshowmore

slackshowless

tiltec:
That seems a good option! Also, it looks a bit different from the other options, so it's less confusing (e.g. no hover background)

PR #1997, opened 2020-03-02
Add features section to docs

nicksellen:
Based on the features doc I wrote before https://gist.github.com/nicksellen/f399835de7c53767fbf502c1f77f11ea

It's far from complete/perfect, but might be useful in its current state already?


nicksellen:
(ah, I built it on top of https://github.com/yunity/karrot-frontend/pull/1996 so lets handle that first)

tiltec:
I just found this again, do you think it would be useful for you or future contributors in its current state? If yes, I would just merge it. 

We could also schedule a "docs day" where we update it and fill in some blanks. I guess I could add a lot there, but I'm lacking the motivation right now :(

nicksellen:
I think we should just merge it as it is and go from there....

PR #1971, opened 2020-01-30
Added the first draft to the About page

brnsolikyl:
Closes ___ (e.g. #230)

## What does this PR do?

Adds the first draft of the About-page

## Checklist

- [ ] added a test, or explain why one is not needed/possible...
- [ ] no unrelated changes
- [ ] asked someone for a code review
- [ ] joined #karrot-dev channel at https://slackin.yunity.org
- [ ] added an entry to CHANGELOG.md (description, pull request link, username(s))

PR #1998, opened 2020-03-04
Add shiny landing page

nicksellen:
## What does this PR do?

- adds a landing page based on the content in #1971 
- at `/` url for logged out users
- not viewable by logged in users

I just focused mostly on design/layout/content presentation, not really the mechanism about where/how this should be stored/translated, and how it should be configurable for other instances, but I do have some ideas on that.

This default content can be added to karrot-frontend translations, by splitting up the titles/paragraphs into seperate translation blocks.

For making it customizable per instance, we could:
- include the default content (i.e. this stuff) in the frontend repo, kind of like I have it here
- instances that want custom content can have a mechanism to override it on the frontend
- they will need to build a custom frontend (webpack) to get their content in
- they will need to manage translations of that content themselves
- customization could just be via plain html files, _or_ support custom vue components (not quite sure of the implications/viability of this right now)

I also did consider a mechanism of storing the content in the backend, and have it so it's customizable by putting some files on the server in the right place. This is simpler in some ways as no custom frontend build is needed. But downsides are that it makes it harder to manage our default content... there _could_ actually be a mechanism to allow custom content on the backend, but default content still in the frontend...

Not yet resolved...

## Links to related issues

#1971 

## Checklist

- [ ] no unrelated changes


nicksellen:
I updated this now:
- all text in i18n
- added some pictures of some groups
- many some others minor bits and bobs

brnsolikyl:
The layout looks good, but I'd rather not see everything together on the landing page, since it's way too much information at first sight, in my opinion. I think we could have a few things there and the rest on another ABOUT or FEATURES page. The things I'd keep on the landing page are the first section with the two first paragraphs (until "... consumerist society") and then the section showing with a few existing groups (nice touch there btw), keeping the buttons "Sign Up" and "Browse Existing Groups" somewhere in between sections or at the end.

All of the rest I'd put at a separate page, for the person interested in reading more. This other page could be reached either through a "READ MORE" button at the end of the landing page and/or through the topbar (ABOUT or FEATURES).

src/base/pages/Landing.vue
```diff
@@ -0,0 +1,352 @@
...
+      <QImg
+        :src="screenshots.pickups"
+        alt="karrot"
+        class="screenshot screenshot-fullwidth q-my-lg"
+      />
+    </p>
+
+    <p v-t="'LANDING.VISION1'" />
+
+    <div class="row inline-images">
```
brnsolikyl: Maybe here one of the images could be of the Bike Kitchen. I have a collection of them here, free to use: https://cloud.disroot.org/s/DmKLL32s9ZxBCCo
I think `1 - bikeworkshop.jpg` could work fine

nicksellen: Yes! I thought having some non-food saving group photos would be nice! I've added the bikeworkshop one now.

nicksellen:
I was aiming for the long product page style which lots of projects use, e.g.:
- https://funkwhale.audio/
- https://rocket.chat/
- https://www.jetbrains.com/idea/

(I didn't especially base it off those, just some examples)

So, it's maybe more of a product page than a singular focused landing page.

To me the intention is not that they read it all, but they read as much as they are interested in. Personally, when I'm viewing a long product page I just flick around a bit and see what catches my eye.

The main sections are:
1. opener
2. big screenshot
3.  vision with user photos
4. feature blocks
5. existing groups
6. participative development
7. and more...

1-5 are ones which have been "presented" in some way, whereas 6-7 are more-or-less just the text from the VISION document, so I could imagine 6-7 going away, as they are more wall-of-text-y, but also would rather merge this when it's acceptable, and not create additional pages until after that.

Another way would be to spread out the content a bit more, or some bigger horiztonal blocks/headers to divide it up more, which could reduce the overwhelmingness?

Maybe we can test it out on a few people too... see what their reactions are.

brnsolikyl:
I see! If we keep this format I think it would definitely be better if the horizontal blocks are wider. It gives another feel to it, less overwhelming as you wrote.

The ideal would be to test on different people, for sure. We need to keep in mind (or figure out) who the target group is and I believe is for the people who would like to create a group, generally community organizers who want to see if Karrot's values and features are a good fit for what they want to do. Of course a landing page can be interesting for people invited to join a group as well, but if they find it through an invitation, joining the group is the main thing keeping them on Karrot.

I also like the idea of moving 6-7 to another page later but leaving them there for now.

nicksellen:
I'll merge this later tonight, or tomorrow if there are no further suggestions. We can always iterate on it from there.

src/base/pages/Landing.vue
```diff
@@ -0,0 +1,352 @@
...
+import { QImg } from 'quasar'
+import { mapGetters } from 'vuex'
+
+import GroupGalleryCards from '@/groupInfo/components/GroupGalleryCards'
+import logo from '@/logo/assets/carrot-logo.svg'
+import screenshotPickups from './images/karrot-screenshot.png'
+
+import screenshotGallery from './images/karrot-gallery.236x236.jpg'
+import screenshotManagePickups from './images/karrot-manage-pickups.236x236.jpg'
+import screenshotOffers from './images/karrot-offers.236x236.jpg'
```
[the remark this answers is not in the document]
nicksellen: Thanks! I've changed it for that one now.

PR #2072, opened 2020-05-26
Change groups at topbar home icon

### LINKED ISSUE #1746: Make it easier to switch between groups
From Bruno:

> For example, a drop-down menu on the group logo at the header to the left, showing other groups that a person belongs to, would probably make a difference.

brnsolikyl:
Closes #1746 

## What does this PR do?

Adds a group switcher to the topbar home icon

## Checklist

- [x] no unrelated changes


brnsolikyl:
Oh, two details that need fixing:
1 - Groups that do not have an image are not aligned with those that have. Still not sure how to fix it.
2 - How to do about inactive groups? They're probably still showing and we do not want them to show there, right?

tiltec:
To align groups without image, you could add a placeholder div with fixed size.

I'm not concerned about listing inactive groups - show them if the users is still member (they get removed anyways after 7 months of inactivity)

What about users that have only one group? I think the Karrot logo was a nice way to get back to the group home page. Of course, they could also use the sidenav menu to achieve the same.

I just thought of having a dropdown-type of thing next to the Karrot/group logo - it would always show your current group (replacing that part of the breadcrumb). When you click on it, jump to group home page. When you click the dropdown-arrow on the right, show the group menu. What do you think about this?

src/sidenav/components/SidenavMenu.vue
```diff
@@ -32,6 +32,39 @@
         </QBadge>
       </QItemSection>
     </QItem>
+    <QItem
```
tiltec, line 35: This looks like leftovers from a previous PR.
You can restore it to the previous version: `git checkout master -- src/sidenav/components/SidenavMenu.vue`

src/locales/locale-en.json
```diff
@@ -277,7 +277,7 @@
   "TOPBAR": {
     "LOGOUT": "Logout",
     "USERPROFILE": "Visit your profile",
-    "CHANGE_GROUP": "Change group"
+    "SHOW_ALL_GROUPS": "Show all groups"
```
tiltec, line 280: TOPBAR.CHANGE_GROUP is used in other places, don't remove it yet!

src/topbar/components/KTopbarUI.vue
```diff
@@ -19,8 +19,49 @@
...
+            <QItemSection
+              avatar
+              class="groups"
+            >
+              <QIcon
+                v-if="!group.hasPhoto"
+                name="fas fa-users"
+                class="groups"
+              />
+              <img
```
tiltec, line 41: To avoid errors like `TypeError: Cannot read property 'thumbnail' of undefined` because of `group.photoUrls.thumbnail`, don't try to show the photo if there is none.

src/topbar/components/KTopbar.vue
```diff
@@ -13,6 +14,7 @@ export default connect({
   },
   actionsToEvents: {
     logout: 'auth/logout',
+    // selectGroup: 'currentGroup/select',
```
tiltec, line 17: seems not needed?

tiltec:
The snapshots seem to diverge a lot from `master` - your packages might be out of date. To refresh them:

```
git checkout master -- src/__snapshots__/storyshots.spec.js.snap
git merge master
yarn
yarn test -u storyshots
```

One "story" also needs updating; you might have noticed the warning `[vuex] unknown getter: groups/mine` when updating the snapshots. To fix it, open `Layout.story.js` and add another mock getter in `createDatastore`:

From
```js
groups: { getters: { all: () => groupsMock } },
```
to
```js
groups: { getters: {
    all: () => groupsMock,
    mine: () => groupsMock,
  } },
```

brnsolikyl:
Thank you for the review, Tilmann! I believe the things you pointed out are fixed. However, I just noticed a little unexpected behavior that needs fixing also before merging. When going to the group gallery, all of groups in which the user's application is pending are appended to the list of groups on the topbar.

tiltec:
> However, I just noticed a little unexpected behavior that needs fixing also before merging. When going to the group gallery, all of groups in which the user's application is pending are appended to the list of groups on the topbar.

Seems they should either be always there or never, regardless if the user visited the group gallery. It just affects the people that have an application pending, so it may not be many on production sites.

Which solution would you prefer? Hide groups for which applications are pending?

tiltec:
By the way, the linter wants some more newlines in the code I gave you, have a look at this:

```
/root/repo/src/topbar/components/Layout.story.js
   9:11  error  Expected a line break after this opening brace   object-curly-newline
  12:5   error  Expected a line break before this closing brace  object-curly-newline
 ```

**Comment by brnsolikyl** [comment]
Probably not the most elegant solution, but everything seems to work now! You're welcome to change nonetheless. I'm just glad I made it work, but I can learn from other approaches.

File: src/groupInfo/datastore/groups.js [hunk 5]
```diff
@@ -41,6 +41,7 @@ export default {
       return Object.values(state.entries).map(getters.enrich)
     },
     mine: (state, getters) => getters.all.filter(e => isMyGroup(e)).sort(applicationsFirstThenSortByName),
+    isMemberGroups: (state, getters) => getters.all.filter(e => isMemberGroup(e)).sort(applicationsFirstThenSortByName),
```
nicksellen, line 44: Minor: you can simplify this:

```js
.filter(e => isMemberGroup(e))
```

to this:

```js
.filter(isMemberGroup)
```

Because `e => isMemberGroup(e)` is just a function that takes one argument (`e`) and calls one function (`isMemberGroup`), that's also the same as just the reference to the function `isMemberGroup` itself...

nicksellen, line 44: Also minor.

I find the name `isMemberGroups` confusing, as normally something starting with `is` will return a boolean value (e.g. the `isMemberGroup` function).

I would probably call it `mineWithoutApplications` ... or maybe call it `mine` and rename the existing `mine` to be `mineWithApplications`.

src/topbar/components/KTopbarUI.vue
```diff
@@ -19,8 +18,50 @@
...
+            v-close-popup
+            :to="{ name: 'group', params: { groupId: group.id } }"
+            clickable
+          >
+            <QItemSection
+              avatar
+              class="groups"
+            >
+              <QIcon
+                v-if="!group.hasPhoto"
```
nicksellen, line 36: Minor: I would swap around the if condition, so have the image first with a `v-if="group.hasPhoto"` and then the users icon as a `v-else` fallback.

Reason: slightly simplifies the if, don't have to think twice, if _not_ ...? and also presents it in the ideal order, the image is the main case, the icon just otherwise...

src/groupInfo/datastore/groups.js
brnsolikyl: Got it! I copied the line above, `(e => isMyGroup(e))`, so I changed both getters now to this simplified reference to the function

brnsolikyl: I went for `mine` and `mineWithApplications`. Hopefully I managed to change it wherever it is referenced. Nothing seems to be broken

nicksellen:
Great work :+1: ... just the snapshot tests failing still... `yarn test -u` will fix it, then commit the `src/__snapshots__` directory :) Then go for a merge!

nicksellen:
I merged master into it (rebase had conflicts) and updated the snapshots :)

PR #2097, opened 2020-06-15
Rename pickups to activities at code/file level

nicksellen:
## What does this PR do?

This is part of https://github.com/yunity/karrot-frontend/issues/1127.

It only implements the code/file level changes such that the whole site should look/work as before.

It also is made to work with the existing backend names (`/api/pickup...`), with the idea we can merge this as soon as possible and then incrementally work on the rest later.

PR #1236, opened 2019-01-24
WIP: A script to rename locale keys and keep translated messages

tiltec:

PR #2107, opened 2020-06-25
More renaming pickups to activities

nicksellen:
This is an extension of https://github.com/yunity/karrot-frontend/pull/2097 to rename everything needed to work with the renamed backend https://github.com/yunity/karrot-backend/pull/1011.


tiltec:
Ah, you decided to rename the translations! This has a lot of pitfalls, I'll just comment on this for now.

I revisited my script from https://github.com/yunity/karrot-frontend/pull/1236/files and this seems relevant:
> Pushing _translated_  messages to transifex has PATCH semantics: every key that's specified will get updated, others will remain
> Therefore, we need to push the whole locale-en.json file, but only the rewritten keys in other locale files.

It's maybe not very clear from the text, but it we actually _shouldn't_ push anything that didn't change. For example, if you push `locale-ar.json` as you have it in the pull request, Transifex would mark every message as translated to Arabic (accepting the original English text as valid translation to Arabic) and show 'Nick Sellen' as translator. It's less bad for well-translated languages such as `de`, but it would still show you as translator for all strings.
In order to minimize the fallout, we should craft those special `locale-xx.json` files that only contain key-value pairs with modified key names and push those. Maybe it's a good time to merge https://github.com/yunity/karrot-frontend/pull/1236 ...

nicksellen:
Ok! I modified the script to do the key renaming by function (reads existing keys, renames by regex match/replace). It seems to behave as you describe, I ran it without the final push bit at the end, and it seems to modify the files as you describe correctly.


For non en-locales:
- only keys which are changing are present in the output files
- if there is no translation available, it puts an empty string there
- if there was a translation at the old key, that is copied over

PR #2128, opened 2020-07-20
Add images to conversations messages

### LINKED ISSUE #1793: Add picture upload to conversations
From Disa in https://community.foodsaving.world/t/weekly-call-about-karrot-development/289/27

> Several people in Bikekitchen Gothenburg have told me they would want to be able to upload pictures to the chats in karrot, and perhaps would use it more if it was possible.

This is related to #963 - my comment from https://github.com/yunity/karrot-frontend/issues/963#issuecomment-458984048 is relevant here:

> it's already possible to reference a picture in [...] Markdown, but users would need to figure out the hosting solution by themselves (needs to be on https to fulfill our Content Security Policy).
This is closely related to having pictures in any message on Karrot, so if we decide to implement that, we might add an option to upload pictures and add them into a Markdown field with a single click. Might look up how Discourse does it.

nicksellen:
Fixes #1793 (frontend) requires backend https://github.com/yunity/karrot-backend/pull/1028

This implements image upload for all conversations and threads (walls, private chats, activity chats, etc...).

It does it in a similar way to offers, 0-n images can be added to the message, and does the multipart upload along with the data.

I made the thumbnail display smaller than the offer images, and you cannot re-order them as you can with the offers images.

## TODO

- [ ] ~~maybe add the images into the email notifications for messages (backend)~~ later.. also backend issue only
- [x] maybe have an **X** button for the mobile image gallery dialog
- [x] maybe make the mobile image gallery dialog full width/screen
- [x] investigate why websocket updated messages don't show photos :/ (backend issue... transaction issue I think, where it sends the websocket before the images are saved... either way they are not in the websocket payload)
- [ ] ~~possibly add error message when image is too large (I think it just silently fails at the moment)~~ can't recreate on desktop right now, maybe investigate more in the future...
- [x] make it work in threads!
- [x] "replies" box has no spacing after images
- [x] move MultiCroppa into a general/util place (as it's not just for offers now)
- [ ] ~~upload progress?~~ (could be useful, _but_ in my initial tests the slowest bit is the server doing it's thing, maybe the thumbnailing?)
- [x] move the action button a bit higher so it doesn't cover the message buttons


nicksellen:
The branch deployment is set up for this now, so you can use try it out at https://add-message-images.dev.karrot.world :)

nicksellen:
> Would be great to have a story to see how images are being rendered in WallConversation and ChatConversation.

I added some stories :)

PR #2149, opened 2020-08-13
Add total member count to MemberList

djahnie:
## What does this PR do?

Fixes the confusion I had several times when I wanted to quickly look up how many members were in my current group.

## Checklist
- [x] no unrelated changes


tiltec:
Looks good!

Maybe the font size could be a bit smaller, so that it looks more like a subtitle and grabs less attention?

Might also be good to hide it when the page is loading, otherwise there's a scary moment ;)
Peek 2020-08-13 11-48

src/users/components/UserList.vue
```diff
@@ -117,6 +124,9 @@ export default {
     inactiveSublabel () {
       return this.inactiveUsers.length + ' ' + this.$tc('JOINGROUP.NUM_MEMBERS', this.inactiveUsers.length)
     },
+    activeUsersCount () {
```
nicksellen, line 127: Here it's called "activeUsersCount", but in the [video from @tiltec](https://github.com/yunity/karrot-frontend/pull/2149#issuecomment-673380955) it looks like it's counting _all_ the members (5 active + 1 inactive).

I'm not actually sure which is a more useful number? Although given we already have the inactive user count in the little Inactive section, maybe active members is most useful? and given you named it "activeUsersCount" maybe that is what was intended too.

In which case you can use `this.activeUsers.length` (instead of `this.users.length`).

A nitpick too, I would expect something named "activeUsersCount" to return me a _number_ rather than the text, so maybe `activeUsersLabel` is more descriptive?

Alternatively there is the template syntax, which could look something like this:

```html
<QItem v-t="{ path: 'GROUP.MEMBERLIST', args: { total: activeUsers.length } }" />
```

djahnie:
Thanks for looking at it, guys! All very valid points! I think I fixed it all now.?

src/users/components/UserList.vue
```diff
@@ -1,10 +1,21 @@
 <template>
   <div class="list-wrapper">
     <QList>
+      <QItem
+        v-if="users.length > 1"
+        v-t="{ path: 'GROUP.MEMBERLIST', args: { total: users.length } }"
```
nicksellen, line 6: Did you intend to use `users` instead of `activeUsers`?

src/locales/locale-en.json
```diff
@@ -307,6 +307,7 @@
     "SERVER_ERROR": "Server error"
   },
   "GROUP": {
+    "ACTIVE": "Active",
```
nicksellen, line 310: This translation has been added, but seems unused?

src/users/components/UserList.vue
```diff
@@ -1,10 +1,21 @@
...
+        class="text-caption"
+      >
+        {{ activeUsersCount }}
```
nicksellen, line 9: `activeUsersCount` no longer refers to anything, so can be removed! The `QItem` can become self-closing tag now, e.g. `<QItem ...props and stuff... />`

djahnie:
Oops. Seems that I was hasty last night... :sweat_smile: 
I'll take another look now!

src/users/components/UserList.vue
djahnie: Yep, that was intentional. Before I had both mentioned but that created a phrase that was too long. And sure, I could just use active ones, but I thought it could make sense to show the total amount, since inactive users get kicked out at some point anyways.

src/locales/locale-en.json
djahnie: True, it's a leftover. Thanks for pointing it out!

src/users/components/UserList.vue
djahnie: Indeed. That was the most obvious of my hasty mistakes yesterday... :sweat_smile:

Issue #2157, opened 2020-08-18
Order of feedback


dwaxweiler:
In my group, some of us are confused by the current default order of the feedback entries, which are ordered by the date they have been written. Example:

![Screenshot_2020-08-18 Karrot](https://user-images.githubusercontent.com/12501775/90478463-77541b00-e12d-11ea-93cf-a57abd759501.png)

We would find it more logical to order them by default by their pickup's date as one person of us needs to go through them in chronologic order of the pickups. The current order could be offered as a second sorting option.

I also think that the user experience will improve by changing the default order as the pickup date is shown way bigger as the time the feedback entry has been written.

nicksellen:
Seems reasonable to me, I created a PR for the backend https://github.com/yunity/karrot-backend/pull/1044 :)

I think we can just change it for everyone, and if somebody wants the old behaviour back, _then_ add an option. This way we don't need any frontend changes, and no more buttons.

tiltec:
I wonder if the frontend has its own sorting logic implemented? It did make sense, because our state is a mixed bucket from different sources/purposes.

nicksellen:
> I wonder if the frontend has its own sorting logic implemented? It did make sense, because our state is a mixed bucket from different sources/purposes.

It does! Just throwing together a fix...

nicksellen:
It's fixed in master now (can see it in action at https://dev.karrot.world/#/group/1/feedback). Will get deployed to production later this week probably :)

PR #2161, opened 2020-08-19
Sort feedback by activity date

nicksellen:
Closes https://github.com/yunity/karrot-frontend/issues/2157

## What does this PR do?

The backend was already switched (in https://github.com/yunity/karrot-backend/pull/1044), but the frontend _also_ does sorting, so this PR fixes that... it would fallback to feedback created at date if there is no activity (which would be odd...).


nicksellen:
I'll just merge, as it's quite small :)

Issue #2080, opened 2020-05-31
Streamline application flow


tiltec:
The usual process to join a group right now is to sign up, find the group and submit the application. The alternative is to send out invitations. Our statistics show that invitations are not used a lot: last 6 months on karrot.world: 96 accepted invitations, 395 accepted applications. Invitations were predominantly used in three groups: two DLC groups (64, 87) and EfA (29).

Our idea in #1984 was to remove the invitation flow in favor of a more streamlined application workflow, starting from the group preview page (e.g. https://karrot.world/#/groupPreview/37).

Reasons for removing the invitation flow were mainly to just have one way to become a group member, which should reduce conceptional complexity for Karrot users. We also want to focus our time on building refined features for Karrot, instead of having many half-baked ones.

It doesn't seem strictly necessary to remove invitations very soon, but we could hide them a bit more first while we streamline the application flow.

brnsolikyl:
A summary of the flow as it is now (for users invited to an existing group):
User gets link to the groupPreview page -> clicks on sign-up button -> sign-up page -> (back to groupGallery page) confirms e-mail -> lands at e-mail verification page with "take me home" button (to groupGallery page) -> finds group of interest again -> application

One idea to skip a few steps:
... -> confirms e-mail -> groupPreview page that the user wanted to apply in the first place, perhaps with a "E-mail verified successfully" notice -> application

brnsolikyl:
Just got complaints from unrelated people in Solikyl, two days in a row, about joining a group on Karrot for the first time.

One of them said that she did held an introduction for a couple of people to join the group, but that they "got stuck in Germany". Asking more I found out that she meant the Playground group.

A similar complaint from another one was that it was not easy to find the right group to do the application after signing up, and that people did not understand why they ended up at the Playground.

Related to https://community.foodsaving.world/t/how-to-handle-the-spam-in-the-playground-group/486/10

tiltec:
That sounds annoying! Yeah, playground-auto-join might be too confusing given that people who visit the site first might not understand the concept of a "group" at all...

Let's make the playground group closed and don't mention it during signup, then see how it turns out!

tiltec:
Do we still need the "open groups" feature at all? As far as I know, all active groups have been migrated to using applications. This would simplify some logic and enable us to refine the application flow without thinking about implications for open groups.

brnsolikyl:
Open groups don't seem very relevant now... :)

[...]

PR #2175, opened 2020-08-28
Remove playground auto-join during signup

tiltec:
Related to #2080 

## What does this PR do?

Remove option to auto-join playground group on signup. This seemed to confuse people.

## Checklist

- [x] no unrelated changes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,6 +25,7 @@ Please document your changes in this format:
 ## [8.7.2] - 2020-08-28
 ### Changed
 - Enable the Luxembourgish locale of Quasar and date-fns when this language is selected @dwaxweiler
+- Do not offer to join the playground group during signup, to avoid confusiong @tiltec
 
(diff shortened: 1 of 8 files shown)


nicksellen (approved):
Great! I wondered if we can actually remove _all_ the `joinGroupAfterLogin` stuff? When I looked at https://github.com/yunity/karrot-frontend/blob/5d36f48e2c16914514ea4a7be27ae4738868e317/src/authuser/datastore/auth.js#L139-L144 it's about group joining, but you can't join groups like that any more, you have to apply... maybe it still "works" because there is an error joining, and it gets redirected to the group preview page from where they can apply?

I'll approve this as we can always think about that another time (and as part of https://github.com/yunity/karrot-frontend/issues/2080#issuecomment-681894374)

tiltec:
Yeah, that's some almost-unused logic there! I'll comment on the issue instead ;)

PR #2179, opened 2020-08-30
Switch to Quasar App

tiltec:
Quasar App is the proposed way of using Quasar framework. We didn't do it so far, because Karrot is older than it and we were dealing just fine with plain webpack. The promise of Quasar App is that it'll manage the development environment for us (in the spirit of vue-cli and create-react-app), so that we can have all fancy features without much extra work (e.g. Capacitor, PWA)

This PR switches Karrot over to using Quasar App, instead of plain webpack. It is half-finished: dev-server and production build work, as well as tests and storybook. Still, a lot might be broken.

~~I didn't migrate the service workers yet, because the integration of our existing code for firebase push seemed fiddly.~~ Service workers are operational! I briefly tested web push on firefox and chrome, should test some more on mobile. Also, does "Add to Homescreen" work?

Other open issues:
- ~~there's a big vendor chunk by default, maybe we want to split it again?~~ vendor chunk is split!
- VSCode complains about "Property ... does not exist on 'CombinedVueInstance" -> might be some configuration error
- ~~addressbar color might be missing on mobile~~ should work now, not tested yet

I changed the ansible playbook to reflect different output from Quasar app, we should run it when we deploy! https://github.com/yunity/yuca/commit/2660249bb92c213984e9d152580a9fee617ee0ea


tiltec:
@nicksellen maybe you want to run this locally on your machine and start the dev server? Hope it works! Just `git checkout quasar-app && yarn && yarn dev`

quasar.conf.js
```diff
@@ -0,0 +1,276 @@
...
+  try {
+    return {
+      key: fs.readFileSync(process.env.KEY),
+      cert: fs.readFileSync(process.env.CERT),
+      ca: fs.readFileSync(process.env.CA),
+    }
+  }
+  catch (e) {
+    console.log('Could not find key/cert/ca files, falling back to our own...', e)
+    return true
```
nicksellen, line 38: This stuff causes this to printed out on `yarn build`:

```
Could not find key/cert/ca files, falling back to our own... [  Error: TypeError [ERR_INVALID_ARG_TYPE] [ERR_INVALID_ARG_TYPE]: The "path" argument must be of type   string or an instance of Buffer or URL. Received undefined
```

Maybe check if env vars are set...

quasar.conf.js
```diff
@@ -0,0 +1,276 @@
...
+
+  return {
+    // https://quasar.dev/quasar-cli/supporting-ts
+    supportTS: false,
+
+    // https://quasar.dev/quasar-cli/prefetch-feature
+    // preFetch: true,
+
+    // app boot file (/src/boot)
+    // --> boot files are part of "main.js"
```
nicksellen, line 68: `main.js` is gone now...

src/subscriptions/firebase.js
```diff
@@ -1,66 +1,26 @@
...
 
-async function removeServiceWorkers () {
-  for (const worker of await getServiceWorkers()) {
-    worker.unregister()
-  }
-}
-
-window.onbeforeunload = () => {
-  // Do not return anything from this function or it triggers a browser dialog
-  if (shouldRemoveServiceWorkersOnUnload) removeServiceWorkers()
```
nicksellen, line 65: My goal with all this stuff was to not leave service workers hanging around if we don't actually need one. They seem like pollution to me. If they don't have push enabled, or any offline processing why have one registered?

That said, I'm fine with the new approach to always have one.

nicksellen (approved):
Great work! I'm quite happy to see this.

I approved, but a few minor things to tweak first perhaps. I can also do them if you're not computer inclined...

build/script.serve.js
```diff
@@ -4,7 +4,8 @@
 
 const express = require('express')
 const compression = require('compression')
-const proxyMiddleware = require('http-proxy-middleware')
+const proxyMiddleware = require('http-proxy-middleware').createProxyMiddleware
+console.log(proxyMiddleware)
```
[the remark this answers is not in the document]
nicksellen: ... ah also elsewhere in this file needs to be `./dist/pwa` now...

src/subscriptions/firebase.js
tiltec: It's not that the user enables offline processing, it'll be a default now - in a way, having service workers seems the fancy version of browser caching. We set our assets to "max expiry" and don't care how long they will be on users machines, but the browser takes care of that. I would expect browsers to occasionally clear old service workers if they are not needed anymore - not sure if they do that already...

tiltec:
I noticed I can't login after "add to homescreen", should fix that...

src/subscriptions/firebase.js
nicksellen: I mean, they are _never_ needed at the moment on the desktop site if you don't enable push notifications, but will be there anyway. As I said though, I'm fine to go with having them always enabled, I don't have any compelling evidence that it's beneficial to minimise their use.

PR #2205, opened 2020-10-11
[CI] update dependencies

tiltec:
Good morning, I found some fresh dependencies for you!

You can check the site at https://circleci-update-dependencies.dev.karrot.world to make sure everything looks ok.


package.json
```diff
@@ -70,7 +70,7 @@
     "vuelidate": "^0.7.4",
     "vuex": "^3.5.1",
     "vuex-connect": "^2.0.1",
-    "vuex-persistedstate": "^3.1.0",
+    "vuex-persistedstate": "^4.0.0-beta.1",
```
tiltec, line 73: we need to pin this to v3 until we migrated to vue 3/vuex 4.

package.json
```diff
@@ -106,14 +106,14 @@
...
-    "jest": "^26.5.2",
+    "http-proxy-middleware": "^1.0.6",
+    "jest": "^26.5.3",
     "jest-canvas-mock": "^2.3.0",
     "jest-environment-jsdom": "^26.5.2",
     "jest-junit": "^12.0.0",
     "jest-serializer-vue": "^2.0.2",
-    "jstoxml": "^1.6.11",
-    "keep-a-changelog": "^0.10.1",
+    "jstoxml": "^2.0.1",
```
tiltec, line 115: @nicksellen maybe  this fixes the escaping?

package.json
```diff
@@ -50,31 +50,31 @@
...
-    "firebase": "^7.23.0",
+    "firebase": "^7.24.0",
     "intl": "^1.2.5",
     "jstimezonedetect": "^1.0.7",
     "leaflet": "^1.7.1",
     "lodash.debounce": "^4.0.8",
     "loglevel": "^1.7.0",
-    "markdown-it": "^11.0.1",
-    "markdown-it-emoji": "^1.4.0",
+    "markdown-it": "^12.0.0",
```
tiltec, line 59: need to check breaking changes

package.json
nicksellen: Yes, should do! Although I was hoping my test would fail when the upgrade happens, for trying to do double escaping... will investigate...

package.json
nicksellen: Looks ok... https://github.com/markdown-it/markdown-it/blob/master/CHANGELOG.md#1200---2020-10-14

package.json
nicksellen: It's not clear to me quite how to pin it, I tried:
- `"^3.1.0"`
- `"3.1.0"`
- putting "vuex-persistedstate": "3.0.0" into `"resolutions"` (https://classic.yarnpkg.com/en/docs/selective-version-resolutions/#toc-how-to-use-it)

... but in all cases doing `yarn upgrade --latest` gives us v4. Hmm. Am I missing something?

package.json
```diff
@@ -39,92 +39,92 @@
...
+    "axios": "^0.21.0",
     "clone-deep": "^4.0.1",
     "core-js": "^3.6.5",
     "date-fns": "^2.16.1",
     "deep-equal": "^2.0.4",
     "deepmerge": "^4.2.2",
     "element-closest": "^3.0.0",
     "escape-string-regexp": "^4.0.0",
-    "firebase": "^7.23.0",
+    "firebase": "^8.0.0",
```
tiltec, line 53: whoah, another major version…

Issue #1704, opened 2019-08-22
Add location-based overview page


djahnie:
Gothenburg is the perfect example of a city that has multiple sharing groups that can benefit from Karrot. To enhance synergies it would be beneficial if there was a place to see all the groups that belong into that area. (We first discussed this in the [2019-08-07 Karrot call](https://community.foodsaving.world/t/weekly-call-about-karrot-development/289/12).)

Requirements/ideas
- visible for everybody, not only logged in users
- autogenerated based on group location (e.g. 50 km radius..?)
- accessible via search function on landing page and/or direct link
- no extra marker on the world map
- no tile like all groups have
- really only shows when looked up (or should it..?) and then shows a visibly different tile for the overview page (different layout? different section? unclear still...)
- when clicked shows all group tiles of that region on the right, where the world map was before (needs different solution for mobile)
- should work with different spellings (e.g. gothenburg, göteborg, gbg - not sure how well supported this is)

brnsolikyl:
From your description it seems like a quite hidden feature. I cannot really understand how the user would see it except when looking for it. And still a bit unclear how it would look like. But the specs are good and I'm all for it, as you know this already. hehe
 
Here's another way I imagine this feature -. Karrot's first page when logged out would no longer be the world map by default but a zoomed in map (the 50 km radius you mentioned, or maybe just the default scale in OSM for the city in question), based on the user's geolocation. The group tiles shown first are "Groups near you" and scrolling down or zooming out you'd have the rest on "Discover the world".

There should also be an easy way to access this page when logged in, although that is part of the larger issue of facilitating the change between groups.

Thinking some steps ahead, this feature would work awesome together with public places #1449 being marked on the map as well. It could easily create a more dynamic and updated version of these websites mapping the sharing economy like smartakartan.se here in Gothenburg or https://pamapam.org/ in Barcelona

djahnie:
Ah, now that I started making a sketch of what I imagined this feature to be first, I came to realize that your idea is much better, @brnsolikyl... :sweat_smile: 
So I'll just close this issue and make a new one maybe..? And add a wireframe of how yours could look like...^^

brnsolikyl:
Don't throw your idea in the trash bin prematurely! Can't it be saved? Maybe there's parts of it that still can be consumed? :smile: 

I think we could still work on the same issue. Anyhow I'm looking forward to see the wireframe!

brnsolikyl:
[...]

PR #2212, opened 2020-10-19
Emphasize nearby groups

nicksellen:
Backend PR: ~~https://github.com/yunity/karrot-backend/pull/1065~~ merged
Next backend PR: https://github.com/yunity/karrot-backend/pull/1072

## What does this PR do?

Emphasizes nearby groups in the group gallery map. The list will order by distance (or member count if distance is not available). It will emphasis groups that are within 3x the distance of the nearest group to you, e.g.: if the nearest group is 50km -> will emphasis groups within 150km. It has a minimum "nearest group" value of 30km (to avoid scenarios like: nearest group is 1km -> will emphasis all groups within 3km = not enough groups!). Arguably this "30km" value could be less?

Might need some refinement when trying it out with the real group locations, or maybe some better sample data...

It also removes special handling for the playground group now we are de-emphasizing that feature.

Branch deployment --> https://change-emphasise-nearby-groups.dev.karrot.world

## Checklist

- [ ] no unrelated changes


src/boot/loadInitialData.js
```diff
@@ -1,8 +1,23 @@
 import { isNetworkError } from '@/utils/datastore/helpers'
+import bootstrap from '@/base/api/bootstrap'
 
 export default async function ({ store: datastore }) {
-  await datastore.dispatch('auth/refresh')
-  datastore.dispatch('groups/fetch')
+  // await datastore.dispatch('auth/refresh')
+  // datastore.dispatch('groups/fetch')
```
tiltec, line 6: can be removed

src/boot/loadInitialData.js
```diff
@@ -1,8 +1,23 @@
...
+
+  const bootstrapData = await bootstrap.fetch()
+  const { user, groups, geoip } = bootstrapData
+  console.log('bootstrapData is', bootstrapData)
```
tiltec, line 10: comment can be removed

tiltec (commented):
Thanks for working on this! Seems a nice approach!

I wonder how it works when no group is close to the users location? Probably it zooms out quite a lot, possibly showing half of the world?

Also, maybe it would be good to add a small comment onto the group list that it's sorted by distance?

nicksellen:
> I wonder how it works when no group is close to the users location? Probably it zooms out quite a lot, possibly showing half of the world?

Yes, it _should_ revert to existing behaviour. Not because it's great, but to just try and constrain this PR to thinking about the new behaviour and not do a wider review of other scenarios.

I updated the top comment with an updated description of the current behaviour, and wondered if the 30km should be reduced, to 10/15/20km...

_or_ we just get this merged, and deployed, and assess it with more real scenarios and iterate on that... I think it can't be much _worse_ than current behaviour!

> Also, maybe it would be good to add a small comment onto the group list that it's sorted by distance?

I _think_ what might be better is to actually seperate it into two sections, "nearby groups", and other groups. Actually a bit like [this idea](https://github.com/yunity/karrot-frontend/issues/1704#issuecomment-526868848) from @brnsolikyl :+1: 

> The group tiles shown first are "Groups near you" and scrolling down or zooming out you'd have the rest on "Discover the world".

But I think I'd prefer to do that in future work, or support @brnsolikyl to do it maybe :)

brnsolikyl:
Yeah, I think that to fine-tune it, we would have to play around with real-world scenarios, so I'm positive of merging it the way it is now.

I'd be happy to give a shot at the frontend later, @nicksellen !

nicksellen:
Great! I'll wait for an approval to be checked then go!

tiltec (approved):

PR #2252, opened 2020-11-29
Composable data concept

nicksellen:
This is the early stages of thinking about reworking our data layer away from vuex towards a "composable data" concept, borrowing from and using the new compositional APIs.

So far, it's more concept, experiment, discussion, and feedback. I included 2 markdown docs that should give some flavour of what I'm thinking about.
- [README.md](https://github.com/yunity/karrot-frontend/blob/change/composable-data/src/lib/composableData/README.md)
- [DESIGN.md](https://github.com/yunity/karrot-frontend/blob/change/composable-data/src/lib/composableData/DESIGN.md)

I also pushed my messy [experimental branch](https://github.com/yunity/karrot-frontend/tree/experimental/composable-data), but if we go forward, I would do a clean re-implementation, use bits of that as needed.

### DIFF
--- a/src/lib/composableData/DESIGN.md
+++ b/src/lib/composableData/DESIGN.md
@@ -1,0 +1,78 @@
+# compostable data design document
+
+This is a more abstract document to reflect on the design goals of the library. The [README](README.md) has the more walkthrough kind of approach for an actual user.
+
+## Aspects to consider
+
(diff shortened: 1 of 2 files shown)


nicksellen:
It'll be worth waiting until Quasar is upgraded to Vue3 and we are upgraded to that Quasar --> https://github.com/quasarframework/quasar/issues/7836

PR #2258, opened 2020-12-01
Don't use activity join hover when not hover available

### LINKED ISSUE #2257: Joining activity on Iphone requires two tapings before getting confirmation dialog
I've heard that someone had this problem in our group and now I tested on an Iphone 7 on Safari iOS 14.1

Normal behavior to join an activity is to tap once and then get the dialog to confirm. In this case you need to tap twice: on the first you'll see your avatar, and only on the second you get the dialog to confirm. 

karrot-joining-pickup-iphone

nicksellen:
Closes #2257

Branch deployment URL: https://fix-ios-no-hover.dev.karrot.world

## What does this PR do?

For joining an activity, we show the profile photo of the person joining (or an "x" for leaving) using hover states. In ios safari it sees that we use hover states and to allow the user to still access the "hover" state it makes the first click perform the "hover" and the second click the actual click.

This makes two changes:
1. use a `@media (hover)` query to not define anything for the `:hover` psuedoclass state (which means devices that don't really have a meaningful hover will not have it defined)
2. put the click handler on a higher up element (as the hover element is no longer shown)

It should be tested on a few devices (https://fix-ios-no-hover.dev.karrot.world) as I was bit unsure how the current site actually works on non-ios safari mobile browsers, e.g. chrome/firefox on android, as those don't have hover states either? or maybe they went a different route and it JustWorks :confused: 

## Checklist

- [ ] no unrelated changes

### DIFF
--- a/src/activities/components/UserSlot.vue
+++ b/src/activities/components/UserSlot.vue
@@ -3,6 +3,7 @@
     class="user-slot-wrapper"
     :class="{ greyedOut: !showJoin, active: showJoin }"
     :style="{ width: size + 'px', height: size + 'px' }"
+    @click.stop="$emit('join')"
   >
     <div :class="{ hoverHide: showJoin }" />
     <div
@@ -14,7 +15,6 @@
         :user="hoverUser"
         :size="size"
         :is-link="false"
-        @click.native.stop="$emit('join')"
       />
     </div>
   </div>
@@ -67,14 +67,18 @@ export default {
   .hoverShow
     display none
 
-.user-slot-wrapper.active:hover
-  border 0
+// ios safari will require two clicks if we define a hover state
+// so only use the hover thing when the browser can support hover properly
+// see https://css-tricks.com/annoying-mobile-double-tap-link-issue/
+@media (hover)
+  .user-slot-wrapper.active:hover
+    border 0
 
-  .hoverHide
-    display none
+    .hoverHide
+      display none
 
-  .hoverShow
-    display inline-block
+    .hoverShow
+      display inline-block
 
 .greyedOut
   cursor ini


brnsolikyl:
It solved the hover problem but now the avatar is not visible after confirmation. It only shows after reloading the page.

nicksellen:
Progress! I'll have a look. Fortunately the Firefox mobile emulator seems to at least recreate some of the iOS quirks.

nicksellen:
Hmmm, cannot recreate it like that. It's odd, as the joining logic is totally separate. It shouldn't need a page reload. Maybe I need to make friends with an iOS person. Maybe the local neighbourhood help WhatsApp group can help me :)

larzon83:
Your changes are working on localhost but not on your deployed url you mentioned.

Video of localhost:

https://user-images.githubusercontent.com/23213583/105422018-55fadc00-5c43-11eb-9de7-f66f374eea8d.mp4

Also strange, I have to tap two times on "Ja" when I want to remove.

nicksellen:
I just rebased on master and pushed, which will redeploy it again.

For me, the behaviour of local and https://fix-ios-no-hover.dev.karrot.world are the same. I'm only using the firefox mobile view with "iPhone 6/7/8 iOS 11" selected. It looks like you're using browserstack or something similar? Or a local emulator?

The two taps for "unsigning up" is odd too :/

Would welcome any fixes in this area! It's hard to test when I can't recreate it, I did think to see if we still have a browserstack account still, I think we did once... or one of the alternative services, but if you have a good setup already, would be great to resolve it.

larzon83:
@nicksellen
I created an new PR: https://github.com/yunity/karrot-frontend/pull/2297

I pulled the latest changes in my fork and did a `npm i` and copied the changes you made here. It works on localhost with the Apple iOS Simulator on my Mac. But [https://fix-ios-no-hover.dev.karrot.world/](https://fix-ios-no-hover.dev.karrot.world/) doesn't work in the Simulator. Could you do another deploy of my PR?
I also fixed the double-tap on leave (also using `@media (hover)`) and renamed a few things for better readability. Hope it's ok :)

nicksellen:
I replied to the deployment point in [my review comment](https://github.com/yunity/karrot-frontend/pull/2297#pullrequestreview-579119948), but wanted to point out that we use `yarn` rather than `npm` for deps, so if you use `npm i` you might get a slightly different set of dependency resolutions (and I think it'll create a `package-lock.json` locally for you), so better to use the same tool.

larzon83:
> I replied to the deployment point in [my review comment](https://github.com/yunity/karrot-frontend/pull/2297#pullrequestreview-579119948), but wanted to point out that we use `yarn` rather than `npm` for deps, so if you use `npm i` you might get a slightly different set of dependency resolutions (and I think it'll create a `package-lock.json` locally for you), so better to use the same tool.

Ok, I install yarn and try with that.

nicksellen:
Closed in favour of https://github.com/yunity/karrot-frontend/pull/2297 :)

PR #2293, opened 2021-01-21
changed: use default group location for defaultMapCenter when creating a new place

### LINKED ISSUE #2288: Improve initial map location when adding a new place
When I add new place to a Karrot group, the initial location of the map seems to be always the same (somewhere north of Darmstadt). I need to zoom out, find the region where the group operates and zoom into it to find the exact location of the place I want to add.

It would be nice if the map view could be initialized to a location that makes sense for the group. For instance:
- the coordinates I picked up when creating the group itself, with some default zoom level. Probably quite easy to implement, but might not be optimal since it might not match the range of current places of the group so well in some cases;
- the coordinates I put down in my user profile - but that's quite hidden, I assume many people haven't set any location there;
- fit the map view to the bounding box of all existing places in that group? Probably a bit harder to implement, but could give some nice results. It looks like you are already doing it for the map of the group in the top left corner so hopefully that logic can be reused…
- let the group participants define the initial map view for the new places? Probably overkill and annoying to implement.

larzon83:
Closes #2288

## What does this PR do?

When creating a new place, use the location of the group instead of somewhere Darmstadt.
No tests atm, because we struggle with the test setup/usage. We try to add them in this PR.

## Links to related issues

https://github.com/yunity/karrot-frontend/issues/2288

## Checklist

- [x] no unrelated changes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,8 @@ Please document your changes in this format:
 ```
 
 ## [Unreleased]
+### Changed
+- use default group location for defaultMapCenter when creating a new place @larzon83 [#2293]
 
 ## [9.1.0] - 2020-01-06
 ### Fixed
--- a/src/places/components/PlaceEdit.vue
+++ b/src/places/components/PlaceEdit.vue
@@ -224,6 +224,11 @@ export default {
         status: 'created',
       }),
     },
+    currentGroup: {
+      required: false,
+      type: Object,
+      default: () => ({}),
+    },
     allPlaces: { required: true, type: Array },
   },
   computed: {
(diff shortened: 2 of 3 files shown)


nicksellen:
It looks good to me!

If you think you can add a useful test, then that would be great, but I would still be happy to approve without a test because:

- the change is quite small
- I manually verified it
- I'm not so enthusiastic about our test setup, we kind of neglected the frontend tests for a while as they seemed more work than would be nice

The changelog would be great though.

When you're ready ping me again for a "review" (which I will pass if it has the changelog) :)

larzon83:
@nicksellen changelog updated. Can be merged. Maybe we create a separate PR for tests :)

nicksellen (approved):

nicksellen:
Thanks! All merged.

It'll be auto deployed to https://dev.karrot.world in the next few minutes.

Deployment to production is when me or @tiltec press some buttons manually, usually when there are a few changes ready to go live (it builds new mobile apps, and deploys them to play store etc., so best not to trigger new updates on every change).

PR #2297, opened 2021-01-28
fix(activities): only hover pictures if device supports it

### LINKED ISSUE #2257: Joining activity on Iphone requires two tapings before getting confirmation dialog

larzon83:
Closes #2257

## What does this PR do?

Optimization of this PR: https://github.com/yunity/karrot-frontend/pull/2258

`UserSlot.vue`:
- add `@media (hover)` to enable hover if device supports it
- remove `hoverHide` div and its style as it's not needed
- rename `hoverUser` prop to "user" for better readability
- rename `hoverShow` class to "show-picture-on-hover"
- fix wrong style declaration `cursor ini` in class `greyedOut` -> now it's `cursor default` as these boxes should not be clickable

`EmptySlot.vue`:
- fix wrong style declaration `cursor ini` in class `greyedOut` -> now it's `cursor default` as these boxes should not be clickable
- add `@click.stop` to stop going to activity detail when clicking on an empty slot

`CurrentUser.vue`:
- add `@media (hover)` to enable hover if device supports it (the leave icon in this case)

## Checklist

- [ ] no unrelated changes

### DIFF
--- a/src/activities/components/ActivityItem.vue
+++ b/src/activities/components/ActivityItem.vue
@@ -170,7 +170,6 @@ export default {
     },
     detail (event) {
       if (event.target.closest('a')) return // ignore actual links
-      if (event.target.classList.contains('user-slot-wrapper')) return // ignore actual links
       this.$emit('detail', this.activity)
     },
   },
--- a/src/activities/components/ActivityItem.vue
+++ b/src/activities/components/ActivityItem.vue
@@ -170,6 +170,7 @@ export default {
     },
     detail (event) {
       if (event.target.closest('a')) return // ignore actual links
+      if (event.target.classList.contains('user-slot-wrapper')) return // ignore actual links
       this.$emit('detail', this.activity)
     },
   },
--- a/src/activities/components/UserSlot.vue
+++ b/src/activities/components/UserSlot.vue
@@ -3,18 +3,17 @@
     class="user-slot-wrapper"
     :class="{ greyedOut: !showJoin, active: showJoin }"
     :style="{ width: size + 'px', height: size + 'px' }"
+    @click.stop="showJoin ? $emit('join') : null"
   >
-    <div :class="{ hoverHide: showJoin }" />
     <div
-      v-if="hoverUser && showJoin"
-      :class="{ hoverShow: showJoin }"
+      v-if="user && showJoin"
+      :class="{ 'show-picture-on-hover': showJoin }"
       :title="$t('ACTIVITYLIST.ITEM.JOIN')"
     >
       <ProfilePicture
-        :user="hoverUser"
+        :user="user"
         :size="size"
         :is-link="false"
-        @click.native.stop="$emit('join')"
       />
     </div>
   </div>
@@ -32,7 +31,7 @@ export default {
       type: Number,
       default: 20,
     },
-    hoverUser: {
+    user: {
       default: null,
       type: Object,
     },
@@ -64,19 +63,20 @@ export default {
   text-align center
   cursor pointer
 
-  .hoverShow
+  .show-picture-on-hover
     display none
 
-.user-slot-wrapper.active:hover
-  border 0
+// ios safari will require two clicks if we define a hover state
+// so only use the hover thing when the browser can support hover properly
+// see https://css-tricks.com/annoying-mobile-double-tap-link-issue/
+@media (hover)
+  .user-slot-wrapper.active:hover
+    border 0
 
-  .hoverHide
-    display none
-
-  .hoverShow
-    display inline-block
+    .show-picture-on-hover
+      display inline-block
 
 .greyedOut
-  cursor ini
+  cursor default
   border-color lightgrey
 </style>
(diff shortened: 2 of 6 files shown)


src/activities/components/ActivityItem.vue
```diff
@@ -170,6 +170,7 @@ export default {
     },
     detail (event) {
       if (event.target.closest('a')) return // ignore actual links
+      if (event.target.classList.contains('user-slot-wrapper')) return // ignore actual links
```
nicksellen, line 173: I'm trying to understand this bit, which means we could do with more comments (including the `closest('a')` bit, which even though I added, I forgot the reason... 

I recalled it though after thinking/reading a bit more - it's because there are some `<a>` links inside the activity item (e.g. link to the place), and we don't want to trigger the detail view (sidebar activity chat) when we click that link. So I guess the comment is kind of right, but still needed some thinking!

As for the second case though, I was confused, and admittedly I didn't read your main PR comment before reading the code! So I spent some time trying to understand what was going on.

You could use the explanation from your PR description as the comment:

> don't go to activity detail when clicking on an empty slot

But... why are the clicks on the `.user-slot-wrapper` element making it up to here? I thought the `.stop` modifier on the click handler in `UserSlot.vue` should prevent propagation of the event?

I wonder if returning `null` in the case `showJoin` is false prevents the `.stop` from working?

src/activities/components/UserSlot.vue
```diff
@@ -3,18 +3,17 @@
     class="user-slot-wrapper"
     :class="{ greyedOut: !showJoin, active: showJoin }"
     :style="{ width: size + 'px', height: size + 'px' }"
+    @click.stop="showJoin ? $emit('join') : null"
```
nicksellen, line 6: I now understand moving the click handler up here is to achieve _don't go to activity detail when clicking on an empty slot_

I'm not totally convinced that is an improvement - as it reduces the clickable area to get into the chat, and I wouldn't have an expectation that they grey slots are clickable in a different way to the whole item.

On the other hand, I'm totally happy to go with this if you think it's an improvement. My point is only very minor. So please keep it if you want :)

nicksellen (changes requested):
Thanks!

I really appreciate the small naming/style changes, and the CSS fixes. Cleaning up and tidying as you go! I like it :+1: 

Also very happy as it seems like it'll fix the issue nicely! I had got kind of stuck on it :rocket: 

My only real change request is the comment, but also if you can check why the event is propagating when it looks like it shouldn't (to my eyes) that would be great! :star: 

To get the tests to pass, it's the snapshot tests that are failing, I can explain more somewhere else, but for now if you just run `yarn test -u`, and then commit the snapshot file that will have changed, that'll cause the tests to pass.

Regarding https://github.com/yunity/karrot-frontend/pull/2258#issuecomment-769453099:

> Could you do another deploy of my PR?

The deployments are automatic, but only for branches inside our repo (i.e. not for your PR because it's from an external repo), easiest/best thing is we add you to "Karrot Devs" group and you can work on branches in our repo directly, and get branch deployments (see [my message in chat](https://chat.foodsaving.world/channel/karrot-dev?msg=7oDZBNDanR6ZAEvC5) for more info).

src/activities/components/ActivityItem.vue
larzon83: Oh yeah, forgot to add that comment in that line^^ sorry.
About the propagation: I was confused myself that this still propagated. Good point at the `null` thing. Will try again.
About the why of this: for me, it seems not logical that clicking on a element that looks disabled does something (in this case, opening the chat window).

src/activities/components/UserSlot.vue
larzon83: I put it there because I started with the changes from your PR. There you say:
> put the click handler on a higher up element (as the hover element is no longer shown)

I would also prefer to have only a dedicated button "open conversation" and not the whole box clickable. Maybe we can talk about this on Monday...

nicksellen: Haha, I forgot what my changes were :)

Yup, would be happy for UX improvements for activity chat opening. It could lead to a bigger change, so maybe to get this one merged we keep it as you've done it here (making it so the empty slots aren't clickable). What do you think?

larzon83: That's what I had in mind too :)

src/activities/components/ActivityItem.vue
larzon83: Fixed propagation with latest commit. The empty slots have their own component (`EmptySlot.vue`) and there was no `@click.stop` so thats why clicks on them propagated... This is now fixed and I therefore removed the `event.target.classList.contains(...)` in `ActivityItem.vue`.

larzon83:
I will install yarn and check again. I'll update tests too then.
As I am now a member of the repo: should we continue this PR from my fork or shall I create a new one directly in the repo so we can have branch deploys?

larzon83:
@nicksellen Updated snapshot. Also I have installed yarn and did all fresh. With `yarn dev`, it also runs fine with localhost in iOS Simulator.

nicksellen (approved):
Looks good now! Happy for a merge :rocket:

larzon83:
https://dev.karrot.world now works fine on my iPhone 😄

PR #2303, opened 2021-02-03
change(activities): improve design and UX of activities listing

larzon83:
## What does this PR do?

This PR adds a button to the bottom of each activity that opens the chat. Till now, the whole activity is clickable which isn't the best UX. Also Design is improved with more space and better alignment of elements.

### Changes made:
- `tests`:
  - updated snapshots
- `ActivityList.vue`:
  - renamed computed property `displayedActivities` to `displayedActivitiesGroupedByDate`
    - `displayedActivitiesGroupedByDate` builds up a new array with dates and inside that, holds the corresponding activities
    - instead of "slice()", a `for ... of` loop is used in conjunction with `break`
  - added new prop `dense` to be able to display the list of activities without groupedByDate headers -> the date will be shown inside the card instead (used eg by `JoinedActivities.vue` and `AvailableActivities.vue`)
- `ActivityItem.vue`
  - Activity Boxes are now grouped by date rather then having the date inside each box (see `ActivityList.vue`)
  - The whole box isn't clickable anymore
  - Adjusted margins and paddings
  - Removed the red/white background gradient for activities with only empty slots
  - activityType icon is now in top right
  - activityType icon now uses quasar tooltip instead of html title -> there is no hover on mobile and therefor no title...
  - added a new section in the bottom of the card with a button to start the chat
    - the button is full width on mobile
    - and not full width and aligned to the right on desktop

### Screenshots and Video

Here are some screenshots showing the before and after for mobile and desktop. Also a video for better feeling :)
activities-1-mobile-oldactivities-2-mobile-new

activities-3-desktop-oldactivities-4-desktop-new

upcoming-activities-1-old upcoming-activities-2-new

https://user-images.githubusercontent.com/23213583/106805667-475eec80-6667-11eb-95aa-5cce48596596.mp4

## Checklist

- [ ] no unrelated changes

### DIFF
--- a/src/activities/components/ActivityItem.vue
+++ b/src/activities/components/ActivityItem.vue
@@ -1,125 +1,166 @@
 <template>
-  <QCard
-    :class="{ full: activity.isFull }"
-    @click.native.stop="detail"
-  >
-    <QCardSection
-      class="no-padding content"
-      :class="{ isEmpty: activity.isEmpty, isUserMember: activity.isUserMember, isDisabled: activity.isDisabled }"
+  <div>
+    <div
+      v-if="activity.dateGroupedHeadline"
+      class="q-px-sm q-pt-lg full-width date-grouped-headline"
     >
-      <div class="q-pa-sm full-width">
-        <div>
-          <QIcon
-            v-if="activity.activityType"
-            v-bind="activity.activityType.iconProps"
-            size="xs"
-            class="q-pr-xs"
-            style="position: relative; bottom: 4px;"
-          />
-          <span class="featured-text">
-            {{ $d(activity.date, 'hourMinute') }}
-            <template v-if="activity.hasDuration"> &mdash; {{ $d(activity.dateEnd, 'hourMinute') }}</template>
-          </span>
-          <template v-if="placeLink">
-            <strong v-if="activity.place">
-              <RouterLink :to="{ name: 'place', params: { placeId: activity.place.id }}">
-                {{ activity.place.name }}
-              </RouterLink>
-            </strong> {{ $d(activity.date, 'dateWithDayName') }}
-          </template>
-          <template v-else>
-            {{ $d(activity.date, 'dateLongWithDayName') }}
-          </template>
-          <span>
-            <strong>{{ $t('CONVERSATION.OPEN') }} <QIcon name="chat" /></strong>
-          </span>
+      {{ activity.dateGroupedHeadline }}
+    </div>
+    <QCard
+      :class="{ full: activity.isFull }"
+    >
+      <QCardSection
+        class="no-padding content"
+        :class="{ isEmpty: activity.isEmpty, isUserMember: activity.isUserMember, isDisabled: activity.isDisabled }"
+      >
+        <div class="content-inner">
+          <div class="row no-wrap items-start justify-between">
+            <div>
+              <strong class="featured-text">
+                {{ $d(activity.date, 'hourMinute') }}
+                <template v-if="activity.hasDuration"> &mdash; {{ $d(activity.dateEnd, 'hourMinute') }}</template>
+              </strong>
+              <template v-if="placeLink">
+                <span v-if="activity.place">
+                  <RouterLink :to="{ name: 'place', params: { placeId: activity.place.id }}">
+                    {{ activity.place.name }}
+                  </RouterLink>
+                </span>
+              </template>
+              <template v-else>
+                {{ $d(activity.date, 'dateLongWithDayName') }}
+              </template>
+            </div>
+            <QIcon
+              v-if="activity.activityType"
+              v-bind="activity.activityType.iconProps"
+              title=""
+              size="xs"
+              class="q-ml-sm"
+              style="top: 1px;"
+            >
+              <q-tooltip v-if="activity.activityType.iconProps && activity.activityType.iconProps.title">
+                {{ activity.activityType.iconProps.title }}
+              </q-tooltip>
+            </QIcon>
+          </div>
+          <div
+            v-if="activity.isDisabled"
+            class="q-my-xs"
+          >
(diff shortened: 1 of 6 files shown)


brnsolikyl:
Looks much better now with all the information much better organized!

Only one thing I have an opinion on:
- I prefer the activity icon to the left. I'm not sure why, but I think it's because identifying the type of activity is one of the first things I'd like to do. But it could also be me used to the old layout. Is there any particular reason for moving it right?

Thx for the work!

nicksellen:
Beautiful!

Big improvement to me, you fixed the guilt I have had ever since I added even more noise into the `ActivityItem` component when I added the activity types!

And I like the grouping by date, the current view seems quite hard to scan down to see the dates as they are so small in the title area.

Unlike @brnsolikyl I actually prefer the activity type icons on the right, as you have them. Maybe a bit more balanced (with everything on the left they might fall over).

There is quite an increase in vertical space used, but I think for most groups they don't have _so_ many activities that it would cause a problem. And I hope they will tell us if it does...

I'll wait with a code review until you're ready (mark as ready for review and request from me).

larzon83:
> Looks much better now with all the information much better organized!
> 
> Only one thing I have an opinion on:
> 
> * I prefer the activity icon to the left. I'm not sure why, but I think it's because identifying the type of activity is one of the first things I'd like to do. But it could also be me used to the old layout. Is there any particular reason for moving it right?
> 
> Thx for the work!

The decision to put the icon to the right is to make the card more balanced like @nicksellen already mentioned. Having it on the left results in too much information put in one place. We also thought about putting icons before the time and place, but this would still be too much. If you scroll the page, we think it's now really good for screening.
But this is also a subjective thing 😄 
Do you have any hard feelings about it? If we are happy with the code (not yet), I will hand it over to Nick for review. Then he could make a branch deploy and you could test it in real...

nicksellen:
Branch deployments happen automatically for every branch :) (the only step that is manual is adding the subdomain to the https cert - as we don't have a setup where we can do acme wildcard certs yet... but I just did the manual setup), so --> https://change-open-activity-chat.dev.karrot.world/

One think I noticed from that is the "upcoming activities" on the top of the wall page take up a lot of vertical space now, mostly because of the grouped dates.

brnsolikyl:
I'm fine with the icon on the right! :)

larzon83:
> One think I noticed from that is the "upcoming activities" on the top of the wall page take up a lot of vertical space now, mostly because of the grouped dates

Improved. See attached screens in PR description.

src/activities/components/ActivityItem.vue
```diff
@@ -181,28 +222,36 @@ export default {
...
+      display inline
+      margin-right .3em
+
+.conversation-section
+  font-weight 500
+  color $secondary
+  box-shadow 0 -1px 0 rgba(0, 0, 0, 0.06)
+
+.q-btn.open-conversation-button
+  ::v-deep .q-btn__wrapper
```
nicksellen, line 248: Minor style: we have used `>>>` elsewhere which works as an alias for `::v-deep`

src/activities/components/ActivityList.vue
```diff
@@ -246,3 +273,10 @@ export default {
   },
 }
 </script>
+
+<style scoped lang="stylus">
+.date-with-day-name
+  font-weight 700
+  color rgba(0, 0, 0, 0.7)
+  text-align center
```
nicksellen, line 281: You can use quasar classes `text-center` and `text-bold` here (we like to use the quasar classes where possible, which you've done in most places :+1:)

(and as there would then be only 1 CSS value left, could perhaps be inline in a style attribute?)

src/activities/components/ActivityList.vue
```diff
@@ -82,16 +82,28 @@
...
-        @join="$emit('join', arguments[0])"
-        @leave="$emit('leave', arguments[0])"
-        @detail="$emit('detail', arguments[0])"
-      />
+      <template
+        v-for="(day, index) in displayedActivitiesGroupedByDate"
+      >
+        <div
+          v-if="!dense"
+          :key="`day-${index}`"
```
nicksellen, line 90: Minor: I think `day.date` would make a better `key`, as it's related to the data, not just the position in the array (we live update a lot of stuff via websockets, so the contents can change - although not sure about this particular case).

nicksellen, line 90: Minor: I think `day.date` would make a better `key`, as it's related to the data, not just the position in the array (we live update a lot of stuff via websockets, so the contents can change - although not sure about this particular case).

nicksellen (approved):
Looks great! Thanks :star:

I really appreciate that it looks like code that is already in our project, you've done well making it fit to our style/conventions.

I added a few minor comments, none of essential, so I approved this PR, but left them there if you feel like doing them.

Now it is approved, I _think_ you should be able to merge this yourself, when you are ready.

src/activities/components/ActivityList.vue
larzon83: If someone would add two activities - one this year and one on the same date a year into the future, then two keys would be the same:
- Tue, February 9, 2021
- Tue, February 9, 2022

This would result in two :keys being `Tue, February 9` each. And Vue would scream :)

Also, if new data arrives, the computed prop that is being looped over would recalculate and so are the keys of the `v-for`. It is save to do this ;)

larzon83: see comment above...

src/activities/components/ActivityItem.vue
larzon83: Ah, in our project at work `>>>` doesn't work. Changed it.

src/activities/components/ActivityList.vue
larzon83: done

PR #2309, opened 2021-02-22
fix zoom level for one active place

### LINKED ISSUE #2290: Curious map zoom level when switching the toggle to filter location by status
When I toggle the switch to filter locations by status, the small map in the top left corner is updated accordingly. There seems to be a problem in the computation of the zoom level and position of this map. See the video below, where the map is scaled back to show all of Europe even if all the locations are concentrated in a small town:

Peek 11-01-2021 14-03

This happens with Firefox 78 ESR on Linux.

pogopaule:
Closes #2290

## What does this PR do?

Removes condition (`!this.hasOneMarker`) in if-statement that causes bug.

If `!this.hasOneMarker` is present, the `bounds ()` method returns `undefined`.

This causes the [`zoom ()`](https://github.com/yunity/karrot-frontend/blob/66fac012705b9607ecd629aec14fe05e7e51d825/src/maps/components/StandardMap.vue#L226) method to return `undefined` which results in the 'zoomed out' bug described in #2290.

**Important note for the reviewer**: Digging through the git history and using the frontend I could not figure out why `!this.hasOneMarker` was part of the condition in the first place. To me it does not make any sense, but maybe there is an edge-case somewhere where it does?

fix-zoom-level

## Checklist

- [x] no unrelated changes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,6 +28,7 @@ Please document your changes in this format:
 
 ### Fixed
 - activities: only hover pictures if device supports it @larzon83 [#2297]
+- set map bounds correctly if there is only on place @pogopaule [#2290]
 
 ## [9.1.0] - 2020-01-06
 ### Fixed
--- a/src/maps/components/StandardMap.vue
+++ b/src/maps/components/StandardMap.vue
@@ -197,7 +197,7 @@ export default {
     bounds () {
       if (this.forceBounds) return this.forceBounds
       if (this.forceCenter && !Number.isNaN(this.forceCenter.lat)) return null
-      if (!this.preventZoom && this.hasMarkers && !this.hasOneMarker) {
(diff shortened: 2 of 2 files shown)


CHANGELOG.md
```diff
@@ -28,6 +28,7 @@ Please document your changes in this format:
 
 ### Fixed
 - activities: only hover pictures if device supports it @larzon83 [#2297]
+- set map bounds correctly if there is only on place @pogopaule [#2290]
```
larzon83, line 31: `one` place... 😄

pogopaule: eagle eye :1st_place_medal:

pogopaule:
I close this PR because I just realized that it does not fix the bug as described in #2290. If there is no active place the map still has the zoom out bug. I will reopen the PR if I have fixed it.

PR #2317, opened 2021-03-16
Landing page new words

brnsolikyl:
Branch deployment: https://landing-page-new-words.dev.karrot.world

## What does this PR do?

Rewrites a few crucial bits of the landing page. Quick PR before new deployment

## Checklist

- [ ] no unrelated changes


nicksellen (approved):
Looks good! If you think it's ready, feel free to merge :)

One thing I did notice is the `karrot-screenshot.png` image has increased from 380kb to 466kb which is quite a bit bigger, so if it's possible to reduce the size that could be useful, but not essential.

brnsolikyl:
> One thing I did notice is the `karrot-screenshot.png` image has increased from 380kb to 466kb which is quite a bit bigger, so if it's possible to reduce the size that could be useful, but not essential.

Size reduced now to 178,3 Kb

I'm going to add that About button component, then I'll merge! :)

brnsolikyl:
I feel it's ready to merge now after adding the About Karrot component. But maybe a quick check before merging would be good, @nicksellen , just to make sure I didn't do anything ugly

brnsolikyl:
Not sure what happened there. I ran `yarn test -u storyshots`, it got updated but still not passing the test...

src/base/components/KAbout.vue
```diff
@@ -11,6 +11,22 @@
...
+          rel="noopener noreferrer"
+          href="https://community.foodsaving.world/t/how-to-get-involved-onboarding-into-the-karrot-team/661"
+          target="_blank"
+        >
+          <QItemSection side>
+            <QIcon name="fas fa-users" />
+          </QItemSection>
+          <QItemSection>
+            <QItemLabel>
+              How to Get Involved
```
nicksellen, line 25: Text should be put into translation system.

src/base/pages/Landing.vue
```diff
@@ -112,6 +112,21 @@
           href="https://github.com/yunity/karrot-frontend"
         >GitHub</a>
       </i18n>
+
+      <p class="text-center q-mt-lg">
+        <QBtn
+          @click="toggleAbout"
+        >
```
nicksellen, line 119: The button doesn't look quite fitting in style to me. The simplest change I think it just to add `unelevated` attribute. Still looks a bit different from the other buttons, but I think it's OK.

aboutbutton

src/locales/locale-en.json
```diff
@@ -18,7 +18,7 @@
       },
       "DEMOCRATIC": {
         "DESCRIPTION": "Karrot is made by its users: a free and open-source software developed with the feedback and participation of the people who use it. It is mainly developed by a small core team that is constantly in contact with people in different groups, in order to make sense of their diverging and converging needs and thus guide the development. We abide by the ethics of integrity and non-commercialization of data.",
-        "DESCRIPTION2": "Developers and designers are obviously welcome to participate in its development. Check out the project on {code}.",
+        "DESCRIPTION2": "Developers and designers are obviously welcome to participate in its development.",
```
nicksellen, line 21: As this line no longer uses the `{code}` variable, it can be removed from the usage too.

So this:

https://github.com/yunity/karrot-frontend/blob/18cff7b940f66bddb3f80e0f5baf40e336b87dcc/src/base/pages/Landing.vue#L106-L114

and be changed to:

```html
<p v-t="'ABOUT_KARROT.SECTIONS.DEMOCRATIC.DESCRIPTION2'" />
```

nicksellen (changes requested):
Looks good! A couple of small things left though...

... regarding the tests, it fails for me locally as it is now, and `yarn test -u storyshots` fixes it here. Maybe you forgot to check in the changed files?

src/base/components/KAbout.vue
nicksellen: @brnsolikyl mentioned that it maybe makes sense to keep it in English, given the "team" is only communicating in English. Makes sense to me, but would also include a code-level comment to explain why it's not translated, e.g.:

```html
<!-- this is in English because -->
```

src/base/pages/Landing.vue
brnsolikyl: I added unelevated. For some reason I like it being different :) But I don't mind changing to fit the style
This button will be temporary anyhow if we go on to create the about page and improving the landing page

brnsolikyl:
> .. regarding the tests, it fails for me locally as it is now, and `yarn test -u storyshots` fixes it here. Maybe you forgot to check in the changed files?

How would I do that? I ran it and committed the first time. Now, whenever I run it, it passes the test and no files are changed.

nicksellen:
I think your dependencies are not up to date locally, and the newer versions (e.g. quasar) generate slightly different html, hence the change in the snapshots across many different components unrelated to the change.

To updated dependencies just run `yarn`... then re-run tests / snapshot updating. I think that might be it :crossed_fingers:

nicksellen (approved):
Great, all looks good to me now! :+1:

PR #2319, opened 2021-03-23
feat: allow to dismiss feedback

### LINKED ISSUE #1508: Hide "give feedback for pickup" on group wall
Andrea from Solikyl Gothenburg suggested:
> Possibility to choose "Do not leave feedback" on the pickup feedback so it disappears from one's main Karrot page.

There's already a notification about this in the "bell notifications" menu, so to keep it simple we might as well remove the wall notification completely.

~Same goes for the applications notification when we added the "Applications" sidenav entry.~ done

pogopaule:
Closes #1508

## What does this PR do?

This PR adds a new button `Dismiss` to the activity feedback form. After dismissing, the user won't be able to give feedback to this activity.

## Links to related issues

Related backend PR https://github.com/yunity/karrot-backend/pull/1124

## Checklist

- [ ] no unrelated changes


src/activities/datastore/activities.js
```diff
@@ -136,6 +138,9 @@ export default {
         dispatch('fetchListByGroupId', groupId)
       }
     },
+    async dismissFeedback ({ commit, rootGetters }, activityId) {
+      await activities.dismissFeedback(activityId)
```
pogopaule, line 142: @larzon83 Could this call also be made asynchronously, i.e. without the await?

src/feedback/components/FeedbackForm.vue
```diff
@@ -33,6 +33,7 @@
       class="text-negative"
       style="margin-top: 3em"
     >
+      <i class="fas fa-exclamation-triangle" />
```
larzon83: good catch. Had that on my todos too :)

src/activities/datastore/activities.js
larzon83: Yes, that is possible (in this case): What's your motivation behind this?

pogopaule: We are not processing the request's response, but having await might give the impression that it should be processed. Other developers could be confused by this. So reducing sources for potential future confusion is by motivation.

larzon83: Yes, thats correct. It's ok for me to remove the async. Maybe we can leave this comment open for Nick to check in too on this.

pogopaule:
A tiny bug we need to fix:
After dismissing three feedbacks I noticed that the number of open feedbacks in the menu does not get updated automatically. A page reload is necessary.


feedback_bug

larzon83:
> A tiny bug we need to fix:
> After dismissing three feedbacks I noticed that the number of open feedbacks in the menu does not get updated automatically. A page reload is necessary.

fixed

nicksellen:
Quick point for the moment, the branch deployment is broken because of my mistake, you can `git cherry-pick 1e5a33a8e46f9ab93b64fa6eb15acc8e6b8c47ea` if you want to bring the fix into your branch :)

src/activities/datastore/activities.js
nicksellen: I think it's nice with the `await`, it means the vuex action will only be completed after the request is done, I think that is a nice principle.

If removing the `async/await` then returning the result of the API call (a promise) would also achieve that, but the `await` is more consistent IMO.

src/feedback/components/FeedbackForm.vue
```diff
@@ -37,24 +37,62 @@
...
+          v-close-popup
+          flat
+          color="primary"
+          :label="$t('BUTTON.CANCEL')"
+        />
+        <QBtn
+          v-close-popup
+          flat
+          color="primary"
+          data-autofocus
```
nicksellen, line 90: We use just plain `autofocus` attribute in most places (a couple of uses of `data-autofocus`), which is best?

I see this is probably copied from me using `data-autofocus` in https://github.com/yunity/karrot-frontend/commit/647b877af#diff-a911f8aaf90e43fc2c4adac22b9d5dfa9070db82e1d75a91c640befb99e96178R84 but I don't remember why :)

src/feedback/components/FeedbackForm.vue
```diff
@@ -64,12 +102,14 @@ import {
   QIcon,
 } from 'quasar'
 import AmountPicker from './AmountPicker'
+import CustomDialog from '@/activities/components/CustomDialog'
```
nicksellen, line 105: Now this `CustomDialog` component is being used outside of the `activities` module, it could go into our general shared components place (`src/utils/components` seems appropriate)

src/feedback/components/FeedbackForm.vue
```diff
@@ -86,18 +126,21 @@ export default {
...
+      isDismissFeedbackDialogVisible: false,
+    }
+  },
   computed: {
     canSave () {
-      if (!this.isNew && !this.hasChanged) {
-        return false
-      }
-      return true
+      return this.isNew || this.hasChanged
```
nicksellen, line 136: Haha, so much nicer than the previous convoluted logic! .. I guess the previous logic started off more complex, then got reduced into a really odd construct. :+1:

src/locales/locale-en.json
```diff
@@ -116,6 +116,13 @@
     "TITLE": "Feedback",
     "TOGETHER_WITH": "You did this together with:"
   },
+
+  "ACTIVITY_FEEDBACK_DISMISS": {
```
nicksellen, line 120: I think this would be better inside the `ACTIVITY_FEEDBACK` object, doesn't feel like a big enough concern to justify a new top level object...

So, for example, could be:
- `ACTIVITY_FEEDBACK.DISMISS_LABEL`
- `ACTIVITY_FEEDBACK.DISMISS_TITLE`
- `ACTIVITY_FEEDBACK.DISMISS_MESSAGE`

(also making them less coupled to the UI components, ~~button~~, ~~prompt~~)

src/activities/datastore/activities.js
nicksellen: ... also I guess you decided to rely on the websocket update to keep the data updated.

Personally, I still update from the request response, but I'm also very happy with it like this - happy to revisit this in a more general way when rethinking the data flow with vue3 and the composition API stuff...

nicksellen (changes requested):
Very nice! Really happy to see this, and the code looks very much inline with what is already there. So great job!

My comments/suggestions are very minor really...

src/activities/datastore/activities.js
larzon83: When this action gets dispatched from the button, we don't process the returned result. So, I think we could remove the async/await and let the action do its thing – I tested it and it works, too. When the backend has finished its stuff, the frontend gets updated via websocket...
Could we even get a returned value with this `actionsToEvents` thing?
Anyway, I myself would also vote for keeping async/await. Also for consistency with other actions.

And yes, we choose the websocket way to keep the data updated. We were lazy refactoring the backend to return the processed activity instead of returning only the status code 😄

src/feedback/components/FeedbackForm.vue
pogopaule: Good point, I was wondering why it was in activities in the first place.

larzon83:
@nicksellen all done now. But coverage is decreased. Don't know how to fix. What should we do?
Should we generally discuss testing inside a monday meeting?

nicksellen (approved):
Looks great, thanks! :100:

PR #2324, opened 2021-04-01
Update CHANGELOG.md

brnsolikyl:
## What does this PR do?

updates changelog with this contribution: #2317

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,8 @@ Please document your changes in this format:
 ```
 
 ## [Unreleased]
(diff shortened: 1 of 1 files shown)


nicksellen (approved):
Thanks! I got sloppy at checking for changelog entries before approving... although I have been good at going through the merged PRs to add them in before releases... better to do it as we go along though.

brnsolikyl:
I'm also not used to it and realized afterwards that I missed it. But the basic procedure is to update the changelog in the same PR, right?

PR #2330, opened 2021-04-07
fix(ActivityHistoryStatistics): replace hard coded string

pogopaule:
## What does this PR do?

The [translation key](https://github.com/yunity/karrot-frontend/blob/master/src/locales/locale-en.json#L681) is already present but was not used.
See https://www.transifex.com/yunity-1/karrot/translate/#de/frontend/300401807

## Checklist

- [x] no unrelated changes


pogopaule:
Is a changelog entry required for this?

nicksellen (approved):
Ah, good catch, thanks! I think we can leave out the changelog entry here...

PR #2333, opened 2021-04-07
Improve calculation of group gallery map bounds

### LINKED ISSUE #2316: Improve method for calculating map bounds
Currently for the group gallery view we try and calculate the usable bit of the map, depending on whether the group list is open or not (the map fills the whole width of the screen but only part of it is usable). There is a better way using `padding-top-left` option in the `l-map` component (via options to `fitBounds` in vue2leaflet), but it has been broken.

It is now fixed (see https://github.com/vue-leaflet/Vue2Leaflet/pull/656), but not released yet. When it is released, then we can rework it a bit.

The issue can be seen in the screenshot below. Ideally the map would be fitting the markers nicely in the available space, we don't need to see any further east than Poland for this data, and Bilbao is a bit squished up on the left.

Screenshot from 2021-03-11 15-44-47

nicksellen:
Closes #2316 

Branch deployment: https://change-improve-map-bounds.dev.karrot.world/

## What does this PR do?

We can now make use of the new `paddingTopLeft` prop in calculating the map bounds for the group gallery using the actual pixel width of the sidebar (previously it only looked about right if using the same screen resolution as I had when I wrote it!).

Here's an example:

mapboundsnew

## Checklist

- [x] no unrelated changes

### DIFF
--- a/src/groupInfo/components/GroupGalleryMap.vue
+++ b/src/groupInfo/components/GroupGalleryMap.vue
@@ -5,6 +5,7 @@
     :force-center="forceCenter"
     :force-zoom="forceZoom"
     :force-bounds="forceBounds"
+    :padding-top-left="paddingTopLeft"
     :show-attribution="false"
   />
 </template>
@@ -32,6 +33,10 @@ export default {
     myCoordinates: {
       default: null,
       type: Object,
+    },
+    paddingTopLeft: {
+      default: () => ([0, 0]),
+      type: Array,
     },
   },
   computed: {
@@ -61,15 +66,7 @@ export default {
       if (this.myCoordinates) {
         coordsForBounds.push(this.myCoordinates)
       }
-      const bounds = L.latLngBounds(coordsForBounds).pad(0.2)
-      if (this.offset.lat !== 0 || this.offset.lng !== 0) {
-        // we have an offset!
-        // make the bounds extended to incorporate the offset into the calculation
-        const sw = bounds.getSouthWest()
-        const offsetPoint = L.latLng(sw.lat + this.offset.lat, sw.lng + this.offset.lng)
-        return bounds.extend(offsetPoint)
-      }
-      return bounds
+      return L.latLngBounds(coordsForBounds).pad(0.2)
     },
     singleGroup () {
       if (this.groupsWithCoordinates.length === 1) {
@@ -90,12 +87,6 @@ export default {
       }
       return null
     },
-    offset () {
-      if (window.innerWidth > 767 && this.$q.platform.is.desktop) {
-        return this.expanded ? { lat: 0, lng: -0.4 } : { lat: 0.05, lng: 0 }
-      }
-      return { lat: 0.0, lng: 0.0 }
-    },
   },
   methods: {
     hasCoordinates (item) {
@@ -112,4 +103,11 @@ export default {
 </script>
 
 <style scoped lang="stylus">
+.debug
+  position fixed
+  top 40px
+  right 40px
+  z-index 10000000
+  padding 20px
+  background-color rgba(255, 255, 255, 0.5)
 </style>
--- a/src/groupInfo/components/GroupGalleryMap.vue
+++ b/src/groupInfo/components/GroupGalleryMap.vue
@@ -54,9 +54,9 @@ export default {
       const groups = this.groupsWithCoordinates.filter(this.hasDistance)
       if (groups.length === 0) return []
       // Any group with max 3x distance of the closest group
-      // Minimum "closest distance" is 30km, so it won't filter for unhelpful
+      // Minimum "closest distance" is 10km, so it won't filter for unhelpful
       // distances like <3km if you happen to be 1km from a group...
-      const closestDistance = Math.max(groups[0].distance, 30)
+      const closestDistance = Math.max(groups[0].distance, 10)
       const maxDistance = closestDistance * 3
       return groups.filter(group => group.distance < maxDistance)
     },
--- a/src/groupInfo/components/GroupGalleryMap.vue
+++ b/src/groupInfo/components/GroupGalleryMap.vue
@@ -101,13 +101,3 @@ export default {
   },
 }
 </script>
-
-<style scoped lang="stylus">
-.debug
-  position fixed
-  top 40px
-  right 40px
-  z-index 10000000
-  padding 20px
-  background-color rgba(255, 255, 255, 0.5)
-</style>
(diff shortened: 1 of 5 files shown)


pogopaule:
If I go to https://change-improve-map-bounds.dev.karrot.world/#/groupPreview, I see the following 

zoomed_in

But what I expect is a zoom level in which all markers are in view, something like this

should_be_zoom

But I am not really sure what ought to be...

src/maps/components/StandardMap.vue
```diff
@@ -160,6 +162,10 @@ export default {
       type: Object,
       default: null,
     },
+    paddingTopLeft: {
+      type: Array,
+      default: null,
```
pogopaule, line 167: Why is default `null` but not `() => ([0, 0])`?

src/groupInfo/components/GroupGalleryUI.vue
```diff
@@ -8,12 +8,17 @@
       :filtered-my-groups="filteredMyGroups"
       :filtered-other-groups="filteredOtherGroups"
       :expanded="expanded"
-      :my-coordinates="myCoordinates"
+      :my-coordinates="false"
```
pogopaule, line 11: Checking out this branch and running it on my machine I get the following error in the console

```
[Vue warn]: Invalid prop: type check failed for prop "myCoordinates". Expected Object, got Boolean with value false.

found in

---> <GroupGalleryMap> at src/groupInfo/components/GroupGalleryMap.vue
       <GroupGallery> at src/groupInfo/components/GroupGalleryUI.vue
         <ConnectGroupGallery> at src/groupInfo/pages/GroupGallery.vue
           <QPage>
             <QPageContainer>
               <QLayout>
                 <MainLayout> at src/base/pages/MainLayout.vue
                   <App> at src/App.vue
                     <Root>
```

src/groupInfo/components/GroupGalleryUI.vue
```diff
@@ -213,8 +221,19 @@ export default {
     hasOtherGroupsToShow () {
       return this.expanded && this.filteredOtherGroups.length > 0
     },
+    paddingTopLeft () {
+      // when the sidebar is open on a wide enough desktop screen
+      // we set the padding so the markers aren't displayed underneath the sidebar
+      if (this.expanded && window.innerWidth > 767 && this.$q.platform.is.desktop) {
```
pogopaule, line 227: Where does `767` come from? Maybe it could be a named constant or something similar?

src/groupInfo/components/GroupGalleryUI.vue
```diff
@@ -182,6 +189,7 @@ export default {
   },
   data () {
     return {
+      width: -1,
```
pogopaule, line 192: Why `-1` as a default?

src/groupInfo/components/GroupGalleryMap.vue
```diff
@@ -61,15 +66,7 @@ export default {
       if (this.myCoordinates) {
         coordsForBounds.push(this.myCoordinates)
       }
-      const bounds = L.latLngBounds(coordsForBounds).pad(0.2)
-      if (this.offset.lat !== 0 || this.offset.lng !== 0) {
```
pogopaule, line 65: Just for my understanding: Was this the code that emulated the paddingTopLeft that is now fixed?

pogopaule (changes requested):

brnsolikyl:
> what I expect is a zoom level in which all markers are in view, something like this

Not sure if I'm correctly addressing your question, but we actually expect the map to focus on local groups first and foremost and not include all the markers (I can't remember the criteria Nick used). On dev it will look a bit weird since there's not a lot of groups. But check it on production, it should make more sense. The idea at least is that it will make it easier for people to find groups nearby them or connect to other groups if there are many.

nicksellen:
Ah yeah you're encountering another intended behaviour to emphasis local groups. It uses geoip and some logic to fit in local groups with a certain range if there are any.

It's possible to run locally and fake the ip using an environment variable if you wanted to try it to the next level :)

It's definitely something that would benefit from tests....

src/groupInfo/components/GroupGalleryUI.vue
nicksellen: Oh yes, I was going use it somewhere to check if it was uninitialised... I'll have a look again! Not entirely sure how much it's possible to rely on the resize Oberserver to initialize the value...

src/groupInfo/components/GroupGalleryUI.vue
nicksellen: That one I just left as it was before. I wondered where it came from before... I can have a look. And yes, at least a constant with some hint what it means.

src/groupInfo/components/GroupGalleryUI.vue
nicksellen: Ah, now I see I just changed it during dev to simplify the logic, reverted it back to how it is supposed to be now.

src/groupInfo/components/GroupGalleryUI.vue
nicksellen: I couldn't understand where it came from, it doesn't correspond to quasar breakpoints, also isn't a reactive value, and so I removed it entirely... seems to still work ok :)

src/maps/components/StandardMap.vue
nicksellen: I checked the leaflet library and they use `null` as default, so I switched to that everywhere now as default.

src/groupInfo/components/GroupGalleryUI.vue
nicksellen: I commented it now, and added in a check for `!== -1` where it gets used.

src/groupInfo/components/GroupGalleryMap.vue
```diff
@@ -112,4 +103,11 @@ export default {
 </script>
 
 <style scoped lang="stylus">
+.debug
```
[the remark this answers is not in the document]
nicksellen: Removed now :+1:

nicksellen:
Ok, I fixed it all up now! I pressed the _resolve_ button myself on each comment, but if in general you'd prefer to do it yourself, I can leave them next time.

I also snuck in a change to reduce the minimum distance calculation in the "nearby groups" calculation to 10km (from 30km). I think I saw a case where groups further away than meaningful were being displayed (it shows up to 3x the distance, so with 30km minimum it would show some up to 90km away)...

pogopaule:
Looks good to me! Well done! :) I leave the pleasure of pressing the merge button to you .

pogopaule (approved):

nicksellen:
Great, thanks for the review! :+1:

PR #2348, opened 2021-04-20
change(amount-picker): improve UX of weight slider limits

### LINKED ISSUE #2338: Feedback weight slider limits
Currently the slider limit is on 70 kg as the screenshot shows:
image

**Problem**

I heard this first-hand from a conversation between two users. One wanted to enter a higher value than 70 kg by using the slider but thought it was not possible and believed it was a hard limit. The other explained then that a higher value could be entered at the weight field next to it.

**Any solutions?**

Maybe a small hint could pop up when you slide to the limit indicating somehow that the value can be entered manually? Any ideas?

larzon83:
Closes #2338

## What does this PR do?

`AmountPicker`:
- add a grey and rounded box around the `AmountPicker`
- increase max slider value to `100`
- if weight slider is on the max right (currently `100`), an info-bar is displayed underneath the slider that informs the user that higher values are possible by directly writing to the input-box

`FeedbackForm`:
- make textarea outlined

### Screenshots
increase-weight-info-desktop-1 increase-weight-info-desktop-2

increase-weight-info-mobile-1 increase-weight-info-mobile-2

## Checklist

- [ ] no unrelated changes


brnsolikyl:
Definitely an improvement!

src/feedback/components/AmountPicker.vue
```diff
@@ -47,25 +47,33 @@
         <QSlider
           :value="limitedValue"
           :min="0"
-          :max="70"
+          :max="100"
```
nicksellen, line 50: The `100` is repeated in 3 places now, so would justify being put into a constant somewhere (or a value in the vue `data`, given we can't use a variable directly in the vue template...).

src/feedback/components/AmountPicker.vue
```diff
@@ -47,25 +47,33 @@
...
           :step="0.5"
           label
-          class="self-center q-mx-md"
+          class="self-center q-mr-lg"
+          style="margin-left: 8px"
```
nicksellen, line 54: Could we use a `q-ml-?` classname here? (or maybe `q-my-lg` if they can be the same (or do I mean `q-mx-lg` I get confused between x and y))

src/feedback/components/AmountPicker.vue
```diff
@@ -47,25 +47,33 @@
...
             <span class="text-caption">kg</span>
           </template>
         </QInput>
       </div>
     </div>
+    <div
+      class="content-more-weight-info"
+      :class="{'is-visible': isOnMaxOfSlider}"
+    >
+      <span v-if="isOnMaxOfSlider">{{ $t('ACTIVITY_FEEDBACK.INCREASE_AMOUNT_HINT') }}</span>
```
nicksellen, line 75: How come we need the `is-visible` _and_ the `v-if`? Can we not `v-if` the whole `div`? (or conversely always show the HINT.. assuming the `is-visible` takes care of showing it or not). Maybe I'm missing something.

src/feedback/components/AmountPicker.vue
```diff
@@ -122,10 +133,32 @@ export default {
...
+  text-align center
+  background hsl(0deg 0% 97%)
+  border-bottom-right-radius 13px
+  border-bottom-left-radius 13px
+
+  &.is-visible
+    background $info
+
+.showOverlay .content,
+.showOverlay .content-more-weight-info
```
nicksellen, line 161: This is quite a lot of custom css here, where at least some of it is available via quasar classes, e.g. `text-center` (I think!), some margins, paddings, and maybe `bg-grey-<some number>`, and there is duplication which might be nice in a variable (the border radius stuff, I wonder if there are already some quasar variables for radius too?).

(in general we try and minimise the custom css and so far it was one of the least maintainable parts, and hard to reuse given we lack a coherent design system, beyond that which quasar provides)

nicksellen (changes requested):
Great, definitely an improvement!

I wonder if the input box could even be part of the message box and not have to refer people to the input box, but just have it right there? I am fine with how it is here though.

The points I added are quite minor really, but if you can see some ways to address them, would be great!

src/feedback/components/AmountPicker.vue
pogopaule: We removed `is-visible` and used the quasar class `bg-info`. What do you think?

larzon83:
> I wonder if the input box could even be part of the message box and not have to refer people to the input box, but just have it right there? I am fine with how it is here though.

You mean something like this?
Bildschirmfoto 2021-04-22 um 12 53 14

I kind of like the idea.

nicksellen:
> You mean something like this?

Yup, exactly!

I think the term "input box" might be too technical too, perhaps something like: _"Need to add more weight? Type it in!"_ is clearer? (especially if the box is right there)

src/feedback/components/AmountPicker.vue
nicksellen: (Ah, also now I tried it out, I see the reason for not having the `v-if` around the whole thing - to reserve the space for the info message so it doesn't make the content below it jump when it's visible)

The `bg-info` seems a good solution!

larzon83:
UX is a hard topic 😄 
I'll ask Kristin...

larzon83:
@nicksellen @brnsolikyl 
We talked with Kristin and and for her there were too many elements to look at and interact with.
So, we did another refactor with the input-field as the main element to interact with. We removed the slider and the libra(?). The interface is now much cleaner and easier to understand. Also no extra info needed.

We really like this version. What do you think?

### Screenshots

weight-desktop-1 weight-desktop-2 weight-desktop-3 weight-mobile

larzon83:
@nicksellen @brnsolikyl 
Side question: is the "(in kg)" really needed inside the headline. There is "kg" already inside the input-field...
If we can remove it, should we remove it from all locale files?

brnsolikyl:
It looks nice without the slider. I'm not sure really if people would prefer to use it and will miss it. I personally think it's just fine without it and indeed looks much cleaner.

> Side question: is the "(in kg)" really needed inside the headline. There is "kg" already inside the input-field...
> If we can remove it, should we remove it from all locale files?

Yep! It's really clear just having kg in the input-field.

src/feedback/components/AmountPicker.vue
```diff
@@ -122,9 +95,30 @@ export default {
...
+$border-radius = 13px
+
+@import '~variables'
+
 .wrapper
   position relative
 
+.content
+  background $grey-2
+  border-radius 13px
```
nicksellen, line 107: Could use the nice variable defined a few lines above :)

src/locales/locale-en.json
```diff
@@ -106,7 +106,7 @@
     "UPCOMING_ACTIVITIES_IN_SERIES": "Upcoming activities in this series"
   },
   "ACTIVITY_FEEDBACK": {
-    "AMOUNT": "Amount of food you were able to save (in kg)",
+    "AMOUNT": "Amount of food you were able to save",
```
nicksellen, line 109: Given we are changing the text, thus trigger the translations to be redone, I wonder if it makes sense to make it non-food specific? What do you think @brnsolikyl?

Could be _Amount you were able to save_ ... well, even _save_ might not be the right term for some cases, now it could be more general... but one step at a time...

src/feedback/components/AmountPicker.vue
```diff
@@ -12,58 +12,45 @@
...
             <span class="text-caption">kg</span>
           </template>
         </QInput>
+        <div
+          v-if="!$q.platform.is.mobile"
+          class="col q-ml-md"
+        >
+          <AmountViewer :amount="value" />
+        </div>
       </div>
```
nicksellen, line 54: I wonder if the validation error message for that field could go somewhere in here?

It currently looks a little out of place and outside the grey box:

errormessage

Not very important though, as I think most people can input a number OK.

nicksellen (approved):
Hey, great! :pizza: :point_left: a tasty snack for you

I added a couple of minor comments with optional points, and maybe we can wait a few days to see if @tiltec or @brnsolikyl want to chip in where I mentioned them.

I approved though, so can be merged when it seems right!

src/feedback/components/AmountPicker.vue
```diff
@@ -12,58 +12,45 @@
...
-        />
-        <!-- don't use type="number" here because browsers might enforce different decimal setting
-        depending on browser locale-->
+      <div class="content-body row no-wrap items-center">
         <QInput
           v-model="valueToNumber"
-          size="4"
-          class="q-mr-md"
+          type="number"
+          inputmode="decimal"
```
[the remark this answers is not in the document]
larzon83: In our testing it was working correctly with different locales. The HTML handled it nicely. But maybe we missed something?

src/feedback/components/AmountPicker.vue
larzon83: ah, I think there were more selectors with this border-radius in our first attempt. Now it's only one appearance. I think we can ditch defined variable in this case :)

src/feedback/components/AmountPicker.vue
```diff
@@ -122,9 +95,30 @@ export default {
...
+    height 60px
+
+    &:after
+      position absolute
+      top 0
+      right 0
+      width 60px
+      height 100%
+      content ''
+      background linear-gradient(90deg, rgba(255, 255, 255, 0) 0%, $grey-2 75%)
```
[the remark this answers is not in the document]
larzon83: You see that this produces a fade-out effect if that images are wider then the box? Would you prefer to remove it completely or do you have an idea on how to achieve this with quasar helper classes?

src/feedback/components/AmountPicker.vue
larzon83: yup, you are right. We'll fix this.

larzon83:
> Hey, great! 🍕 👈 a tasty snack for you

Haha thx! 👅

src/feedback/components/AmountPicker.vue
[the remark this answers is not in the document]
larzon83: I was talking about the whole block, too :)
I don't see a way to achieve this with just quasar helper classes.

src/feedback/components/AmountPicker.vue
larzon83: It's better now. But not possible to have it inside the grey box as it's a general error-div in the parent FeedbackForm:
Bildschirmfoto 2021-05-16 um 22 59 14

But I noticed errors in the console:
Bildschirmfoto 2021-05-16 um 22 55 46

I think we need to bring back the Number conversion in the set() of `valueToNumber`:
```js
set (v) {
  let value = parseFloat(v, 10)
  value = isNaN(value) ? 0 : Math.max(0, value)
  this.$emit('input', value)
},
```

The `isNaN` check will then prevent the error-div from showing up entirely.

larzon83: Number conversion is back. No more errors in console and the validation error message will not appear at all.

src/feedback/components/AmountPicker.vue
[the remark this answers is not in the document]
larzon83: @tiltec do you have any feedback on this? Or maybe you know any edge cases to test out?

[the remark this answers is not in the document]
nicksellen: Is this the last point we're waiting for before merging?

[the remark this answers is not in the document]
larzon83: Yep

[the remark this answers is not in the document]
tiltec: There's most likely a difference between different browser/system locales (in contrast to just switching Karrot's locale). Also, rendering was quite different between desktop and mobile browsers.

But it's still reasonable to use it! Not sure about this case - it probably works either way? Maybe Quasar handles parsing these differences?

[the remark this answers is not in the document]
larzon83: @nicksellen you mentioned, you have/had a BrowserStack account. I could check there different browsers and system locales...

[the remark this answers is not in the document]
tiltec: I set up a SauceLabs open source account one some time ago, maybe that's helpful for you too? I sent you an invite via email (as found in rocket chat)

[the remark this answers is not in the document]
larzon83: thx. Will see what I get from it...

nicksellen:
I'm going to merge this one, maybe we can search for issues once it's deployed to dev. Can always fix later :)

PR #2360, opened 2021-05-08
Swap gitbook for mdbook

nicksellen:
## What does this PR do?

Gitbook [has changed](https://docs.gitbook.com/resources/gitbook-legacy/v2-differences), and we need to move away. https://github.com/rust-lang/mdBook seems a reasonable alternative, and I didn't need to make many changes.

One downside is it means anybody working on the docs needs to install mdbook too. It was available in community repo in arch though, so maybe available easily? Otherwise downloading from github releases is fairly easy.

This is also useful as gitbook installs a really old version of npm, which itself installs an old version of `graceful-fs` that doesn't work with node 12+, and that npm gets called when doing a cordova install... that's actually why I started doing this in the first place :)

## Checklist

- [ ] no unrelated changes


nicksellen:
I didn't check any checklist items as it's not like a code change... Also doesn't even get run until it's in master anyway. But promise to fix any breakages after its merged!

Review still appreciated though, to get some eyes on it.

nicksellen:
I'm gunna merge this, as master is currently broken, and I think this will fix it... comments still welcome though!

pogopaule:
Looks good to me + works on my machine

PR #2383, opened 2021-06-14
replace usages of group.members

tiltec:
This removes usages of `group.members`, to allow its removal from the backend serializer in https://github.com/yunity/karrot-backend/pull/1151


nicksellen:
Isn't removing the member list only for the public group info? We would still have the member list for the groups the user is part of right? Isn't that enough to maintain `myMemberIds` as it is now?

tiltec:
We just have the member list of the current group in the frontend, not all groups the user is member of.

nicksellen:
> We just have the member list of the current group in the frontend, not all groups the user is member of.

Ah I see, if we published events of the form `join|leave + group id + user id` (i.e. no details, just ids) on any join/leave action, and publish to all online members of the group, would that be sufficient for the frontend to do the right thing?

That would also fit more in a direction I'm kinda interested in -> broadcasting more minimal info (ids), and allowing the frontend to decide more if it needs to reload some full info...

tiltec:
Should work, ideally the current user gets their own _join_ and _leave_ events to simplify frontend logic. We should be careful to only send out events after the changes have been committed to the database, otherwise the frontend might reload an earlier version of the user list.

tiltec:
I introduced two new websocket event type, should work now. It does more work than necessary, but  this shouldn't be problematic.

I also fixed a slightly related bug about missing user info in application chat.

Will update changelog and fix tests later!

PR #2387, opened 2021-06-16
fix: remove iconpicker-app-extension

### LINKED ISSUE #2385: Improve bundling of icon picker
We are including the entire icon picker with all the icon sets in our main bundle :/ This massively increases the size, beyond what is actually needed.

Two problems:
1. all icons sets bundled but we only use `fontawesome-v5`
2. they're all bundled into the main app bundle, instead of one that loads of demand (it's a very infrequently used page)

Fixes for either or both are very welcome!

iconpickerbundle

https://dev.karrot.world/bundlesize.html

larzon83:
Closes #2385

## What does this PR do?

This PR removes `@quasar/quasar-app-extension-qiconpicker` and instead adds `@quasar/quasar-ui-qiconpicker"`.

The app-extension pushes all icons to the quasar boot config and therefor all icon-sets are included in the main app bundle thus massively increasing bundle size.

The ui-qiconpicker on the other hand can be used on a component level. So, a separate bundle is created and only loads when needed. See this screencast:

https://user-images.githubusercontent.com/23213583/122192156-acd3ad00-ce93-11eb-9438-fee4ad999c7a.mp4

Bundle size before and after:
size-1-before size-2-after

## Checklist

- [ ] no unrelated changes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,6 +27,7 @@ Please document your changes in this format:
 ### Fixed
 - place-header: correctly display linear-gradient in Safari @larzon83 [#2372]
 - Redirect again to last visited group @tiltec #2373
+- remove iconpicker-app-extension @larzon83 [#2387]
 
 ## [9.3.0] - 2021-06-03
 ### Added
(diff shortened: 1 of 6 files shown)


nicksellen (approved):
Great, much better, and nice and simple fix :star:

PR #2444, opened 2021-09-22
Place default page

### LINKED ISSUE #1703: Choice between "wall" or "activities/pickups" as the default page for a place
In cases where the wall is the most used feature for a specific place, it would be great to be able to choose while editing the place to either make the "Wall" or "Activities/Pickups" as the default page when coming to a place's page.

brnsolikyl:
Closes #1703 
Branch deployment https://place-default-page.dev.karrot.world/#
Still a work in progress and I mean to do only the UI. I'll need help with the rest.

## What does this PR do?

Allows choice between wall and activities as default place page. Also some other smaller improvements on the place edit page.

## Checklist

- [ ] no unrelated changes


nicksellen:
I implemented the backend for it. I called it `default_view`. And I have the frontend implementation locally ready here, so it actually works, but I'll wait to merge the backend first... so it doesn't break your work so far.

nicksellen:
I implemented the backend and merged that, so it's deployed to dev now (https://github.com/yunity/karrot-backend/pull/1197). I also just pushed a commit to your branch to actually read/write the value from the backend, so your UI is now working!

I would suggest removing the location toggle for this one, it has a few extra considerations about what to set the address value to, so I think makes sense in another PR. 

I didn't do a code-level review yet, but when you're ready I can look at that too.

From the UI perspective sometimes is bothering me about the central alignment of the activity / wall option, and it doesn't quite feel like it fits with the other form controls, but I can't quite imagine how to resolve that.

brnsolikyl:
Alright, I removed the location toggle and it seems good to go for the moment.

> From the UI perspective sometimes is bothering me about the central alignment of the activity / wall option, and it doesn't quite feel like it fits with the other form controls, but I can't quite imagine how to resolve that.

I agree, although for whatever subjective reason I think aligning to the left makes it worse. Maybe @larzon83 has an idea and some spare time to make it pretty? :)

larzon83:
> I agree, although for whatever subjective reason I think aligning to the left makes it worse. Maybe @larzon83 has an idea and some spare time to make it pretty? :)

May like this?
Bildschirmfoto 2021-11-25 um 22 06 22

brnsolikyl:
Oh looks consistent, much better! The only downside I see with it is that the user cannot see both options right away. But nonetheless, I'd definitely go for this design!

larzon83:
@brnsolikyl cool. You should be able to implement it yourself (look at the status select). Otherwise, I can help ;)

brnsolikyl:
Great! I've got a question of procedure then... there will be a lot of commits doing and undoing things on this PR. Is it better to start a fresh branch with only the necessary commits and a new PR? Or it doesn't really make a difference?

nicksellen:
We squash the commits when we merge. This means it ends up being just one commit in the end.

So, go wild!

nicksellen:
Looks good now! Merging :+1:

PR #2453, opened 2021-11-20
remove reference to rocket chat from PULL_REQUEST_TEMPLATE.md

amengsk:
## What does this PR do?

replace rocket chat by matrix room

## Links to related issues

https://community.karrot.world/t/moving-our-chat-where-to-find-us-the-team-and-the-community-besides-this-forum/787


nicksellen:
Thanks, looks good. I'll merge this. I did wonder if we should set up an redirect link, so if we change our team chat in the future again we only need to update one place. But that can be a seperate thing...

The link goes to the space, which I think shows a suitable view for them to join a specific room.

PR #2460, opened 2021-12-06
Add day of week filter for activities

### LINKED ISSUE #2310: Filter activities by day of the week
I forward the idea of one friend at Foodsharing Luxembourg: It would be nice to have a filter by day of the week (Monday, Tuesday, Wednesday, Thursday, Friday, Saturday, Sunday) to the activities page as they can only be active on a few days of the week.

danclee336:
Closes #2310 

## What does this PR do?

Add "day of week" filter for activities on the activities page.

A written test is not needed as we can test out the functionality manually. I first created
activities that cover all the days of a week as a whole group but not individually. Then I tried to apply all possible combinations of the filter to see if only activities of the selected days of the week appear.

## Checklist

- [x] no unrelated changes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -71,6 +71,7 @@ Please document your changes in this format:
 - Improve calculation of map bounds on group gallery @nicksellen [#2333]
 - main-layout: use scaled down background-image @larzon83 [#2371]
 - Use roboto font in latin-ext variant @nicksellen [#2345]
+- Add day of week filter for activities [#2310]
 
 ### Fixed
 - Place statistics feedback weight should be summed @nicksellen [#1137](https://github.com/karrot-dev/karrot-backend/pull/1137)
--- a/src/__snapshots__/storyshots.spec.js.snap
+++ b/src/__snapshots__/storyshots.spec.js.snap
@@ -1165,10 +1165,7 @@ exports[`Storyshots ActivityItem started 1`] = `
 exports[`Storyshots ActivityList Default 1`] = `
 <div>
   <!---->
-  <div class="row no-wrap items-center justify-end q-px-sm q-mt-sm">
-    <!---->
-    <!---->
-  </div>
+  <!---->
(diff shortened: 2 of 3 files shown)


nicksellen:
I just tried it out, and see a number of improvements, but not sure if you are ready to have it reviewed yet... are you still working on it?

danclee336:
@nicksellen Yes, I am ready to have it reviewed!

nicksellen:
> @nicksellen Yes, I am ready to have it reviewed!

Ok! Here's a screenshot when I run it locally:

dayofweekfilter

Two main points (before looking at the code):
- I think it takes up too much space at the top, could be in a select form component (see the day of week filter in the activity edit form for inspiration)
- it doesn't work! no activities are displayed, and the console is full of errors

I'll write a few comments directly on the code too...

> But I'm unsure what's the problem with the circleci deploy-branch check.

Don't worry about that check, it can't run properly when on a fork, and I should prevent it from doing so...

nicksellen:
After looking at the code, I think it will need quite some more work.

Currently you try to use a data variable for each day of the week (`this.mon`, etc.), it would be better to keep an array of all the day _numbers_ that you want included. You also need to declare any data variables you use in the `data()` initialization function.

The filter function then won't need so much repeated code.

Also, using the formatting functions is not a good idea for that purpose, it is unnecessary work every time the filter is run (which might be frequent), it also won't work if another language is selected.

The date field in each activity should be a proper javascript date, which has a `getDay()` function (see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/getDay).

And as I mentioned above that a select would be a better component to use, here is the code from activity series edit that renders that select:

https://github.com/karrot-dev/karrot-frontend/blob/1e42c53512e315f20f9785e0ffc3207182069f02/src/activities/components/ActivitySeriesEdit.vue#L174-L207

I'm aware this might seem a lot of things, and I don't have the capacity to teach you all of these things, so if it's too much for you right now and you can't find other sources of help, perhaps trying to solve a different problem might work better :)

PR #2525, opened 2022-04-20
Archived type filter

### LINKED ISSUE #2506: Activity type filter shows archived types
The activity type filter on the activities page let's you choose activity types that have been archived (e.g. `Party` type is archived):

archivedfilter

... one caveat is that it should still show archived activity types if there are activities for that type in view (as archiving a type does not delete the already scheduled activities).

nitishvijai:
Closes #2506

## What does this PR do?

This pull request should fix the issue where archived types should be shown in the filter if there are activities of that type in view, otherwise should not be in the filter.

Tested manually in the following cases: an archived activity type whose activities are in view, an archived activity type when no activities of that type exist, and when activities of an archived type are being deleted. However, looking to add a unit test once we get more direction on how to proceed from Nick. `yarn lint` and `yarn test` pass locally on our changes.

image

image

## Checklist

- [x] no unrelated changes

### DIFF
--- a/src/activities/components/ActivityList.vue
+++ b/src/activities/components/ActivityList.vue
@@ -334,6 +334,7 @@ export default {
       icsDialog: false,
       numDisplayed: NUM_ACTIVITIES_PER_LOAD,
       types: [],
+      activityListTypeFilter: this.filterActivityTypes.slice(),
     }
   },
   computed: {
@@ -355,12 +356,15 @@ export default {
     },
     typeOptions () {
       if (!this.filter) return []
+
+      this.checkArchivedTypes()
+
       return [
         {
           label: this.$t('ACTIVITYLIST.FILTER.ALL_TYPES'),
           value: 'all',
         },
-        ...this.filterActivityTypes.map(activityType => {
+        ...this.activityListTypeFilter.map(activityType => {
           return {
             label: activityType.translatedName,
             value: String(activityType.id),
@@ -425,6 +429,40 @@ export default {
     clearFilters () {
       this.slots = 'all'
       this.type = 'all'
+    },
+    checkArchivedTypes () {
+      const archivedTypes = {}
+
+      for (const activityType in this.activityListTypeFilter) {
+        // count the number of archived types and store in a hashmap, defaulting to zero
+        if (this.activityListTypeFilter[activityType].status === 'archived') {
+          archivedTypes[this.activityListTypeFilter[activityType].id] = 0
+        }
+      }
+
+      console.log(this.activities)
+
+      for (const num in this.activities) {
+        // count the number of activities for each archived type and update the hashmap
+        const activity = this.activities[num]
+        if (activity.activityType.id in archivedTypes) {
+          const id = activity.activityType.id
+          archivedTypes[id] += 1
+        }
+      }
+
+      for (const archivedTypeId in archivedTypes) {
+        if (archivedTypes[archivedTypeId] === 0) {
+          // For any archived types with zero activities in view, remove the type from the filter
+          for (const [index, activityType] of this.activityListTypeFilter.entries()) {
+            if (parseInt(activityType.id) === parseInt(archivedTypeId)) {
+              // delete archived type from the filter
+              this.activityListTypeFilter.splice(index, 1)
+              break
+            }
+          }
+        }
+      }
     },
     copyLink () {
       return copyToClipboard(this.icsUrl).then(() => {
--- a/src/activities/components/ActivityList.vue
+++ b/src/activities/components/ActivityList.vue
@@ -338,6 +338,46 @@ export default {
     }
   },
   computed: {
+    activityTypes () {
+      const archivedTypeList = []
+      const archivedTypes = {}
+
+      for (const type of this.activityListTypeFilter) {
+        // populate a list of all activity types locally to avoid side effects
+        archivedTypeList.push(type)
+      }
+
+      for (const activityType of archivedTypeList) {
+        // count the number of archived types and store in a hashmap, defaulting to zero
+        if (activityType.status === 'archived') {
+          archivedTypes[activityType.id] = false
+        }
+      }
+
+      for (const activity of this.activities) {
+        // check to see if each archived type has any activities and update the hashmap
+        if (activity.activityType.id in archivedTypes) {
+          const id = activity.activityType.id
+          archivedTypes[id] = true
+        }
+      }
+
+      for (const archivedTypeId in archivedTypes) {
+        console.log(archivedTypeId)
+        if (archivedTypes[archivedTypeId] === false) {
+          // For any archived types with zero activities in view, remove the type from the filter
+          for (const [index, activityType] of archivedTypeList.entries()) {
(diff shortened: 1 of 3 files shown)


src/activities/components/ActivityList.vue
```diff
@@ -355,12 +356,15 @@ export default {
     },
     typeOptions () {
       if (!this.filter) return []
+
+      this.checkArchivedTypes()
```
nicksellen, line 360: Instead of this kind of imperative style, in vue it is better to use a computed property for this. If you haven't learnt about computed properties yet, maybe https://v2.vuejs.org/v2/guide/computed.html will help.

So, I suggest to create a computed property called `activityTypes` which will run your logic and return the activity types for display in the filter.

src/activities/components/ActivityList.vue
```diff
@@ -426,6 +430,40 @@ export default {
...
+      for (const activityType in this.activityListTypeFilter) {
+        // count the number of archived types and store in a hashmap, defaulting to zero
+        if (this.activityListTypeFilter[activityType].status === 'archived') {
+          archivedTypes[this.activityListTypeFilter[activityType].id] = 0
+        }
+      }
+
+      console.log(this.activities)
+
+      for (const num in this.activities) {
```
nicksellen, line 445: You can use a for/of loop here, e.g.:

```js
for (const activity of this.activities) {
```

and saves the line `const activity = this.activities[num]` below.

src/activities/components/ActivityList.vue
```diff
@@ -426,6 +430,40 @@ export default {
...
+          const id = activity.activityType.id
+          archivedTypes[id] += 1
+        }
+      }
+
+      for (const archivedTypeId in archivedTypes) {
+        if (archivedTypes[archivedTypeId] === 0) {
+          // For any archived types with zero activities in view, remove the type from the filter
+          for (const [index, activityType] of this.activityListTypeFilter.entries()) {
+            if (parseInt(activityType.id) === parseInt(archivedTypeId)) {
```
nicksellen, line 458: Are these not already `int`s?

src/activities/components/ActivityList.vue
```diff
@@ -426,6 +430,40 @@ export default {
...
+      const archivedTypes = {}
+
+      for (const activityType in this.activityListTypeFilter) {
+        // count the number of archived types and store in a hashmap, defaulting to zero
+        if (this.activityListTypeFilter[activityType].status === 'archived') {
+          archivedTypes[this.activityListTypeFilter[activityType].id] = 0
+        }
+      }
+
+      console.log(this.activities)
```
nicksellen, line 443: Can remove this line.

src/activities/components/ActivityList.vue
```diff
@@ -426,6 +430,40 @@ export default {
...
+
+      for (const num in this.activities) {
+        // count the number of activities for each archived type and update the hashmap
+        const activity = this.activities[num]
+        if (activity.activityType.id in archivedTypes) {
+          const id = activity.activityType.id
+          archivedTypes[id] += 1
```
nicksellen, line 450: The actual count doesn't really matter here, the only important thing is if there are none, or some.

nicksellen (changes requested):
Thanks! :star: 

A left a few comments, I didn't try it out at this moment, but assume it works I'd approve with a few changes.

I think it'd be possible to write it in a more compact way, but I can accept it :)

src/activities/components/ActivityList.vue
nitishvijai: I believe `archivedTypeId` is a string by default (this is just the ID of the archived type), so I convert both values to integers to do a strict comparison and as a safeguard.

Do you think we should convert it to an integer when populating the dictionary, or is this fine the way it is?

src/activities/components/ActivityList.vue
nitishvijai: I'm a little confused by this. I moved the logic into the methods list since we originally had it all in the `typeOptions()` function, but 'yarn lint' threw an error about possible side effects. If we created `activityTypes` as a computed property, what would we do with the return statement on lines 362-374 (where we map each `activityType` in `activityListTypeFilter` to a filter value)?

nicksellen: Yes, in a computed function you need to make sure you aren't mutating values, you always return a fresh array with the data, based on the source values.

You would use `this.activityTypes` on line 367, which would be a list containing the right activity types (non-archived + ones in use), calculating in the computed property.

src/activities/components/ActivityList.vue
nicksellen: Ah yeah, it's because you're storing things in a map, and the keys are always strings. So you can't put them in as int's. It would require a different reworking, which I can't explain right now, so it's fine as it is for this purpose :)

src/activities/components/ActivityList.vue
nitishvijai: Ah I see. I just tried this (creating a computed property, `activityTypes`, with the logic to deal with archived types) but I get the issue again about unintended side effects when calling lint. 

I'm not sure how we can deal with this besides using the `checkArchivedTypes` method.

nitishvijai: I think I was able to figure it out by the way! I created a new local list in the `activityTypes` computed property and filled in everything from `activityListTypeFilter` so that I could avoid mutating and causing any potential side effects to the data or other fields. Pushing a new commit to address this.

src/activities/components/ActivityList.vue
```diff
@@ -334,9 +334,50 @@ export default {
...
     }
   },
   computed: {
+    activityTypes () {
+      const archivedTypeList = []
+      const archivedTypes = {}
+
+      for (const type of this.activityListTypeFilter) {
+        // populate a list of all activity types locally to avoid side effects
+        archivedTypeList.push(type)
```
nicksellen, line 347: Seems you're just cloning a list here, which you can do in a simpler/cleaner way (off the top of my head):

```js
const newList = [...oldList]
// or
const newList = oldList.slice()
```

src/activities/components/ActivityList.vue
```diff
@@ -334,9 +334,50 @@ export default {
...
+            if (parseInt(activityType.id) === parseInt(archivedTypeId)) {
+              // delete archived type from the filter
+              archivedTypeList.splice(index, 1)
+              break
+            }
+          }
+        }
+      }
+
+      return archivedTypeList
```
nicksellen, line 379: This variable name seems misleading, it's not a list of archived types... maybe just call it `activityTypes`?

src/activities/components/ActivityList.vue
```diff
@@ -334,9 +334,50 @@ export default {
       icsDialog: false,
       numDisplayed: NUM_ACTIVITIES_PER_LOAD,
       types: [],
+      activityListTypeFilter: this.filterActivityTypes.slice(),
```
nicksellen, line 337: I think this is not needed here, and I think it'll also prevent `activityListTypeFilter` being reactive to changes to the prop value.

nicksellen (approved):
Ok, looks good! :rocket:

PR #2536, opened 2022-06-06
Improve testing

nicksellen:
This PR is for https://community.karrot.world/t/nlnet-responsiblities/892 __Add frontend integration tests and aim for 75% test coverage__ task.

This is a change of testing approach, loosely based on Kent C. Dodds blog post [Write tests. Not too many. Mostly integration.](https://kentcdodds.com/blog/write-tests).

The premise is that the tooling is good enough to write integration tests that run reliably, and are easy enough to write. The older paradigm of a bazillion unit tests was partly based on it being difficult to write integration tests.

So, the new approach is:
- use https://testing-library.com which encourages higher level testing from the user perspective
- two types of tests:
  - page-level tests (e.g. testing stuff in `src/*/pages/*.vue`) that load the page component but not the router
  - `App` tests, that load the root `App` component and the router
- use the new `mockBackend` which mocks at the axios-layer, for all tests that need it


tiltec:
Regarding integration tests, I wonder if we should add the Cypress test framework? It seems to be recommended by Quasar: https://github.com/karrot-dev/karrot-frontend/pull/2536

nicksellen:
My current sense is to try and create integration tests that are not end to end tests.

So, this sense of them from [Kent C Dodds](https://kentcdodds.com/blog/the-testing-trophy-and-testing-classifications):


The e2e testing tools are quite heavy. So, one idea, which I didn't yet explore further is something along the lines of:
- use our existing jest-based setup
- load the entire app, complete with a real store
- mock at the API/REST-layer only

The sense of the _goals_ I have here are:
- be able to test how the an assemble of components behaves in the presence of the store
- be the main default way to write a test for a feature
- hit the sweet spot between a tool that actually runs a browser and clicks around (slow, flakey) and complex components that have a mountain of barely functioning mocks surrounding it (complex, fragile, not testing a lot)

It might be that:
- loading the entire app this way is too hard, but maybe page-level components would still be useful _integration_ tests
- it reveals how entangled everything is, so it can be best-effort, and use the knowledge to inform the data layer explorations
- we need a quite complex mock API (I remember discussing/debating this before, at some point it feels like implementing a whole new backend :) )

However, a bunch of things push integration tests more into e2e tooling. [quasar-testing](https://github.com/quasarframework/quasar-testing/tree/dev) seems still being upgraded to v2, and lists unit testing (with jest) and e2e testing (with cypress), so the type of tests I'm talking about may not be very well supported by the tooling.

tiltec (approved):

test/mockBackend/mockAxios.js
```diff
@@ -0,0 +1,160 @@
...
+    }
+    const handlerResponse = handler({ ...config, data, pathParams })
+    if (!Array.isArray(handlerResponse)) throw new Error('mock handler must return an array')
+    const [statusCode, body] = handlerResponse
+    if (typeof statusCode !== 'number') throw new Error('mock handler array must have numeric status code as first arg')
+    return [statusCode, body && underscorizeKeys(body)]
+  })
+}
+
+export function cursorPaginated (path, getEntries, options = {}, makeResults = null) {
```
nicksellen, line 86: `makeResults` could go into the `options`

test/mockBackend/issues.js
```diff
@@ -0,0 +1,69 @@
...
+  }
+}
+
+let nextVotingId = 1
+export function generateVoting (params = {}) {
+  return {
+    id: nextVotingId++,
+    createdAt: subDays(new Date(), 6),
+    expiresAt: addDays(new Date(), 7),
+    acceptedOption: 74,
```
nicksellen, line 54: This would be an invalid option, maybe `null` best for now, and we can implement the more sophisticated functionality if needed...

test/mockBackend/mockAxios.js
tiltec: since `options` is passed as-is into mockAxios `get`, it feels a bit off to put other stuff inside...

nicksellen: the `get` is only our `get` function, both `get` and `cursorPaginated` are part of our "mock axios" "library", so I think having a common object for the object fits, it's _our_ options object..

the only other use of the options is `{ requireAuth: false }`, it doesn't get passed into the axios mock adapter lib.

tiltec: makes sense now! thanks for the explanation, I changed the code

PR #2539, opened 2022-06-11
[CI] Update locales

tiltec:


src/locales/locale-ko_KR.json
```diff
@@ -40,11 +40,11 @@
         "TITLE": "공유하기"
       },
       "OPENSOURCE": {
-        "DESCRIPTION": "모든 건 투명하며 변경할 수 있습니다. 각 그룹이 플랫폼의 인스턴스를 호스팅할 수 있는 능력이 있다면 그러시기를 권장합니다. 물론 karot.com의 호스트 버전에 대한 액세스도 여전히 제공할 것입니다.",
+        "DESCRIPTION": "모든 건 투명하며 수정할 수 있습니다. 저희는 각 그룹이 플랫폼의 인스턴스를 호스팅할 수 있는 능력이 있다면 그러시기를 권장합니다. 물론 karrot.com의 호스트 버전에 대한 액세스도 여전히 제공할 것입니다.",
```
nicksellen, line 43: Just noticing in this translation it refers to karrot.com which is not our domain. If you visit it you end up at https://www.kabbage.com/ which says it's something to do with american express. I suspect it's just a mistake rather than anything malicious, but my korean is not very good...

PR #2544, opened 2022-06-18
Fix upgrade bugs

nicksellen:
Branch deployment: https://fix-upgrade-bugs.dev.karrot.world

Fixes a few things that were broken after the upgrade:

## Map

We switched from vue2leaflet to @vue-leaflet/vue-leaflet lib, and it had some odd behaviour. I couldn't get to the bottom of it simply, and the last release of the library is a year ago (there are more recent commits though).

I decided to just remove that layer and use leaflet directly and implement the vue bindings myself, which wasn't as bad as I might have imagined.

I implement new `KMap` and `KMarker` components. `StandardMap` now uses those instead of the LMap/etc components. The rest of the map stuff is pretty much unchanged.

## Small fixes

- sizes of message actions (thread/reaction buttons) has been restored to their normal size
- anchor links (e.g. settings#notifications) didn't work as vue-router changed from `selector` prop to `el`


tiltec (approved):
Looks good! Hopefully this should also resolve our map issues that we had before :) Let's merge it and see if it deploys nicely.

PR #2555, opened 2022-06-27
Add ICS token auth support

tiltec:
First implementation that uses token auth for ICS export URLs and supports token refresh.

image

At the moment, the "Download" URL also has the token appended, which is usually unnecessary because it's usually downloaded with the browser session. It however allows Karrot app users to download the ICS file, because the download is usually deferred to a browser that isn't logged in.

Requires backend PR https://github.com/karrot-dev/karrot-backend/pull/1229

Closes https://github.com/karrot-dev/karrot-frontend/issues/2541

### DIFF
--- a/src/locales/locale-en.json
+++ b/src/locales/locale-en.json
@@ -104,7 +104,8 @@
       "SUBSCRIBE_TITLE": "Subscribe",
       "SUBSCRIBE_TEXT": "Use this URL to subscribe as an ICS calendar:",
       "MORE_INFOS": "More info about ICS",
-      "REFRESH_TOKEN": "Everyone who knows your secret URL can read your activities. To avoid this, please don't share the URL with other people. If this happenend to you, click here to create a fresh secret URL and invalidate previous URLs:"
+      "TOKEN_NOTICE": "Anyone who has the secret link can access. No sign-in required.",
+      "RESET_TOKEN": "Reset secret link"
     }
   },
   "ACTIVITYMANAGE": {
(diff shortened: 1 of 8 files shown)


src/locales/locale-en.json
```diff
@@ -103,7 +103,8 @@
       "DOWNLOAD_BUTTON_LABEL": "Download ICS",
       "SUBSCRIBE_TITLE": "Subscribe",
       "SUBSCRIBE_TEXT": "Use this URL to subscribe as an ICS calendar:",
-      "MORE_INFOS": "More info about ICS"
+      "MORE_INFOS": "More info about ICS",
+      "REFRESH_TOKEN": "Everyone who knows your secret URL can read your activities. To avoid this, please don't share the URL with other people. If this happenend to you, click here to create a fresh secret URL and invalidate previous URLs:"
```
nicksellen, line 107: Minor language point, instead of _can **read** your activities_ would use _see_, _view_, or _access_ instead.

It also seems a bit wordy, so maybe a total rewrite (borrowing a bit from [googles dialog](https://www.varonis.com/blog/the-dangers-of-shared-links) for shared links):

_Anyone who has the secret link can access. No sign-in required._

Then maybe have the button as _Reset Secret Link_

Just an idea...

nicksellen (approved):
Woo, looks great! I think this makes it very useful. I'll approve, and you can decide which suggestions to incorporate or not.

amengsk (approved):
looks great to me as well
thank you @tiltec

PR #2560, opened 2022-07-08
Use vue-query for offers

nicksellen:
Closes ___ (e.g. #230)

Initial exploration for using vue-query. Just using it for the offers part at the moment as that is simpler than other parts.

Notable points:
- `<ProfilePicture>` now accepts a user object _or_ just a user id (so we can use non-enriched offer objects)
- it handled the cursor pagination nicely, and knows how to refresh it too (get the first page, and go from there)
- it automatically handles refreshing after returning to page after being away for a bit
- I went full-in for composition api and `<script setup>` - see offers/pages/*.vue files...
- it's quite easy to bridge data to existing vuex data, see my `useCurrentGroupId()` thing

TODO:
- the offers form accepts a `status` object, but I only implemented the `validationErrors` bit of it... need to decide whether to map the vue-query status to our existing form, or switch the code to use vue-query-style status...
- think a bit about the query client refetching/invalidating ...
- websocket updates
- probably more stuff...

(briefly describe what this PR is about)

## Checklist

- [ ] no unrelated changes


src/App.vue
```diff
@@ -15,10 +15,18 @@
...
+import { useClearDataOnLogout } from '@/utils/composables'
+import { useOffersUpdater } from '@/offers/queries'
+
 export default {
   components: {
     LoadingProgress,
   },
+  setup () {
+    // Global kind of things can be registered here
+    useOffersUpdater()
```
tiltec, line 27: Essentially, we might always be watching if any websocket update invalidates any query cache... that's a lot of listening, but probably just fine.

src/utils/queryHelpers.js
```diff
@@ -0,0 +1,29 @@
+import { computed, unref } from 'vue'
+import { isNetworkError, isServerError, isValidationError } from '@/utils/datastore/helpers'
+
+export function extractCursor (url) {
+  if (!url) return null
+  return new URL(url, url.startsWith('http') ? null : 'https://karrot.world').searchParams.get('cursor')
```
tiltec, line 6: That domain probably shouldn't be hardcoded here...

Not sure I understand the reason for this, is it related to local dev?

src/offers/mutations.js
```diff
@@ -0,0 +1,106 @@
+import { useMutation, useQueryClient } from 'vue-query'
+import api from '@/offers/api/offers'
+import { unref } from 'vue'
+import { queryKeyOfferDetail, queryKeyOfferList } from '@/offers/queries'
+import { useRoute, useRouter } from 'vue-router'
+import { withStatus } from '@/utils/queryHelpers'
+import { useCurrentGroupIdRef } from '@/group/queries'
+
+// Store the ids we are currently mutating, so we can better decide when to refresh from websocket
+const mutatingOfferIds = new Set()
```
tiltec, line 10: Not sure I fully understand - I guess it shouldn't invalidate the offer while the user is editing it? Or is it about some data race when clicking the "save"/"create" button?

tiltec (commented):
Nice, this is a serious change in direction for frontend data handling!

I wonder where to draw the line between local state (to be handled with vue-query) and global state, e.g. currentOffer seems to be somehow global as it also affects the topbar. Is it reasonable to keep this kind of state in vuex (maybe later pinia) instead of the topbar/breadcrumb component having to know about all kinds of queries across Karrot?

src/offers/components/OfferDetailHeader.vue
```diff
@@ -28,19 +28,27 @@
...
+    return {
+      offer,
+      currentUserId: useCurrentUserIdRef(),
+    }
+  },
   computed: {
-    ...mapGetters({
-      offer: 'currentOffer/value',
-    }),
+    canEdit () {
```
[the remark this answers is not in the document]
nicksellen: Yes, if it's being repeated a lot then can extract it into a function/composable somewhere... permissions modules perhaps..

src/offers/queries.js
```diff
@@ -0,0 +1,189 @@
...
+ * Returns a query object with data also available "offer" key
+ */
+export function useOfferQuery ({
+  id,
+}) {
+  const query = useQuery(
+    ['offer', id],
+    () => api.get(unref(id)),
+    {
+      // TODO: does this handle our currentOffer/clear case?
```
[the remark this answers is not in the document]
nicksellen: Yeah, it doesn't do exactly the same, but the case for the currentOffer it to make sure it keeps the current offer up to date, so I _think_ with the _enabled_ bit (and the fact that the `id` is a ref) means that it should always have the correct one....

I'm not sure what data value is provided when `enabled` resolves to `false` though... could check.

src/App.vue
nicksellen: I think 1 per query-type, e.g. 1 for offers (well, 2 as there is `offers:offer` and `offers:offer_deleted` to listen on, so not too many...

I thought it makes sense to move the logic to do on websocket update in the files where the queries take place, so they are together, but alternatively could centralized it (like we do now with the vuex commits).

The other idea I had is to only ever use the `id` coming from the websocket, so we can transition to a lighter "procotol" for websocket updates in the future, just based on ids.

src/utils/queryHelpers.js
nicksellen: This function is to take something like: `/api/offers/?status=active&group=12&cursor=abc123`

and turn it into `abc123`

So, just extracting the query parameter from a path. The domain there is just because the URL constructor requires a domain.

I agree it looks confusing though, an improvement would be to just initialize a `URLSearchParams` object directly (and use substring on the `?` character), I'll do that.

src/offers/mutations.js
nicksellen: It's to avoid extra unneeded queries when receiving websocket updates. Without this logic the websocket update will invalidate the offers query and reload the data, and then the response to the mutation request will also do that, and two requests to the backend will go out. I think some situations like that would get debounced, but in this case it would make two separate requests.

Another option is to rely on the websocket update alone, but I tend to not like to rely on that approach, although I think currently we do in places and not others (depending on who wrote the code :), so maybe we should actually decide on a policy there).

I also looked at just updating the specific offer that is saved/edited/archived in the query (using `setQueryData`), but for the query it's not always obvious how a given edit might impact all the queries (e.g. when you archive it, it needs to be added to the status=archived query and removed from the status=active one), so I think just invalidating the lot is better.

src/offers/queries.js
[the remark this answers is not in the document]
nicksellen: I added a test for `useOfferQuery` now, it's `undefined` when enabled is false

nicksellen:
> I wonder where to draw the line between local state (to be handled with vue-query) and global state, e.g. currentOffer seems to be somehow global as it also affects the topbar.

a composable (`useBlah`) could obtain it's data from vue-query data, or vuex/pinia (or other places), so that part is kind of abstracted behind a "composable-interface" - and in that case, it's using the global data from vuex, and using it to make a vue-query query... could also get the offer id from the route params, or have them separated with something to get the current offer id, and then pass it into a plain `useOfferQuery` in the component... 

my sense is that a lot of the global data (from server) could be put into vue-query...

maybe there is still some need for global local state (i.e. not from serve), if it's only a little bit maybe just standalone `reactive()` objects are fine, with functions to update it... 

> Is it reasonable to keep this kind of state in vuex (maybe later pinia) instead of the topbar/breadcrumb component having to know about all kinds of queries across Karrot?

at the moment _something_ has to know how to get all the breadcrumb values from somewhere, currently in the breadcrumb datastore... another idea could be that page-level components set the breadcrumbs, so there stops being a single place that has to pull together all the sources, I think that might work quite nicely.

PR #2567, opened 2022-07-23
Migrate to vue-query

nicksellen:
Branch deployment URL: https://add-more-vue-query.dev.karrot.world

## What does this PR do?

This switches our frontend data layer from vuex to vue-query.

There were a few motivations for this:
- our vuex data layer had got very complex and abstract to work with
- pinia has replaced vuex as the recommended store (we're not switching to pinia)
- vuex has a serious bug in that getters are not being cached https://github.com/vuejs/vuex/issues/2102 - this caused enormous slowdown and made karrot unusable, there is no fix available
- our pattern of enriching data caused a lot of re-renders

One of the complexities of the existing data layer is passing values down through many layers of components. We did originally have a philosophy of _props in, events out_ for components, so they would be usable in isolation, but this pushed a lot of abstract complexity to the data layer. The data that the components needed would have to be managed far away from those components in the stores.

So, a new approach!
- no enriching of data in stores, the data for each object will just be what the server returned
- instead of enriching the base objects, we have helpers to calculate those kind of values (e.g. `isCurrentUser`)
- data is loaded into the components where it is needed via composable functions (aka. hooks)
- stores are replaced in some cases with lightweight _services_ (which are singleton/cached objects), in other cases queries can be used directly in components

This also has a big impact on testing strategy, that is dealt with over here --> https://github.com/karrot-dev/karrot-frontend/pull/2536

### TODO

This is a slightly daunting list of the modules (`src/*/datastore/*.js`) to convert over.

- [x] activities / activities
- [x] activities / activityPlugin
- [x] activities / activitySeries
- [x] activities / activityTypes
- [x] activities / activityTypeStylesheetPlugin
- [x] agreements / agreements - CAN BE REMOVED
- [x] alerts / banners
- [x] alerts / toasts
- [x] applications / applications
- [x] applications / applicationsPlugin
- [x] authuser / auth
- [x] authuser / deleteAccount
- [x] authuser / verifymail
- [x] base / base
- [x] base / dependentState
- [x] base / geo
- [x] base / i18n
- [x] base / i18nPlugin
- [x] base / persistedState
- [x] base / routeError
- [x] base / route
- [x] base / routeMeta - REMOVED
- [x] base / routerPlugin
- [x] communityFeed / communityFeed
- [x] feedback / feedback
- [x] group / currentGroup
- [x] group / timezones
- [x] groupInfo / groups
- [x] history / history
- [x] invitations / invitations REMOVED
- [x] issues / issues
- [x] issues / issuesPlugin
- [x] messages / conversations
- [x] messages / currentThread
- [x] messages / detail
- [x] messages / latestMessages
- [x] notifications / notifications
- [x] offers / currentOffer
- [x] places / places
- [x] places / plugin
- [x] sidenav / sidenavBoxes
- [x] status / status
- [x] subscriptions / fcm
- [x] topbar / breadcrumbs
- [x] topbar / loadingprogress
- [x] topbar / loadingProgressReporter
- [x] topbar / search
- [x] unsubscribe / unsubscribe
- [x] users / users
- [x] utils / about
- [x] utils / connectivity
- [x] utils / helpers
- [x] utils / presence
- [x] utils / pwa
- [x] utils / refresh
- [x] utils / refreshPlugin

## Checklist

- [ ] no unrelated changes


src/activities/pages/GroupActivities.vue
```diff
@@ -200,30 +208,45 @@ export default {
       return new Date(Math.floor(new Date().getTime() / roundTo) * roundTo)
     }
 
+    const queryClient = useQueryClient()
+
+    /*
+      TODO: something funny happening where it sort of freezes when there is query cache...
```
tiltec, line 214: Would be interesting to figure out if this is happens during rendering or somewhere in the data management layer...

nicksellen: Yeah, indeed. My current iteration of the code is to have it not do any caching for activities and only use backend filters. Which is fast locally, and waiting until that backend one is merged to try it out with more data...

I did try making it only keep the first page cached, but didn't solve that freezing...

nicksellen:
This PR is a bit crazy, sorry! It's a bit of a creative exploration. It's a bit like when I rearrange a room, just start pulling all the furniture out, moving things around, there is a crazy chaos for a while, then slowly the new form emerges...

My progress is quite good though. If it was needed to make a more incremental PR, I could extract some parts out to another PR, but I'll carry on with this way for now.

PR #2575, opened 2022-08-30
Add more tests!

tiltec:
Follow-up to #2536 

Goal is to get test coverage above 75% by having some more good unit and integration tests.


src/utils/composables.js
```diff
@@ -67,7 +67,7 @@ export function useIntegerRouteParam (name) {
   const route = useRoute()
   return computed({
     get () {
-      return route.params[name] && parseInt(route.params[name], 10)
+      return route?.params[name] && parseInt(route.params[name], 10)
```
nicksellen, line 70: I would prefer to always have a router, which might be a fake one (from `import '>/mockRouter'`) or a real one... but I think this might conceal bugs rather than reveal them. Hard to notice if we accidentally don't have a router when we are expecting to. WDYT?

test/mockBackend/activitySeries.js
```diff
@@ -0,0 +1,27 @@
+import { db } from '>/mockBackend/index'
+
+let nextId = 1
+export function generateActivitySeries (params = {}) {
+  if (!params.place) throw new Error('must provide place')
+  if (!params.activityType) {
+    // Auto-choose one based on place -> group
+    const place = db.orm.places.get({ id: params.place })
+    params.activityType = db.orm.activityTypes.get({ group: place.group }).id
```
nicksellen, line 9: This will throw an error if there is not _exactly_ one activity type, I would suggest we create a `sample` method in the orm thing, so we can do:

```js
params.activityType = db.orm.activityTypes.sample({ group: place.group }).id
```

it would randomly select one, and if none, throw an error (as it would expect that there is at least one).

src/storyshots.spec.js
```diff
@@ -1,74 +1,66 @@
...
-    setTimeout(async () => {
-      if (snapshotFilename) {
-        expect(tree).toMatchSpecificSnapshot(snapshotFilename)
-      }
-
-      done()
-    }, waitTime)
+    const tree = render(story.render(), withDefaults())
+    await flushPromises()
+    expect(tree.html()).toMatchSnapshot()
```
nicksellen, line 64: That's really nice and succinct now! :star:

src/utils/composables.js
tiltec: Agree, don't really like this solution either. But as far as I understood, `vue-router-mock` is somehow based on vue-test-utils, so I wonder how it works in storybook. `>/routerMocks` requires jest, so that wouldn't work.

test/mockBackend/activitySeries.js
tiltec: good idea!

src/utils/composables.js
tiltec: I undid this change, accepting that storybook won't work (for now).

test/mockBackend/activitySeries.js
tiltec: I would leave this for another time though...

PR #2577, opened 2022-09-10
Add UI for editing place types

tiltec:
Branch deployment URL: https://add-place-type-edit.dev.karrot.world

A nice UI for editing place types, very similar to the one for activity types.

Place type can be selected in place edit form. Custom icons will be shown in the place list and on the group map.

This lays the groundwork for refining display of places.

## Edit place types
image

image

## Select type for place
image

## place type markers on map

image

## notes

- custom icons might clash with pre-defined icons for users and groups
- place statuses are untouched, they still read "cooperating" etc...

Based on backend PR https://github.com/karrot-dev/karrot-backend/pull/1159

Related to https://github.com/karrot-dev/karrot-frontend/issues/2244


nicksellen:
Looks great!

> favorite places are not at top anymore

I think this is problematic, as groups with a lot of places likely use this a lot to keep it manageable.

However, I think with the new "explore places" UI it probably resolves this... I imagine for that one the favourites would stay on the sidebar?

> might be nice to let users specify the place type order in sidenav

I think alphabetical ordering is OK here, as groups already use that convention, currently they are not ordered that way.

**Suggestion: always order alphabetically by name** (translated name)

> it's a bit weird with singular names as headers in sidenav

It didn't strike me as odd before I read your comment, and even after, seems OK to my eye/brain...

> custom icons might clash with pre-defined icons for users and groups

yes, so be I think.... users can handle that. if you thought it more a problem, we could filter those out, as the icon picker let's us choose exactly which ones to show...

> sidenav feels a bit full now, maybe it's time for the map to go away?

I suspect the place UI re-jig will resolve that? although I agree in general, sidenav gets more and more stuff in it. but I think solving it is maybe out of scope of this, I would be in favour of going to more customizable level (on group and user levels).

> place statuses are untouched, they still read "cooperating" etc...

yes, indeed. it's a bit odd for non-foodsaving groups at the moment to see "co-operating"... I suspect we could just relabel that one to "active" (as I think the backend value is?), and it would be ok? ... and the custom place statuses is on the cards too... 

either way, out of scope of this one, but nice to remember it...

In addition I thought it would be nice to use the green colour when editing place types, as the green is a kind of "normal" state...

Screenshot_2022-09-12_12-42-33

I noticed it starts in "default" place type I think, with a circle icon, if I then select a place type, I can't get back to "default" or "no place type".... but I can't save without one set anyway... although the validation doesn't validate that (get server error). Minimal change is probably just adding it to the validation.

Screenshot_2022-09-12_12-43-13

I also wonder if we should _officially_ acknowledge the use of places for non-geographically purposes, and include a translatable name for: "Discussion" or "Working Group"? (although requires the backend change to also have those names I think).

Can throw some of these questions into the weekly meeting tomorrow (with or without attendance) if you'd like any decisions including the others :)

nicksellen:
Oh, and the icon for "place type" in the form is an eye, presumably a copy and paste error? I actually started removing some of these icons on forms, I'm not convinced they help, and clutter up the form and layout on small screens/mobile.

src/group/pages/EditPlaceTypes.spec.js
```diff
@@ -0,0 +1,89 @@
...
+  it('adds new place type', async () => {
+    const { click, type } = userEvent.setup()
+
+    const { queryByRole, findByTitle, findByRole, findByText } = render(EditPlaceTypes, withDefaults())
+    await flushPromises()
+
+    await click(await findByTitle('Add new place type'))
+    const textbox = await findByRole('combobox', { name: 'Name' })
+    // somehow we need to type the name before we can choose it
+    // seems like a problem between jsdom/testing-library and Quasar
```
nicksellen, line 59: My guess is that because the select is closed at first, you need to click the `findByRole('combobox', { name: 'Name' })` to open it... then can click the Distribution Place within it?

src/locales/locale-en.json
```diff
@@ -684,6 +685,32 @@
...
+    "ADD": "Add new place type",
+    "NAME": "Name",
+    "STATUS": "Status",
+    "STANDARD_NAME_HEADING": "Standard names",
+    "STANDARD_NAME_HINT": "Name will be available in other languages",
+    "CUSTOM_NAME_HINT": "Name will be used exactly as you write it and not translated",
+    "CUSTOM_NAME_USE": "Use custom name {name}",
+    "CUSTOM_NAME_PROMPT": "Type a custom name, or choose a standard name",
+    "MANAGE_TYPES": "Manage types...",
+    "SHOW_ARCHIVED": "Show archived",
```
nicksellen, line 710: Many of these translations are duplicate from activity types, I have been tending to re-use things in this case, even if the i18n keypath is a bit nonsensical... and then imagining if we can migrate keypaths properly one day (did we manage this before?)

src/places/components/PlaceEdit.vue
```diff
@@ -25,6 +25,55 @@
...
+            v-model="edit.placeType"
+            :options="placeTypeOptions.filter(({ status, value }) => status === 'active' || edit.placeType === value)"
+            map-options
+            emit-value
+            :label="$t('STOREEDIT.PLACE_TYPE')"
+            outlined
+            class="q-mb-lg"
+          >
+            <template #before>
+              <QIcon name="fas fa-eye" />
```
nicksellen, line 38: This is the copy-and-paste eye I guess? ... and my suggestion is actually to remove all the #before icons....

src/places/components/PlaceEdit.vue
```diff
@@ -265,8 +318,26 @@ export default {
...
   setup () {
+    const { groupId } = useCurrentGroupService()
+
+    const { getPlaceTypesByGroup, getPlaceTypeById } = usePlaceTypeService()
+
+    const { getTranslatedName } = usePlaceTypeHelpers()
+
+    const placeTypes = computed(() => getPlaceTypesByGroup(groupId))
+
+    // TODO sorting?
```
nicksellen, line 329: Yes please!

src/places/mutations.js
```diff
@@ -38,3 +40,28 @@ export function useSavePlaceMutation () {
     },
   ))
 }
+
+// TODO what about websocket place type updates?
```
nicksellen, line 44: Could be a nice backend addition, although not expecting them to change a lot, so maybe OK. And could make the query re-load them here and there with the right options...

src/places/queries.js
```diff
@@ -51,3 +53,19 @@ export function usePlaceListQuery (queryOptions = {}) {
...
   }
 }
+
+export function usePlaceTypeListQuery (queryOptions = {}) {
+  const query = useQuery(
+    queryKeyPlaceTypeListAll(),
+    () => placeTypeAPI.list(),
+    {
+      placeholderData: () => [],
+      staleTime: Infinity,
```
nicksellen, line 63: Could rjiggle this to cause them to be reloaded now and again, given there are no socket updates... (if I understood correctly). Not sure the best options for that.

nicksellen (approved):
Nice, great work! I added some suggestions for a few changes here and there, but approved in general.

src/locales/locale-en.json
tiltec: They are duplicated, yeah.. I think it's not too bad, these small messages are fast to translate. Transifex has some hints when a string was already translated, you can just click and it fills the previous one.

src/group/pages/EditPlaceTypes.spec.js
tiltec: the comment is also copy-pasta and likely wrong. in this case we really need to type the name as it's not a predefined option.

tiltec:
I removed the place types in the sidenav, as this is going to change in the refine-place-display PR. Which resolves approximately half of the comments ;)

I also removed the icons in PlaceEdit. looks a bit leaner now.

I just noticed a small backend quirk: if all place types are disabled and I create a new place without type, the backend picks the "Unspecified" type, presumably as default.

> I noticed it starts in "default" place type I think, with a circle icon, if I then select a place type, I can't get back to "default" or "no place type".... but I can't save without one set anyway... although the validation doesn't validate that (get server error). Minimal change is probably just adding it to the validation.

@nicksellen not sure I follow - I can save successfully without place type (backend auto-fills the default), but maybe I shouldn't be able to? Can add validation, sure :)

> When creating a new group empty of places, nothing happens when clicking the plus button to create a place

fixed (by undoing the sidenav)

> I agree with Nick in adding "Discussions" or "Working group". Not only because it's an already existing use case, but also because it's a nice way to point people towards this possibility.

@brnsolikyl these pre-defined values also need to be added in the backend, so let's do this another time?

PR #2581, opened 2022-09-15
Refine places display

### LINKED ISSUE #2559: Can't use 'back' button to navigate back from place wall/activities
...because it adds two history entries due to redirect.

```
https://karrot.world/#/group/26/place/2122
-> https://karrot.world/#/group/26/place/2122/activities
```

I think this due to https://github.com/karrot-dev/karrot-frontend/pull/2444, but could have been there earlier.

tiltec:
Branch deployment URL: https://refine-places-display.dev.karrot.world

Add a new "place gallery" page to have a nice display of all places with additional information. The sidenav place list will only contain _favorite_ places that users marked themselves (with the 'star' button on the place page).

Code-wise, I refactored a bit, mostly by merging three components `SidenavPlaces`, `SidenavPlacesUI` and `PlaceList`. Hence quite some removed lines in the PR.

## New place gallery
image

[gallery.webm](https://user-images.githubusercontent.com/4410802/192114894-dad4cb1f-98b5-4850-a058-d0858617a669.webm)

Closes #2559


tiltec:
Since I now treat 'archived' as yet another status, maybe it's too _easy_ to see archived stores? But I would leave it and see if users tell us...

nicksellen:
Great! I think it's a really nice improvement :) Elevating places...

A few things off the top of my head before diving into the code:

Would be nice to be able to click the whole box, like in the group gallery.

Maybe can include the place image (the random stuff) in the card? ... and I imagine one day we will have place images.

Add to favourites from the card?

Screenshot_2022-09-19_22-49-23

When there are no activities, or favourite icon, the whitespace looks a bit spacious, could a) make it collapse when there is no "status" row, b) move the status things to the bottom of the card, c) something else, d) nothing.

Screenshot_2022-09-19_22-50-50

Could not show "Unspecified" as the place type if that's the value set... sort of special case...

Screenshot_2022-09-19_22-36-47

I would go for a + button on the top toolbar, rather than this offer-style plus button (I think the offers page would be better off with that too)... so we move towards a more consistent UI/UX.... and I think the + button is better, or, more likely to be able to be standard....

Screenshot_2022-09-19_22-38-43

I think these would look better aligned left, rather than justified. At least that's how they are on the activities page. Filters on the left, buttons on the right. Here I am stuck in the middle with you. Oops went off piste there.

Something funny happens with the activity counts. Looking at the network tab, I _think_ the intention is to fetch activity list with page size of 1200, and do the counts from there. Sometimes it does that, sometimes it loads 10 at a time (and makes lots of requests). And the number in the UI flashes from a low number to a high number... so not sure what's going on. Maybe two things are fighting to set the page size (the page size isn't part of the query key, so I think that is quite likely to be the case somewhere...).

One idea would be to include it as part of the place status data, to add an `upcoming_activity_count` or similar. Not needed right now unless motivation is present :)

I wonder if for some groups it becomes _harder_ to find the places, as they are no longer in a big list to compare... I guess we can find out... 

I wondered at some point if using the QTable component could be a good standard, as it has a grid style mode too --> https://quasar.dev/vue-components/table#grid-style ... and then could offer users the choice of which view they prefer. Again, not needed for now, unless motivation went that way...

src/places/pages/Places.vue
```diff
@@ -1,22 +1,401 @@
...
+        :model-value="search"
+        :label="$t('BUTTON.SEARCH')"
+        outlined
+        hide-bottom-space
+        dense
+        @update:model-value="debouncedSearch"
+        @keyup.enter="event => search = event.target.value"
+      />
+      <QCheckbox
+        v-model="onlyFavorites"
```
nicksellen, line 86: British English spelling? :pray:  `onlyFavourites`?

src/places/pages/Places.vue
```diff
@@ -1,22 +1,401 @@
...
+
+const {
+  activities,
+  isFetching: isFetchingActivities,
+  hasNextPage,
+  fetchNextPage,
+} = useActivityListQuery({
+  groupId,
+  dateMin: newDateRoundedTo5Minutes(),
+  pageSize: 1200,
```
nicksellen, line 298: As I said elsewhere, I _think_ because this isn't part of the cache key, other calls to `useActivityListQuery` might fight with it and cause the number flickering that I've seen.

One solution could be to create a separate query for the purpose of getting the counts, e.g. `useActivityCountQuery` .. that does the fetching ALL pages + creating the counts logic?

src/sidenav/components/SidenavPlaces.vue
```diff
@@ -1,26 +1,130 @@
...
+            <QBadge
+              color="secondary"
+            >
+              {{ getUnreadWallMessageCount(place) > 99 ? '99+' : getUnreadWallMessageCount(place) }}
+            </QBadge>
+          </QItemSection>
+        </QItem>
+      </QList>
+
+      <div v-if="subscribedPlaces.length < 1">
```
nicksellen, line 50: Nice to think of this case :)

nicksellen (approved):
Really nice treat to have :) :star: 

I'll approve it, but some of the suggested changes might be nice to have before merge :)

tiltec:
Coming back to this :) I think I addressed most of the points you mentioned, here are the others:

> Maybe can include the place image (the random stuff) in the card? ... and I imagine one day we will have place images.

I couldn't figure out a nice design that doesn't take too much space.. also, the `RandomArt` for places seems less nice than those for groups, too busy and not that different from each other. So, I would pass on this for now...

> Could not show "Unspecified" as the place type if that's the value set... sort of special case...

I left it in to have one special case less, maybe it's a motivation to set custom place types ;)

src/maps/components/GroupMap.vue
```diff
@@ -183,7 +183,7 @@ export default {
       return { opacity: this.showOverlay ? 0.5 : 1 }
     },
     placesWithLocation () {
-      return this.places.filter(hasLocation)
+      return this.places.filter(place => place.status === 'active').filter(hasLocation)
```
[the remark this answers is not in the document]
tiltec: I kind of ignored the sidenav map... I guess it could also just show the subscribed places, but then it would also be nice to show all places on demand. Leave it for now?

PR #2589, opened 2022-09-23
Trust for role

tiltec:
Implement another "trust" button to let editors grant the _approved_ role to users.

- 'approved' trust button next to editor trust button that only shows when the role is enabled for the group
  - currently can only be done manually on the server, as there's no API or UI for customizing roles yet
- can select 'approved' role for activity participant types

Based on https://github.com/karrot-dev/karrot-frontend/issues/2361

Needs https://github.com/karrot-dev/karrot-backend/pull/1167 backend.


src/activities/helpers.js
```diff
@@ -51,13 +53,13 @@ export function useActivityHelpers () {
...
-      description: t('ROLES.NEWCOMER_DESCRIPTION'),,
-    },
-     */
+    ...(group.value?.roles.approved
+      ? [{
+          label: group.value.roles.approved.displayName,
+          value: 'approved',
+          description: group.value.roles.approved.description,
+        }]
+      : []),
```
nicksellen, line 62: I guess next step is actually iterating through the roles from the backend, but this is fine for now :)

src/group/services.js
```diff
@@ -68,6 +68,7 @@ export const useCurrentGroupService = defineService(() => {
   const isGeneralPurpose = computed(() => theme.value === 'general')
   const isBikeKitchen = computed(() => theme.value === 'bikekitchen')
   const isEditor = computed(() => getIsEditor(userId.value))
+  // TODO overlaps with group.roles, maybe rename to authUserRoles?
   const roles = computed(() => getUserRoles(userId.value))
```
nicksellen, line 72: Yes, good point.

src/group/api/groups.js
```diff
@@ -66,12 +66,13 @@ export default {
...
-    return axios.post(`/api/groups/${groupId}/users/${userId}/trust/`)
+  trustUser (groupId, userId, role) {
+    return axios.post(`/api/groups/${groupId}/users/${userId}/trust/`, { role })
   },
 
-  revokeTrust (groupId, userId) {
-    return axios.delete(`/api/groups/${groupId}/users/${userId}/trust/`)
+  revokeTrust (groupId, userId, role) {
+    // passing data in delete(url, config)! luckily, the config object accepts a data property
+    return axios.delete(`/api/groups/${groupId}/users/${userId}/trust/`, { data: { role } })
```
nicksellen, line 75: I'm a bit mixed about HTTP DELETE with a body, https://stackoverflow.com/questions/299628/is-an-entity-body-allowed-for-an-http-delete-request has mixed opinions, I didn't do a deep dive, but potentially some clients/servers have issues (e.g. someone mentions android not being able to send them), and it might cause a weird issues that's hard to understand.

Just something to keep in mind I think, no change requested.

src/locales/locale-en.json
```diff
@@ -929,6 +932,23 @@
     "WHERE_FROM": "Where you are from (not required)",
     "YOU_TRUST_USER": "You trust {userName} in {groupName}."
   },
+  "TRUST_FOR_ROLE": {
+    "FIRST_TRUST": "Be the first one trust {userName} for {role} in {groupName}!",
```
nicksellen, line 936: > Be the first one trust

should presumably read:

> Be the first one _to_ trust

src/locales/locale-en.json
```diff
@@ -929,6 +932,23 @@
...
+    "OTHER_PEOPLE_TRUST_USER": "{otherUser} trusts {userName} for {role} in {groupName}. | {count} people trust {userName} for {role} in {groupName}.",
+    "OTHER_PEOPLE_TRUST_YOU": "{otherUser} trusts you for {role} in {groupName}. | {count} people trust you for {role} in {groupName}.",
+    "PEOPLE_TRUST_USER": "You and {otherUser} trust {userName} for {role} in {groupName}. | You and {count} other people trust {userName} for {role} in {groupName}.",
+    "YOU_TRUST_USER": "You trust {userName} for {role} in {groupName}.",
+    "NO_ROLE_INFO": "You currently don't have role {role} in {groupName}. As soon as {threshold} members expressed their trust for {role} in you, you will receive the {role} role.",
+    "TRUST_PROGRESS": "{trustNeeded} more people need to express their trust in {userName} to grant the {role} role.",
+    "GIVE_DIALOG": {
+      "MESSAGE": "Do trust {userName} for {role} in {groupName}?",
```
nicksellen, line 944: > Do trust

I think should be:

> Do _you_ trust

test/mockBackend/groups.js
```diff
@@ -23,6 +23,7 @@ export function generateGroup (params = {}) {
     timezone: faker.address.timeZone(),
     status: 'active',
     theme: 'foodsaving',
+    roles: {},
```
nicksellen, line 26: Is it not an array?

src/users/components/ApproveButton.vue
```diff
@@ -0,0 +1,243 @@
...
+    const {
+      getIsCurrentUser,
+    } = useAuthHelpers()
+
+    const {
+      userId: authUserId,
+    } = useAuthService()
+
+    const isCurrentUser = computed(() => getIsCurrentUser(user.value))
+    const hasRole = computed(() => membership.value.roles.includes('approved'))
```
nicksellen, line 128: This component seems _mostly_ generic, but then specific to `approved`. I wonder how far off it is to have one trust button that can be used for editor or approved role? Seems close?

nicksellen (changes requested):
Great work! A couple of missing words in i18n files.

And a few comments from using it locally:

revokeapproved

It shows me I can revoke the trust, and will make a delete request, which succeeds, but the approved role doesn't go away, which was confusing. I know you mentioned that somewhere, as it would need a bunch of notifications, etc... I wonder if it could just remove the role and leave the notifications till later... right now it's a bit confusing (especially as the threshold is 1).

missingrolename

In both screenshots the `role` is not in missing in the text, missing i18n var?

missingrolename2

Same for history, missing role name in text.

notinlist

Would be nice to see approved status in the member list too, can wait. I think it'll come up as a need quite quickly though.

> editors can give trust for approved (editors only)

I think this bit got missed out - the idea was _only editors_ can give approval. Although @nathalie-sz and @brnsolikyl proposed _editors or approved_ can approve (https://community.karrot.world/t/applicant-trial-pickup-proposal/575/59).

The first use case we are aiming to support is from @dwaxweiler, who [said this](https://community.karrot.world/t/applicant-trial-pickup-proposal/575/60) about it:

> Concerning the discussion about who approves: My concern is that not every member should be able to approve because in our group, for example, we have designated a person who checks whether the new person has fulfilled all criteria before approving them.

... personally, I would vote for limiting it to just editors to give trust for approval role right now.

Sorry if that seems a lot of stuff at the last minute! I have capacity to help out!

src/users/components/ApproveButton.vue
tiltec: "trust for role" is still somewhat WIP right now, I guess when it has been tested more in production we can replace the editor trust button with it.

My idea was also to have a different component, with less popups, then scales better with more roles. The current implementation would just add more round buttons with unclear icons next to each other ;) Rather, I would have a section in the profile page that has some space for text and explanations.

PR #2630, opened 2022-12-15
Minor improvements to activity item UI

nicksellen:
## What does this PR do?

Two subjective design tweaks:

1. on mobile, only show icons for public link and ICS download, so the buttons don't wrap onto multiple lines
2. make the background of the edit button transparent

Before:

wrappingbuttons

After:

nonwrappingbuttons

## Checklist

- [ ] no unrelated changes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,7 @@ Please document your changes in this format:
 ```
 
 ## [Unreleased]
+- minor improvements to activity item ui @nicksellen #2630
 
 ## [10.0.1] - 2022-12-04
 ### Fixed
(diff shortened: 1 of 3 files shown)


brnsolikyl:
I don't think I'm capable to do a code review, but from the perspective of UI, definitely better! I notice the activity item is getting quite busy though. Maybe we'll need to rethink it in the future if we add more stuff.

nicksellen:
Yes, I agree. This is just the smallest change to resolve the wrapping.

I wonder from time to time whether individual activities should have their own page. With participant types they could get really big in the list....

I wonder if there is also a way to show all the participant type slots next to each other on the same line... But then not sure where the descriptions would go...

PR #2651, opened 2023-05-10
Add attachments

nicksellen:
Branch deployment URL: https://add-attachments.dev.karrot.world
Backend PR: https://github.com/karrot-dev/karrot-backend/pull/1256

# Attachments!

- can add them to any conversation message (wall, chat, thread, etc.)
- can add up to 6 attachments, each of up to 10mb
- always uploaded exactly as you give it, no cropping, no resizing
- any file type is supported
- image types will have previews and thumbnails created (1600px and 200px size respectively)
- full screen gallery for any viewable images
- can download, retaining original filename
- upload progress dialog
- this doesn't apply to non-message images (offer images, profile photo, ... those are still as before)

Known bugs:
- ~~gallery view in sidebar chat isn't fullscreen~~ fixed!

TODO:
- [ ] cannot click "download" button on attachments in list, as it's a hover-only thing
- [ ] add a way to use camera on mobile (if it doesn't already work)
- [x] MAKE IT WORK IN PRIVATE CHATS AND APPLICATION CHATS, seems a permission issue to view them...?
- [x] add ellipsis or scrolling when filename is too long
- [ ] ~~BACKEND: use aiofiles.open for sending file, as described here --> https://docs.djangoproject.com/en/4.2/ref/request-response/#fileresponse-objects~~ NO - full request async in django is tricky as middlewares are not all async... used nginx X-Accel-Redirect in the end for faster sending
- [x] BACKEND: maybe add sha256 etag

Probably out of scope next steps:
- allow s3-compatible storage backend
- [x] add headers for caching (and maybe include a content-hash in the path if possible it changes)
- attachment captions
- preview audio/video in browser
- image editing (crop, rotate, ...)
- include attachments (as links?) in email notifications?

attachments

attachments2

gallery

### DIFF
--- a/src/base/components/UploadProgressDialog.vue
+++ b/src/base/components/UploadProgressDialog.vue
@@ -1,0 +1,50 @@
+<template>
+  <QDialog
+    ref="dialogRef"
+    position="bottom"
+    persistent
+    seamless
+    @hide="onDialogHide"
+  >
+    <QCard
+      class="q-dialog-plugin"
+      style="width: 300px; margin-bottom: 0;"
+      square
+    >
+      <QLinearProgress
+        v-if="progress"
+        :value="progress.progress"
+        color="pink"
+      />
+      <QCardSection v-if="progress">
+        Uploading {{ humanStorageSize(progress.loaded) }} / {{ humanStorageSize(progress.total) }}
+      </QCardSection>
+    </QCard>
+  </QDialog>
+</template>
+
+<script setup>
+import {
+  QDialog,
+  QCard,
+  QCardSection,
+  QLinearProgress,
+  useDialogPluginComponent,
+  format,
+} from 'quasar'
+
+const { humanStorageSize } = format
+
+defineProps({
+  progress: {
+    type: Object, // the progress object axios gives us, which I think is just the native browser progress event
+    default: () => {},
+  },
+})
+
+defineEmits([
+  ...useDialogPluginComponent.emits,
+])
+
+const { dialogRef, onDialogHide } = useDialogPluginComponent()
+</script>
--- a/src/base/components/UploadProgressDialog.vue
+++ b/src/base/components/UploadProgressDialog.vue
@@ -7,17 +7,25 @@
     @hide="onDialogHide"
   >
     <QCard
-      class="q-dialog-plugin"
+      v-if="progress"
+      class="q-dialog-plugin bg-orange-1"
       style="width: 300px; margin-bottom: 0;"
       square
     >
       <QLinearProgress
-        v-if="progress"
         :value="progress.progress"
-        color="pink"
+        color="orange"
       />
-      <QCardSection v-if="progress">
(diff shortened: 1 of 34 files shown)


src/messages/components/AttachmentGallery.vue
```diff
@@ -0,0 +1,148 @@
...
+<script setup>
+import {
+  QCarousel,
+  QCarouselSlide,
+  QCarouselControl,
+  QBtn,
+  Platform,
+} from 'quasar'
+import { computed, ref } from 'vue'
+import { useEvent } from 'vue-composable'
```
nicksellen, line 63: Could use `import { useEventListener } from '@vueuse/core'` instead, as we already have vueuse.

src/base/components/UploadProgressDialog.vue
```diff
@@ -0,0 +1,50 @@
...
+      style="width: 300px; margin-bottom: 0;"
+      square
+    >
+      <QLinearProgress
+        v-if="progress"
+        :value="progress.progress"
+        color="pink"
+      />
+      <QCardSection v-if="progress">
+        Uploading {{ humanStorageSize(progress.loaded) }} / {{ humanStorageSize(progress.total) }}
```
tiltec, line 20: Should be translatable

src/messages/components/Attachments.vue
```diff
@@ -0,0 +1,465 @@
...
+      }
+    case 'text/plain':
+      return {
+        name: 'Text',
+        icon: 'fas fa-file-alt',
+      }
+  }
+  if (contentType.startsWith('image/')) {
+    return {
+      name: 'Image',
```
tiltec, line 402: could be translatable, also "Text" above and "Audio" below

src/messages/components/ConversationCompose.vue
```diff
@@ -83,12 +89,15 @@
           :prevent-white-space="false"
           small
         />
+        <pre>images: {{ message.images }}</pre>
```
tiltec, line 92: debug statement could be removed

src/messages/components/Attachments.vue
```diff
@@ -0,0 +1,465 @@
...
+
+import { showToast } from '@/utils/toasts'
+import { isViewableImageContentType } from '@/utils/utils'
+
+import AttachmentGallery from '@/messages/components/AttachmentGallery.vue'
+
+const { humanStorageSize } = format
+
+const MAX_ATTACHMENT_COUNT = 6
+const MAX_ATTACHMENT_SIZE = 10 * 1024 * 1024
```
tiltec, line 123: out of sync with backend, maybe could be part of bootstrap/config endpoint?

src/messages/components/AttachmentGallery.vue
tiltec: Seems good to avoid another dependency.

tiltec (approved):
Great! I'm sure the groups will find it useful!

src/messages/components/Attachments.vue
nicksellen: done, added to backend :)

PR #2655, opened 2023-06-12
Use vite instead of webpack

nicksellen:
Branch deployment: https://change-vite.dev.karrot.world

## What does this PR do?

Switches from webpack to vite :partying_face: ... or maybe, _"Oh god, not more javascript churn"_ :laughing: 

Anyway, it's here, and it involved quite a lot of changes:
- jest is out, [vitest](https://vitest.dev/) is in, API is mostly drop-in replacement (with global mode enabled, would be better to import those things in the future, `describe`, `expect`, etc)
- vue-croppa is out (the vue2 compat mode was playing up, so I used it as an opportunity, so new image handling in:
  - profile picture
    - can also change this directly on the profile page now!
  - group logo
  - public activity banner
  - offer images (these are no longer cropped now too)
- I did a few UI tweaks whilst I was developing too, just because...
  - rounded corners on profile pictures
  - settings page to look more streamline (like group edit page)
- probably a bunch of little things too

Big change is not using babel for the production build change, it's [esbuild](https://esbuild.github.io/) now, which does _not_ build to _es5_... so we are probably losing compatability with some browsers. It also doesn't use browserslist, but I spent time time with the build target setting to come up with `['es2015', 'edge18', 'firefox60', 'chrome55', 'opera2017', 'ios12']` as as old as I could without esbuild complaining.

One thing is _doesn't_ do that our old pipeline did is polyfill features that are missing. I don't know if that will bite us somewhere... I guess have to check sentry...

Still needs some good testing in browser, **particularly some mobile testing would be useful**.

The production bundle seems to have more files, but it's fast.

I changed the service worker thing to not cache all the built files (I don't think we should have had it like that from the beginning anyway).

Bundlesize for rollup is available as before at /bundlesize.html e.g. https://change-vite.dev.karrot.world/bundlesize.html (I don't think the numbers are actually accurate though after comparing a few manually). There are _loads_ of small files.

Everything builds to `/assets/*` now with hashes, so I added that to the yuca config too for max expiry caching.

### Rationale

I have noticed that working on Karrot frontend has felt increasingly sluggish. It takes a long time for `yarn dev` to get up to speed, and changes to the code take a while to update in the browser.

I've also been having issues where `<script setup>` components can't access any new variables until I restart the whole dev server (very time consuming...). I didn't investigate that deeply.

I wondered if vite would help with any of this. It does, although not without caveats.

`yarn dev`  is super fast with vite

First load of site is kinda slow, as it does a bazillion requests for all the npm modules.

Code change -> change in browser is super fast, that bit is really pleasing.

It _basically works_ but there is a bunch of smaller details I didn't do:
- [x] karrot splash page
- [x] check production bundling
- [x] i18n handling
- [x] landing page images
- [x] map marker not always having map available?
- [x] check other webpack options we set
- [x] set target browser options
- [x] testing
- [ ] ~~storybook??~~ sorry, no energy for storybook right now
- [x] probably some other stuff

## Checklist

- [ ] no unrelated changes


tiltec:
Could be good! Quicker response times during development is a plus. I guess it's not a problem that the production build toolchain is even more diverging from the dev one...

nicksellen:
> Could be good! Quicker response times during development is a plus. I guess it's not a problem that the production build toolchain is even more diverging from the dev one...

Did you try it out? Would be interested if it feels improved?

The prod build toolchain would also use vite, and actually I notice the existing webpack one seems to serve up a lot of individual modules anyway, so maybe they're not that different after all in concept (performance differences are still quite large).

tiltec:
Yes, I tried it out and it is faster, even considering the slow initial browser load.

I didn't do a longer dev session, with the webpack setup my problem was that RAM was getting fuller across hot reload cycles, up to a point where I needed to restart yarn dev and the browser or my system would grind to halt. Would be interesting if this is improved now.

14.06.2023 13:59:11 Nick Sellen ***@***.***>:

.circleci/config.yml
```diff
@@ -396,8 +394,11 @@ jobs:
             EOL
             git add package.json yarn.lock
             git commit -m "[CI] Update dependencies" || exit 0
-            yarn test -u storyshots
-            git commit -a -m "[CI] Update snapshots" || true
+
+            # 2023-07-03 (ns): disabling as storyshots are disabled
+            # yarn test -u storyshots
```
nicksellen, line 399: I wonder if this should still run the tests after updating deps, or perhaps that'll get run elsewhere in the branch too?

PR #2704, opened 2024-02-04
Add backup and migration docs

nicksellen:
I added some backup + migration docs, maybe @dwaxweiler is up to review them?

### DIFF
--- a/docs/src/self-hosting/backups.md
+++ b/docs/src/self-hosting/backups.md
@@ -1,0 +1,151 @@
+# Self-hosting backups
+
+If you are using co-op cloud you can use this guide to configure backups.
+
+<div class="warning">
+The backup tool is still under development by the co-op cloud team, so we need to use an unreleased version for now. Once it's released properly this guide can be made simpler.
+</div>
+
+## High-level overview
+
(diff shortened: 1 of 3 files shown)


dwaxweiler (commented):
How awesome, @nicksellen! I have managed to create my first backup. Thank you for this great guide. I have commented on two nitpicks.

docs/src/self-hosting/backups.md
```diff
@@ -0,0 +1,151 @@
...
+```
+
+### 2. Prepare Backblaze B2 account
+
+Create a Backblaze B2 account at https://www.backblaze.com/sign-up/cloud-storage
+Make sure to select the region you want, e.g. "EU Central"
+
+Then create a bucket to store the backups. You don't need to use the encryption or versioning as the backup will be encrypted and versioned by restic already.
+
+Create an application key that has "Read and Write" access to the bucket. Probably select the "Allow List All Bucket Names" option (not sure though).
```
dwaxweiler, line 44: The option "Allow List All Bucket Names" was not necessary for me.

nicksellen:
Great, I'll merge. But! I actually just remade the docs in their own rep. 

So  https://docs.karrot.world/ is now deployed from https://codeberg.org/karrot/karrot-docs. 

I included this backup and migration content. But from now on we can add stuff to the new repo, contributions welcome!

PR #2710, opened 2024-02-29
Add screenshare and mini participant view

nicksellen:
Adds two features to video calls:
- screensharing
- mini view, so you can still use the rest of Karrot whilst in a call

Screenshare:

image

Mini view:

image


nicksellen:
Moved to https://codeberg.org/karrot/karrot-frontend/pulls/2710

